PR #1, opened 2021-02-26
Process and lineage

romgrk:
Common branch for processes and lineage. WIP


backend/fms_core/serializers.py
```diff
@@ -57,9 +57,19 @@ class Meta:
...
+    extracted_from = serializers.SerializerMethodField()
+
     class Meta:
         model = Sample
         fields = "__all__"
+        extra_fields = ('extracted_from')
+
+    def get_extracted_from(self, obj):
+        if obj.extracted_from is None:
+            return ''
```
romgrk, line 69: Is returning a string required here? If not I'd choose a different empty value.

PR #2, opened 2021-03-05
Sample volume

jleblond:
The only missing thing is handling the Volume updates (to be discussed with the team).


backend/fms_core/migrations/0015_v3_1_0.py
```diff
@@ -5,6 +5,8 @@
 import django.utils.timezone
 import json
 
+from ..utils import float_to_decimal
```
UlysseFG, line 8: You can remove volume related stuff from the first migrations ...

backend/fms_core/migrations/0015_v3_1_0.py
```diff
@@ -98,6 +100,12 @@ def create_lineage_from_extracted_and_revisions(apps, schema_editor):
         # Save to database
         version.save()
 
+def add_values_to_sample_volume(apps, schema_editor):
```
UlysseFG, line 103: Same as above.

backend/fms_core/resources/extraction.py
```diff
@@ -206,12 +209,15 @@ def after_save_instance(self, instance, using_transactions, dry_run):
         # Update volume and depletion status of original sample, thus recording
         # that the volume was reduced by an extraction process, including an ID
         # to refer back to the extracted sample.
+        parent_volume_left = self.extracted_from.volume - self.volume_used
```
UlysseFG, line 212: i would set self.extracted_from.volume in the ( if field.column_name == "Volume Used (uL)":) section above. Once we remove the volume history there won't be any reason to do it after saving. You can in the meantime use self.extracted_from.volume below instead of parent_volume_left...

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -60,9 +60,7 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, ind
   const isLoaded = samplesByID[id] && !sample.isFetching && !sample.didFail;
   const isFetching = !samplesByID[id] || sample.isFetching;
   const sampleKind = sampleKindsByID[sample.sample_kind]?.name
-  const volume = sample.volume_history
-    ? parseFloat(sample.volume_history[sample.volume_history.length - 1].volume_value).toFixed(3)
-    : null;
+  const volume = sample.volume ?? undefined
```
UlysseFG, line 63: I would keep toFixed(3) on the volume...

backend/fms_core/resources/extraction.py
```diff
@@ -106,6 +106,9 @@ def import_field(self, field, obj, data, is_m2m=False):
             # initializing the volume history of the EXTRACTED sample, so the
             # actual history entry is of the "normal" type (UPDATE).
             vol = blank_str_to_none(data.get("Volume (uL)"))  # "" -> None for CSVs
+
+            obj.volume = vol
```
UlysseFG, line 110: i would set obj.volume directly above and use it below to make sure not to leave those intermediate variables floating when we remove the volume_history, but it is not that important ...

UlysseFG (commented):
Left a couple comments ... Stuff about temporary state due to volume_history being still there as long as we do a good job cleaning when we remove the field...

backend/fms_core/resources/extraction.py
jleblond: I dont understand, where would you put obj.volume = vol ?

UlysseFG: My point was the line is not useful.

jleblond: The blank line?

PR #3, opened 2021-03-11
Merging changes done to qc on master

UlysseFG:
Changes made to QC for patches applied to MASTER.


UlysseFG:
Removed RESPIRATORY SECRETION after revision from Alex. B.

romgrk:
Can you remind me what's the git workflow? I thought master was supposed to be ahead of qc, but it sounds like those changes are already made on qc and I'm confused how the work is done.

UlysseFG:
That PR is done so master receives changes made to prod in patches meant for quick release. Normal workflow depends if the change is meant for immediate release (in this case we work on the QC branch(es at the moment)) as a patch or if it is following the planned release in which case we work mainly on master until the version is on track to go to prod. So you are right about master being ahead. I push it the reverse way for patch to make sure things make it to prod before i push it to master not to have to do that 2 times in case of last minute changes (like this patch).

PR #4, opened 2021-03-11
Sample transfer

jleblond:


backend/fms_core/models/sample.py
```diff
@@ -279,23 +281,37 @@ def add_error(field: str, error: str):
             )
 
         if self.extracted_from:
-            if self.extracted_from.sample_kind.name in Sample.BIOSPECIMEN_TYPES_NA:
+            try:
+                ProcessSample = apps.get_model("fms_core", "ProcessSample")
+
+                sample_lineage = SampleLineage.objects.get(child_id=self.id)
```
UlysseFG, line 287: Not sure why you insist on using this approach instead of the relation (child_of) if it forces you to use exception handling to get it to work...

backend/fms_core/models/sample.py
```diff
@@ -279,23 +281,37 @@ def add_error(field: str, error: str):
...
+                protocol_name = ProcessSample.objects.get(id=sample_lineage.process_sample_id).protocol_name
+                if protocol_name == 'Extraction':
```
UlysseFG, line 289: this is some very strange way to implement this ... you have a property that should give you the extracted_from. You should implement the above in the property and remove it from here ... or maybe we want to remove the extracted_from property but going both way at once is not right.

backend/fms_core/models/sample.py
```diff
@@ -279,23 +281,37 @@ def add_error(field: str, error: str):
...
+                    if self.sample_kind.name not in Sample.BIOSPECIMEN_TYPES_NA:
+                        add_error("sample_kind", "Extracted sample need to be a type of Nucleic Acid.")
+                elif protocol_name == 'Transfer':
+                    if self.sample_kind != self.extracted_from.sample_kind:
+                        add_error("sample_kind", "Sample kind need to remain the same during transfer")
+            except SampleLineage.DoesNotExist:
                 add_error(
                     "extracted_from",
-                    f"Extraction process cannot be run on sample of type {', '.join(Sample.BIOSPECIMEN_TYPES_NA)}"
+                    (f"Sample Lineage not found for {self.name} "),
```
UlysseFG, line 304: There is no reason for this to ever happen ... not sure why test for this ... if extracted_from is true there is definitively a SampleLineage with this sample as child ... if this get caught it's that we have done a bad job coding not that we have an exception.

backend/fms_core/models/transferred_sample.py
```diff
@@ -0,0 +1,23 @@
...
+from .sample import Sample
+from .protocol import Protocol
+from .process import Process
+from .process_sample import ProcessSample
+
+
+__all__ = ["TransferredSample"]
+
+
+class TransferredSampleManager(models.Manager):
```
UlysseFG, line 11: Not sure we need this model ... displaying the transfered samples on the backend do not appear really useful and if we keep doing this we'll end up having dozens of proxy models. Maybe we need to decide if we want to display all this in the backend.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+            'source_depleted',
+            'volume_used',
+            'transfer_date',
+            'comment',
+        )
+    def before_import(self, dataset, using_transactions, dry_run, **kwargs):
+        skip_rows(dataset, 7)  # Skip preamble
+
+    def import_obj(self, obj, data, dry_run):
+        self.extracted_from = Sample.objects.get(
```
UlysseFG, line 73: Rename this to match what it contains ... self.transfered_from possibly. The current name is just confusing.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+            destination_parent_container_barcode = get_normalized_str(data, "Destination Parent Container Barcode")
+
+            if destination_parent_container_barcode:
+                shared_parent_info = dict(
+                    barcode=destination_parent_container_barcode,
+                )
+                try:
+                    parent = Container.objects.get(**shared_parent_info)
+                except Container.DoesNotExist:
+                    raise Exception("Destination parent container does not exist")
```
UlysseFG, line 106: I don't think you should fail the template submission if the destination parent container is missing. I think  this should be optional. They can always localize it later. Also if the destination container exists we don't want to move it so we ignore the destination parent container even if it is there.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+
+            try:
+                obj.container = Container.objects.get(barcode=get_normalized_str(data, "Destination Container Barcode"))
```
UlysseFG, line 109: I would use get_or_create() here instead of the exception handling approach. You can list the params in a dict skipping the missing optional variables.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+                )
+
+            obj.coordinates = get_normalized_str(data, "Destination Location Coord")
+
+            return
+
+        if field.column_name == "Volume Used (uL)":
+            vu = blank_str_to_none(data.get("Volume Used (uL)"))  # "" -> None for CSVs
+            if vu:
+                self.volume_used = float_to_decimal(vu)
```
UlysseFG, line 128: you do not need to put this in self.volume_used if you put it in volume... the self.volume_used variable was to get the value to the process_sample, but since volume is part of the instance you can use that value instead to create your process_sample.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+            else:
+                self.volume_used = None
+
+            obj.volume = vu
```
UlysseFG, line 132: Do not forget to convert to decimal your vu.

backend/fms_core/resources/transfer.py
jleblond: Volume_used is still an attribute on the parent sample... I think it is right to modify it while we still keep it. But maybe I should add the volume currently used to the volume_used already stored...

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,187 @@
...
+        self.extracted_from.volume = float_to_decimal(self.extracted_from.volume - instance.volume)
+        self.extracted_from.volume_history.append(create_volume_history(
+            VolumeHistoryUpdateType.TRANSFER,
+            self.extracted_from.volume
+        ))
+
+        self.extracted_from.save()
+
+        super().after_save_instance(instance, using_transactions, dry_run)
+        reversion.set_comment("Imported extracted samples from template.")
```
UlysseFG, line 172: Change comment to reflect it is a transfer.

backend/fms_core/resources/transfer.py
jleblond: I check whether the field is filled. If it is filled with an inexisting container barcode, yes I think I should fail the submission.

backend/fms_core/models/transferred_sample.py
jleblond: I need to register TransferredSample model at some places. So should I just remove the TransferredSampleManager class, and fetch the objects directly in TransferredSample?

backend/fms_core/resources/transfer.py
jleblond: I got the same code conventions from Extraction. I really dont mind changing it, and even prefer using get_or_create, but if so, I'd prefer it to be used uniformly with get_or_create instead of try/except in the rest of the codebase.

backend/fms_core/resources/transfer.py
UlysseFG: I agree with your logic, I missed the test. if we are to localize the container it has to exist first ...

backend/fms_core/resources/transfer.py
UlysseFG: volume_used is to be removed ... I left it there until the extraction was done ... i must have forgotten to remove it in the wake of the repo migration... The volume_used was solely used with extractions so since it is move to ProcessSample the volume_used on sample should be removed ...

backend/fms_core/models/sample.py
jleblond: How would you access the protocol name from there? And you have to get the child_id, and not the parent, because the parent could result from multiple extractions/transfers/other processes.

backend/fms_core/models/sample.py
jleblond: noted. If I can suggest using a different name than 'child_of' for the attribute, I had forgotten it was a SampleLineage, I thought child_of meant the parent Sample.

backend/fms_core/models/transferred_sample.py
UlysseFG: No i meant we do not need both ... transferredSample is not a real model it is just a stub for admin to display a list of transfered samples, but i do not think we should do that for each process. We will not be going back to this interface and it has no use for admin since there is (or will be ;) ) an API endpoint to list those transfers and extraction ...

jleblond: Let's keep it for now. The back-end allows double-checking the data entered by displaying it in a convenient table-format before saving it. It is a feature that is requested on the front-end, but for now it's not there.

backend/fms_core/example_data/csv/samples_transfers.csv
```diff
@@ -0,0 +1,10 @@
+Sample Transfer Template,,,,,,,,,,,,
+,,,,,,,,,,,,
+,,,,,,,,,,,,
+,,,,,,,,,,,,
+,,,,,,,,,,,,
+,,,,,,,,,,,,
+,,,,,,,,,,,,
+#,Source Container Barcode,Source Location Coord,Destination Container Barcode,Destination Location Coord,Destination Container Name,Destination Container Kind,Destination Parent Container Barcode,Destination Parent Container Coord,Source Depleted,Volume Used (uL),Transfer Date,Comment
+1,tube001,,newtubefortransfer,,newtubefortransfer,Tube,,,,10,2021-01-01,Test Comment
```
UlysseFG, line 9: I would test cases with containers with coords and localize the destination container... It may help us catch errors down the line.

backend/fms_core/resources/transfer.py
```diff
@@ -0,0 +1,184 @@
...
+
+            return
+
+        if field.column_name == "Volume Used (uL)":
+            vu = blank_str_to_none(data.get("Volume Used (uL)"))  # "" -> None for CSVs
+            vu = float_to_decimal(vu)
+            obj.volume = vu
+            obj.volume_history = [create_volume_history(
+                VolumeHistoryUpdateType.TRANSFER,
+                str(float_to_decimal(vu))
```
UlysseFG, line 131: float_to_decimal is redundant here.

backend/fms_core/models/sample.py
UlysseFG: child_of returns the parent sample but can be filtered with any value of the intermediate table, including ProcessSample. This is up to you i really just dislike exception handling of normal operations.

backend/fms_core/models/sample.py
UlysseFG: it should all be checked in the above: if self.extracted_from:
modify extracted_from to something like 
self.child_of.filter(parent_sample__child=self, parent_sample__process_sample__process__protocol__name__exact="Extraction").first()

backend/fms_core/resources/transfer.py
UlysseFG: Ok lets put a Comment here in the code to remember to convert this and the equivalent in extraction to something using get_or_create ?

PR #5, opened 2021-03-12
DB diagram changed. Package update to cover default rc-util install n…

UlysseFG:
…ot installing the most recent version. Moving env to base directory.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -23,13 +23,14 @@ Python package dependencies are listed in `requirements.txt`.
   
 ## Running locally for development
 
-  1. Set up a virtual environment with Python 3.6 or later, and install 
+  1. Set up a virtual environment with Python 3.8 or later, and install 
      dependencies into it:
      
      ```bash
-     virtualenv -p python3 ./env
+     cd .. # move into base repository directory
+     virtualenv -p python3.8 ./env
      source env/bin/activate
-     pip install -r requirements.txt
+     pip3.8 install -r ./backend/requirements.txt
      ```
      
   2. Create a database for the application to use locally in Postgres. The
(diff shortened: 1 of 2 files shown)


backend/README.md
```diff
@@ -130,7 +131,7 @@ coverage run ./manage.py test
 
 ## Database diagram
 
-![Database Schema Diagram](docs/fms_database_diagram.png)
+[Database Schema Diagram](https://dbdiagram.io/d/604aa285fcdcb6230b23c0e7)
```
jleblond, line 134: If this is on master, maybe it should include the latest tables, such as Sample Kind, Sample Lineage, Protocol, etc. for example?

UlysseFG: yeah you are right i must be sleep walking to do that on master with the old diagram ... I'll include the new changes

UlysseFG: Ok updated the diagram to be in sync with master instead ...

jleblond (approved):

backend/README.md
```diff
@@ -23,13 +23,14 @@ Python package dependencies are listed in `requirements.txt`.
...
+  1. Set up a virtual environment with Python 3.8 or later, and install 
      dependencies into it:
      
      ```bash
-     virtualenv -p python3 ./env
+     cd .. # move into base repository directory
+     virtualenv -p python3.8 ./env
      source env/bin/activate
-     pip install -r requirements.txt
+     pip3.8 install -r ./backend/requirements.txt
```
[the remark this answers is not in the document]
UlysseFG: This is mainly to stay in sync with the deployment i've done on de DEV server and that will eventually be extended to QC. The /instance/ directory with the uwsgi config and auto_deploy was moved to the repo root. The uwsgi version is dependent on the python version so i moved the env along to cover it.

PR #7, opened 2021-03-15
Script for backup restore

jleblond:


romgrk:
Maybe we should move the `. env/bin/activate` inside the script files. I didn't put it there because my virtualenvs are always autoactivated/deactivated depending on my folder, but otherwise those script files are not runnable by themselves.

You can use `__dirname="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"` to get the file directory, independent from the CWD.

For the npm command, maybe add a line about it in the readme how it needs to be run with `$FILE` set, cause there is no guard against not having `$FILE` set. Also maybe namespace it with something like `db:...`.

PR #8, opened 2021-03-15
Remove volume used

jleblond:


backend/fms_core/tests/test_models.py
```diff
@@ -197,7 +197,7 @@ def test_extracted_sample(self):
         volume_used = Decimal('0.01')
         parent_sample = self.parent_sample
         s = Sample.objects.create(**create_extracted_sample(sample_kind=self.sample_kind_DNA,
-                                                            volume_used=volume_used,
+                                                            volume=volume_used,
```
UlysseFG, line 200: see above comment. the volume_used value should be inserted in the ProcessSample only... This applies to the other similar changes below.

backend/fms_core/tests/constants.py
```diff
@@ -46,7 +46,7 @@ def create_sample(sample_kind, individual, container, coordinates='', **kwargs):
     }
 
 
-def create_extracted_sample(sample_kind, individual, container, volume_used, coordinates='', **kwargs):
+def create_extracted_sample(sample_kind, individual, container, volume, coordinates='', **kwargs):
```
UlysseFG, line 49: I'm not sure i get all those changes here ... Removing volume_used is fine but setting the volume for all the following test to volume_used is likely to break those test since this is not the intended behaviour (volume should not be = volume_used but to the value stored in the template for volume. the parent volume should = volume - template_volume_used. You can put a param to set volume but set it to 0 by default as it is the current behaviour.

backend/fms_core/tests/test_models.py
```diff
@@ -358,7 +329,7 @@ def test_negative_volume(self):
                                                   comment="ProcessSample test_negative_volume")                                                                       
                 SampleLineage.objects.create(parent=parent_sample, child=negative_volume_used, process_sample=ps)
             except ValidationError as e:
-                self.assertTrue('volume_used' in e.message_dict)
+                self.assertTrue('volume' in e.message_dict)
```
UlysseFG, line 332: You should use failure in test to review our code behaviour instead of "fixing" the test ;). The volume_used error is still a valid error since it point the user to the column used in the template... The validation done in ProcessSample should generate a volume_used error message here.

backend/fms_core/tests/test_models.py
```diff
@@ -291,41 +287,10 @@ def test_sample_kind(self):
...
-        invalid_volume_used.save()
-        p = Process.objects.create(protocol=self.extraction_protocol, comment="Process test_volume_used")
-        with self.assertRaises(ValidationError):
-            try:
-                ps = ProcessSample.objects.create(process=p,
-                                                  source_sample=self.parent_sample,
-                                                  execution_date=timezone.now(),
-                                                  volume_used=volume_used,
-                                                  comment="ProcessSample test_volume_used")
-                SampleLineage.objects.create(parent=self.parent_sample, child=invalid_volume_used, process_sample=ps)
```
UlysseFG, line 308: You may want to keep the first part of this test which is still valid given that you need a ProcessSample.volume_used for extraction

UlysseFG (commented):
One little thing ...

backend/fms_core/tests/test_models.py
jleblond: yes but I think that is not where it should be tested. this is Sample Model test, and volume_used is no longer a Sample attr. If there is a need for test it should be made on ProcessSample

UlysseFG: This is actually an extraction test as far as i can see ... as you mention the test is performed on ProcessSample (in the try). The test could also be moved to a ProcessSample specific section it would be logical, but i would not drop the test altogether...

PR #9, opened 2021-03-16
Improve Sample Transfer resource code

jleblond:
Some improvements.

Note: for a get_or_create function, for the cases where the attrs in get would differ from the ones for the create, I learned that there is the default= attribute that prevents those fields from being looked up in the get, but are used for the create. ;P


UlysseFG (approved):
Looks good we may start thinking of using a timezoned timestamps ... the utc times stored are clashing a bit with the timezone dates at other places ... just a reflexion we may have ...

PR #10, opened 2021-03-17
Process list detail

UlysseFG:


backend/fms_core/viewsets.py
```diff
@@ -422,8 +443,74 @@ class SampleKindViewSet(viewsets.ModelViewSet):
...
+
+        protocol_names_by_id = {protocol.id: protocol.name for protocol in Protocol.objects.all()}
+
+        return Response({
+            "total_count": ProcessSample.objects.all().count(),
+            "protocol_counts": {
+                protocol_names_by_id[c["process__protocol"]]: c["process__protocol__count"]
+                for c in self.queryset.values("process__protocol").annotate(Count("process__protocol"))
+            },
+        })
```
romgrk, line 513: Do we really want to have process by name rather than by ID in the frontend? What's the advantage of having them by name? I'd rather have a single identifier.

frontend/src/components/JumpBar.js
```diff
@@ -107,6 +107,7 @@ function getPath(type, id) {
     case 'container':  return `/containers/${id}`;
     case 'sample':     return `/samples/${id}`;
     case 'individual': return `/individuals/${id}`;
+    case 'process':    return `/processes/${id}`;
```
romgrk, line 110: Just to confirm, protocols/processes don't need to be searchable through the jumpbar?

frontend/src/modules/processes/reducers.js
```diff
@@ -0,0 +1,140 @@
...
+        case PROCESSES.LIST_PROTOCOLS.ERROR:
+            return {
+                ...state,
+                isFetching: false,
+                error: action.error,
+            };
+        default:
+            return state;
+    }
+};
```
romgrk, line 42: Ideally it should be one file one reducer, I'd put these in their own file with their own action namespace (`PROTOCOLS.LIST` rather than `PROCESSES.LIST_PROTOCOLS`).

backend/fms_core/viewsets.py
UlysseFG: I was using the names for protocol to copy the way this is done with the sample kind but it does not have to be name. We could use the id's instead, but having a human readable summary may be a good thing to have... I am not sure if that counts for anything in our case though.

frontend/src/components/JumpBar.js
UlysseFG: I don't think so ... Users won't have any useful identifier to search for i think.

frontend/src/modules/processes/reducers.js
UlysseFG: ok i'll see if i can move them out without blowing something up ;)

backend/fms_core/viewsets.py
UlysseFG: ok changed the summaries to use IDs for protocols and sample kinds.

frontend/src/modules/processes/reducers.js
UlysseFG: Ok i moved Protocols into their own files... Sample Kind may need to receive the same treatment if we want this approach to be uniform...

romgrk: Do you intend to use all the code in the newly created protocols files or did you just copy the template from another file? Unless we plan to use all the code, I'd keep only the required minimum.

UlysseFG: You mean the get and list_table and minimal filter description ? I chose to let them there. This would be more helpful if you could point me to the code you find extraneous ...

romgrk: In `protocols/actions.js`, you only use `.list()`, all the other functions are unused, and the associated parts in the reducer are unused.

UlysseFG: Thanks i'll clean this up.

jleblond:
Please check the filter for the Protocol type. On my end, if I select 'Transfer' it still displays the Extractions

backend/fms_core/viewsets.py
```diff
@@ -518,13 +597,11 @@ def summary(self, _request):
...
 
-        sample_kind_names_by_id = {sample_kind.id: sample_kind.name for sample_kind in SampleKind.objects.all()}
-
+        
         return Response({
             "total_count": Sample.objects.all().count(),
-            "extracted_count": SampleLineage.objects.all().count(),  # WARNING !!! will need a filter when transfer are added
             "kinds_counts": {
-                sample_kind_names_by_id[c["sample_kind"]]: c["sample_kind__count"]
+                c["sample_kind"]: c["sample_kind__count"]
```
jleblond, line 604: This means the sample kinds being displayed by id rather than by name?

frontend/src/components/filters/FilterInputNumber.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import {Input} from "antd";
+import FilterLabel from "./FilterLabel"
+import * as style from "./style"
+
+
+const FilterInput = ({
```
jleblond, line 7: Whether this is used or not, the name of a component should match the name of its file, especially if its a default export Component (meaning its the only component being exported in that file). Otherwise, this is possibly confusing, and in our case we have another component named 'Filterinput' too in FilterInput.js

backend/fms_core/serializers.py
```diff
@@ -55,6 +58,29 @@ class Meta:
...
+
+class ProcessSampleExportSerializer(serializers.ModelSerializer):
+    protocol_name = serializers.CharField(read_only=True, source="process.protocol.name")
+    child_sample_name = serializers.CharField(read_only=True)
+    source_sample_name = serializers.CharField(read_only=True)
+
+    class Meta:
+      model = ProcessSample
+      fields = ('id', 'protocol_name', 'source_sample_name', 'child_sample_name', 'volume_used', 'execution_date', 'comment')
+      extra_fields = ('protocol_name', 'source_sample_name', 'child_sample_name')
```
jleblond, line 83: You dont need to have the fields declared in extra_fields too, it is already declared in fields

backend/fms_core/viewsets.py
```diff
@@ -422,8 +443,78 @@ class SampleKindViewSet(viewsets.ModelViewSet):
...
+class ProcessSampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
+    queryset = ProcessSample.objects.all().select_related("process").prefetch_related("lineage")
+    queryset = queryset.annotate(child_sample=F("lineage__child"))
+    queryset = queryset.annotate(child_sample_name=F("lineage__child__name"))
+    queryset = queryset.annotate(source_sample_name=F("source_sample__name"))
+
+    serializer_class = ProcessSampleSerializer
+
+    ordering_fields = (
+        *_list_keys(_process_sample_filterset_fields),
```
jleblond, line 463: This wont work in the case of CSV export, which uses a CSVRenderer from a library.
In viewset in get_renderer_context, for your action, you will need to declare the header and the labels of your context (see as an example line 558) If you find a better method, feel free to change it, but I had been looking for a while, and havent found anything else.

backend/fms_core/viewsets.py
UlysseFG: yeah this means the count is returned by id instead of name. Since i did the change on protocol i made the change for the kinds too so we would be coherent.

backend/fms_core/viewsets.py
UlysseFG: thanks

frontend/src/components/containers/ContainerEditContent.js
```diff
@@ -14,7 +14,7 @@ import {container as EMPTY_CONTAINER} from "../../models";
 import api, {withToken} from "../../utils/api";
 
 const requiredRules = [{ required: true, message: 'Missing field' }]
-const barcodeRules = [{ pattern: /^[a-zA-Z0-9.\-_]{1,199}$/ }]
+const barcodeRules = [{ pattern: /^[^$]{1,199}$/ }]
```
romgrk, line 17: So this means a barcode is anything except a dollar sign, right?

jleblond: Yup it had to be changed, the backend was correct but not the frontend. It is already on master, as part of another PR

UlysseFG: Yeah sorry i rebased my code to import changes to cypress

frontend/src/components/filters/FilterInputNumber.js
UlysseFG: Corrected.

backend/fms_core/serializers.py
UlysseFG: Removed extra_fields

frontend/src/components/processes/ProcessesListContent.js
```diff
@@ -0,0 +1,164 @@
...
+import api, {withToken}  from "../../utils/api"
+
+import {listTable, setFilter, setFilterOption, clearFilters, setSortBy} from "../../modules/processes/actions";
+import {actionsToButtonList} from "../../utils/templateActions";
+import {withSample} from "../../utils/withItem";
+import {PROCESS_FILTERS} from "../filters/descriptions";
+import getFilterProps from "../filters/getFilterProps";
+import getNFilters from "../filters/getNFilters";
+import FiltersWarning from "../filters/FiltersWarning";
+import mergedListQueryParams from "../../utils/mergedListQueryParams";
```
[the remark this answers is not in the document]
UlysseFG: re-ordered.

jleblond (approved):
Looks good to me. Just to be clear I would put the link to the ProcessSample (or Process) object shown on the Process ID column, not the type of Protocol. But this is very minor, and could be done later if the same comment is made by others.

UlysseFG:
My plan was to eventually link the real process (from the process table) on the Process ID column ... I must admit the protocol column links are not that intuitive as people are not used to click on those tags ... Linking them on the Process ID column would be inaccurate though since these are not Process Sample IDs.

jleblond:
Ok I see, I would then just make a clear distinction between what you renamed Process (ProcessSample) and the column Process (which are Process objects, but you wanted to rename them Process Batch or so?)

UlysseFG:
To the user process will be a process but the processSample table distinction is hidden to them. The ProcessesDetailContent.js should be renamed ProcessesSamplesDetailContent.js for our own sake the real Process page will be ProcessesDetailContent.js.

UlysseFG:
I agree there is still something odd with linking protocol column to the process sample detail... I'm not sure what would be best you have a suggestion ? I could add a column expressly for linking the process sample detail but i don't know if i want to display the processSample ID to confuse people with multiple ID for processes...

jleblond:
Maybe call it 'Applied Process Details' or so, and have the other column being Process Batch? something like that? it does not have to reflet the internal structure of our models, but just so it provides clear information about one being a batch, and the other one being detailed information

PR #11, opened 2021-03-19
Add update process

jleblond:
Still WIP. Very long migration, not optimized yet, but it should give the expected results for Sample Update Processes.
Let me know if you have comments on the methodology, and if you see easy ways of optimizing its speed.


backend/fms_core/migrations/0018_test4.py
```diff
@@ -0,0 +1,70 @@
...
+        previous_volume_value = None
+        for index, sample_version in enumerate(sample_versions):
+            data = json.loads(sample_version.serialized_data)
+            update_comment = data[0]["fields"]["update_comment"]
+            if index != 0: # skip first version, as this will be an object creation
+                volume_history = data[0]["fields"]["volume_history"]
+                last_vh = volume_history[-1]
+                if last_vh["update_type"] == "update":
+                    process_sample_comment = ''
+                    if update_comment != previous_update_comment:
```
UlysseFG, line 33: What happens if someone submit 2 updates with the same comment back to back ?

backend/fms_core/migrations/0018_test4.py
```diff
@@ -0,0 +1,70 @@
...
+                        process_sample_comment = update_comment
+
+                    revision = Revision.objects.get(id=sample_version.revision_id)
+                    process, _ = Process.objects.get_or_create(comment=f'{revision.id}',
+                                                               protocol=update_protocol)
+                    volume_used = float_to_decimal(float(previous_volume_value) - float(volume_history[-1]["volume_value"]))
```
UlysseFG, line 39: As far as i see, this does not work if the second version (first update) is not a volume update. This will subtract the first vh volume value from None which will lead to a -total_sample_volume update that does not exist...

backend/fms_core/migrations/0018_test4.py
```diff
@@ -0,0 +1,70 @@
...
+            update_comment = data[0]["fields"]["update_comment"]
+            if index != 0: # skip first version, as this will be an object creation
+                volume_history = data[0]["fields"]["volume_history"]
+                last_vh = volume_history[-1]
+                if last_vh["update_type"] == "update":
+                    process_sample_comment = ''
+                    if update_comment != previous_update_comment:
+                        process_sample_comment = update_comment
+
+                    revision = Revision.objects.get(id=sample_version.revision_id)
```
UlysseFG, line 36: Might be faster to preload all the revisions in a dict id: date_created. You are likely reextracting the same revision many times.

backend/fms_core/migrations/0018_test4.py
```diff
@@ -0,0 +1,70 @@
...
+                                                 source_sample_id=sample.id,
+                                                 execution_date=revision.date_created,
+                                                 volume_used=volume_used,
+                                                 comment=process_sample_comment)
+
+            previous_update_comment = update_comment
+            previous_volume_value = data[0]["fields"]["volume_history"][-1]["volume_value"]
+
+
+    for process in Process.objects.filter(protocol__name="Update"):
```
UlysseFG, line 50: This possibly slows you down too might be faster to do Process.objects.get_or_create(comment=revision_dict[revision_id][comment] + " Revision ID [" + str(revision_id) + "]" ...

UlysseFG (commented):
Looks good as far as i see. This looks reliable...

backend/fms_core/migrations/0018_test4.py
jleblond: I dont think there is a way around it... and Id rather not have the information displayed directly (remember, it is still stored in versions) than having the wrong information being displayed as an update comment (i.e. having as an update comment: 'extraction done blablabla' for a sample update)

UlysseFG: Yeah this is all right ;)  i don't see a way around it either ...

PR #12, opened 2021-03-22
Updated container creation template to include Tube box 6x6

UlysseFG:
Tube box was not included in the selection drop down.


jleblond:
Is this new container kind in other templates too? I know there are some templates which have a Container Kind field too.

UlysseFG:
Not a new kind ... it's been there a while but i think the template was not updated at the time. The other container template do not really use that information as far as i know. The transfer template should only have the sample-holing container types. I don't know of any other template that would need to be updated, but i welcome any suggestion.

jleblond (approved):

PR #14, opened 2021-03-23
Migration optimization

jleblond:
* I have read about other ways of optimizing the django orm queries, either by using bulk_update/bulk_create (which bypass the save()/presave/postsave methods in the models btw), or using iterator() instead of caching the whole queryset into the memory, but I did not have much proof of the efficiency when running our migration.

* The best I could do is to regroup all the versions modifications into a single operation rather than many. Going through the 80K Sample objects in Version were the slower steps, and we do not modify Sample Versions during the migration when making modifications to Sample objects anyways, so the order of operations does not really matter in our case. 

On my local machine, I was able to reduce the migration time from 4min30 to 2mins30, which is still a significant improvement imo.

### DIFF
--- a/backend/fms_core/migrations/0016_v3_1_0.py
+++ b/backend/fms_core/migrations/0016_v3_1_0.py
@@ -44,7 +44,7 @@ def create_lineage_from_extracted_and_revisions(apps, schema_editor):
     extraction_protocol = protocol_model.objects.get(name="Extraction")
     extracted_samples_info = {}
 
-    for revision in revision_model.objects.filter(comment="Imported extracted samples from template.").iterator():
+    for revision in revision_model.objects.filter(comment="Imported extracted samples from template.").all():
         extracted_samples = version_model.objects.filter(revision_id=revision.id,
                                                          content_type__model="sample",
                                                          object_repr__icontains="(extracted, ")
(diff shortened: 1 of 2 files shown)


backend/fms_core/migrations/0016_v3_1_0.py
```diff
@@ -64,7 +44,7 @@ def create_lineage_from_extracted_and_revisions(apps, schema_editor):
     extraction_protocol = protocol_model.objects.get(name="Extraction")
     extracted_samples_info = {}
 
-    for revision in revision_model.objects.filter(comment="Imported extracted samples from template."):
+    for revision in revision_model.objects.filter(comment="Imported extracted samples from template.").iterator():
```
UlysseFG, line 47: your iterator() iterates on a queryset of 1... this would be more useful on the queryset iterating on the version_model.

backend/fms_core/migrations/0017_v3_1_0.py
```diff
@@ -12,16 +12,34 @@ def add_values_to_sample_volume(apps, schema_editor):
         sample.volume = float_to_decimal(sample.volume_history[-1]["volume_value"] if sample.volume_history else 0.0)
         sample.save()
 
+
+def handle_sample_versions(apps, schema_editor):
```
UlysseFG, line 16: Nice idea to regroup them in a single iteration ... We only need to be aware that they are no longer in the same transaction (since they are in a different migration) as the rest of the related operations.

UlysseFG (approved):
Looks good.

PR #15, opened 2021-03-23
Change clean validations in Sample

jleblond:
I realize there was some data validation error making the app crash when creating a sample in the old DjangoAdmin SampleForm interface. Even though this interface will not be used, it still pointed to fixes that had to be done for the Sample model in clean()

### DIFF
--- a/backend/fms_core/models/sample.py
+++ b/backend/fms_core/models/sample.py
@@ -275,18 +275,32 @@ class Sample(models.Model):
 
         self.normalize()
 
+        # Sample Kind
         sample_kind_choices = (Sample.BIOSPECIMEN_TYPE_NA_CHOICES if self.extracted_from
                                else Sample.BIOSPECIMEN_TYPE_CHOICES)
-        if self.sample_kind.name not in frozenset(c[0] for c in sample_kind_choices):
-            add_error(
-                "sample_kind",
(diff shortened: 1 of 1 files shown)


backend/fms_core/models/sample.py
```diff
@@ -275,18 +275,32 @@ def add_error(field: str, error: str):
...
-            add_error(
-                "sample_kind",
-                (f"Sample Kind name {self.sample_kind.name} not valid for "
-                 f"{' extracted' if self.extracted_from else ''} sample {self.name}"),
-            )
 
-        if self.transferred_from:
-            if self.sample_kind != self.transferred_from.sample_kind:
-                add_error("sample_kind", "Sample kind need to remain the same during transfer")
+        if self.sample_kind_id is not None:
```
UlysseFG, line 282: There should be an error if sample_kind_id is None ...

jleblond: This is checked automatically by django with the field being blank=False (or not blank not declared in the field for Sample model).

UlysseFG, line 282: Might be good to add an error if this is None since we can more easily format those message to be human readable.

jleblond: There will be an error, it is automatically generated by Django with fields having the attribute blank=False (or not having the attribute blank, which results by default to blank=False)

PR #17, opened 2021-03-25
User password & permissions

romgrk:
Closes https://206.12.92.46/issues/327

This PR implements a profile page that allows the user to modify some of their details. User permissions not included for the moment.

### DIFF
--- a/frontend/src/components/profile/ProfileContent.js
+++ b/frontend/src/components/profile/ProfileContent.js
@@ -1,0 +1,240 @@
+import React, {useState, useRef} from "react";
+import {connect} from "react-redux";
+import {useHistory, useParams} from "react-router-dom";
+import {Alert, Button, Form, Input, Tag, Typography} from "antd";
+import { UserOutlined } from "@ant-design/icons";
+
+import AppPageHeader from "../AppPageHeader";
+import PageContent from "../PageContent";
+import * as Options from "../../utils/options"
+import {updateSelf} from "../../modules/users/actions";
+
+const {Title} = Typography
+
+const hiddenField = {
+  position: "absolute",
+  top: -10000,
+  left: -10000,
+}
+
+const requiredRules = [{ required: true, message: 'Missing field' }]
+
+const mapStateToProps = state => ({
+  isFetching: state.users.isFetching,
+  user: state.users.itemsByID[state.auth.currentUserID],
+  error: state.users.error,
+  groups: Object.values(state.groups.itemsByID),
+});
+
+const actionCreators = {updateSelf};
+
+const ProfileContent = ({isFetching, groups, user, error, updateSelf}) => {
+  const history = useHistory();
+  const [state, setState] = useState({ message: undefined, success: undefined });
+
+
+  /*
+   * Form Data submission
+   */
+
+  const [key, setKey] = useState(1)
+  const [formData, setFormData] = useState(deserialize(user))
+  const [formErrors, setFormErrors] = useState({})
+
+  if (formData === undefined && user !== undefined && !user.isFetching) {
+    setFormData(deserialize(user))
+    setKey(key + 1)
+  }
+
+  const onValuesChange = (values) => {
+    setFormData(deserialize({ ...formData, ...values }))
+  }
+
+  const onSubmit = () => {
+    let data
+    try {
+      data = serialize(formData, user)
+    } catch(err) {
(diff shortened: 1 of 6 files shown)


frontend/src/styles/antd-adjustments.scss
```diff
@@ -25,6 +25,15 @@
   }
 }
 
+/*
+ * Forms
+ */
+
+.ant-form-item-label {
+  font-weight: bold;
```
jleblond, line 33: The change is effective for all the form labels on the app?

frontend/src/components/profile/ProfileContent.js
```diff
@@ -0,0 +1,240 @@
...
+            {/* Google Chrome is very dumb and tries to autocomplete things
+              even when we explicitely set autocomplete="new-password". Thus
+              we create fake fields, so our real fields stay pristine.
+              The tabIndex & placeholder values are filled to make it keyboard
+              inaccessible. */}
+            <input
+              type="password"
+              autoComplete="password"
+              placeholder="Fake password field"
+              style={hiddenField}
```
jleblond, line 120: I am curious about this approach... Is it necessary to the app? What caught my attention the most is the in-page css '  top: -10000, left: -10000,'

frontend/src/styles/global.css
```diff
@@ -33,8 +33,7 @@ select:-webkit-autofill:focus {
   transition: all 150ms;
 }
 .ant-layout-sider-collapsed .App__title {
-  width: 0;
-  opacity: 0;
+  transform: rotate(-90deg) translate(-60px, -10px);
```
jleblond, line 36: Haha that's cute. A suggestion, maybe having something like a circled F, which is a design often used for sidebar panels such as these, could be conceivable too :)

frontend/src/styles/antd-adjustments.scss
romgrk: Yes

frontend/src/components/profile/ProfileContent.js
romgrk: The comment explains why it's necessary.

For the inline-css, it's the way the rest of the app is styled, that I dislike but it's done for consistencty. The only places where stylesheets are used is where inline CSS cannot be used, either because the `style` attribute isn't supported or because inline styles don't allow for the required CSS features (eg, targetting a subchild when a parent has a certain calss, `.sider.expanded .title { ... }`).

frontend/src/styles/global.css
romgrk: Like the new commit?

jleblond: I noticed about your new commit that when the panel is displayed by default, the letters overlap like this;
image

Also, personally Im not a fan of rotating animation, but maybe others have different opinions about that :P

frontend/src/components/profile/ProfileContent.js
```diff
@@ -0,0 +1,240 @@
...
+              />
+            </Form.Item>
+            <Form.Item label="First Name" {...props("first_name")} rules={requiredRules}>
+              <Input />
+            </Form.Item>
+            <Form.Item label="Last Name" {...props("last_name")} rules={requiredRules}>
+              <Input />
+            </Form.Item>
+            <Form.Item label="Groups">
+              {groups.map(Options.renderGroup).map(o => <Tag key={o.value}>{o.label}</Tag>)}
```
UlysseFG, line 152: This seem to display all groups regardless of them being associated to the user.

UlysseFG (changes requested):
Very nice. looks good.

frontend/src/components/profile/ProfileContent.js
romgrk: Fixed

frontend/src/styles/global.css
romgrk: Fixed

UlysseFG:
Trying to check on the fixes but getting : 
TypeError: title is null
App.js line 197

romgrk:
Now?

UlysseFG (approved):

UlysseFG:
Yep good...

romgrk:
I should probably add tests before merging though.

jleblond (approved):
Looks great! Thanks Romain!

Yes if you have time, it'd be super appreciated if you could create integration tests for users too :)

PR #18, opened 2021-03-25
Visit single item pages in integrations test

jleblond:
This should be working fine. I am mostly looking for feedback on cypress DOM functions used, and possibly some insights on the code structure of the tests for the near future. Is it usually organized in a similar fashion to components? (i.e. having a file for Containers tests, another for Samples tests, etc.) how are the files split?

### DIFF
--- a/frontend/cypress/integration/all.spec.js
+++ b/frontend/cypress/integration/all.spec.js
@@ -27,6 +27,10 @@ const navigateTo = (section, button) => {
     cy.get('button').contains(button).click()
 }
 
+const navigateToMenuItem = (item) => {
+  cy.get('.ant-layout-sider-children').contains(item).click()
+}
+
 const submit = () =>
   cy.get('button').contains('Submit').click()
 
@@ -44,17 +48,23 @@ context('All tests', () => {
   })
 
   context('Containers', () => {
+    const singleContainerBarcode = 'test-container-add';
     it('creates single container', () => {
       navigateTo('Container', 'Add')
-      const name = 'test-container-add'
       const comment = 'This is a comment.'
-      cy.get('#name').type(name)
+      cy.get('#name').type(singleContainerBarcode)
       cy.get('#kind').click()
       cy.get('.ant-select-dropdown .ant-select-item-option').first().click()
-      cy.get('#barcode').type(name)
+      cy.get('#barcode').type(singleContainerBarcode)
       cy.get('#comment').type(comment)
       submit()
       cy.get('body').should('contain', `Container ${name}`) // Details title
+    })
+
+    it('visit container detail page', () => {
+       navigateToMenuItem('Containers')
+       cy.get('.ant-table-cell').contains(singleContainerBarcode).click()
+       cy.get('body').should('contain', `Container ${singleContainerBarcode}`)
     })
 
     it('creates multiple containers (template import)', () => {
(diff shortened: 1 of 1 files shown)


frontend/cypress/integration/all.spec.js
```diff
@@ -27,6 +27,10 @@ const navigateTo = (section, button) => {
     cy.get('button').contains(button).click()
 }
 
+const navigateToMenuItem = (item) => {
+  cy.get('.ant-layout-sider-children').contains(item).click()
+}
+
```
romgrk, line 33: How is this different than `navigateTo` with a single argument?

frontend/cypress/integration/all.spec.js
```diff
@@ -44,19 +48,25 @@ context('All tests', () => {
...
       cy.get('#kind').click()
       cy.get('.ant-select-dropdown .ant-select-item-option').first().click()
-      cy.get('#barcode').type(name)
+      cy.get('#barcode').type(singleContainerBarcode)
       cy.get('#comment').type(comment)
       submit()
       cy.get('body').should('contain', `Container ${name}`) // Details title
     })
 
+    it('visit container detail page', () => {
```
romgrk, line 64: `visits`

frontend/cypress/integration/all.spec.js
```diff
@@ -76,6 +86,13 @@ context('All tests', () => {
       cy.get('button').contains('Go Back').click()
       // cy.get('body').should('contain', '1-8 of 8 items')
     })
+
+    it('visit sample detail page', () => {
+      const sampleName = 'Sample_DNA1'
```
romgrk, line 91: Add a comment to indicate this one comes from the template

frontend/cypress/integration/all.spec.js
```diff
@@ -93,6 +110,11 @@ context('All tests', () => {
       cy.get('.ant-alert-success').should('contain', 'Template submitted')
       cy.get('button').contains('Go Back').click()
     })
+    it('visit process detail page', () => {
```
romgrk, line 113: Add a new line before `it()` for consistency with the rest.

frontend/cypress/integration/all.spec.js
```diff
@@ -93,6 +110,11 @@ context('All tests', () => {
...
+       navigateToMenuItem('Protocols')
+       cy.get('.anticon-eye').first().parent('a').click()
+       cy.get('body').should('contain', `Process`)
+    })
```
romgrk, line 117: Sidenote, we are we using a eye icon for more details on protocols, but the rest of the pages show the details when clicking on the name/id of the item? The UX isn't consistent.

romgrk (commented):
Usually there would be one file per thing tested, here we have all in `all.spec.js` because we don't test much yet but we should split it at some point.

How to define what things to separate is up to us. It could be containers/samples/protocols/etc. But splitting things is trivial so we can continue with this as long as we don't feel the need to split.

frontend/cypress/integration/all.spec.js
jleblond: This is the solution we found because in that table (in Protocols) we have 3 different models being displayed, those of Protocols, Process, and ProcessSample, while we only want to represent those to the users as 'Processes'. The eye convey additional information that is accurately representated by an item id. It could have been a magnifying glass, an eye, a text with 'More details' or so. We are aware this is not the perfect solution, but it is what we have for now. Any suggestions?

frontend/cypress/integration/all.spec.js
jleblond: Your navigateTo method requires a button, these are lists on the side, not action buttons. Any suggestions?

romgrk: The second argument is for buttons, the first only targets list items: https://github.com/c3g/freezeman/blob/dbc1ed9d06f6628bf615da4d12436a98a3fd43ca/frontend/cypress/integration/all.spec.js#L25

frontend/cypress/integration/all.spec.js
romgrk: Use eyes everywhere? I don't care either way, but it should be consistent, and at the very least not at opposite sides of the screen for different tables.

jleblond: Noted. This will be handled and discussed in a separate issue.

PR #19, opened 2021-03-25
Modification of the models and updating the containers. Samples and I…

UlysseFG:
…ndividuals not complete.


UlysseFG:
WIP : The migration script for sample and invidual is not done but should essentially be cut and paste from Container. Did not test so it may be shaky...

backend/fms_core/migrations/0018_v3_1_0.py
```diff
@@ -184,5 +198,9 @@ class Migration(migrations.Migration):
             name='created_at',
             field=models.DateTimeField(auto_now_add=True, help_text='Date the instance was created.'),
         ),
+        migrations.RunPython(
```
UlysseFG, line 201: This should not be moved there ... auto_now_add fields can't be changed once they are created. the idea is to initialize them while they are just fields with a default value, then use alterfield to activate auto_now_add  for new records.

backend/fms_core/models/imported_file.py
```diff
@@ -13,6 +14,8 @@ class ImportedFile(TrackedModel):
     filename = models.CharField(max_length=500)
     # 100 characters allocated to the path length itself
     location = models.CharField(max_length=600)
+    added = models.DateTimeField(auto_now_add=True)
```
UlysseFG, line 17: i had removed those fields since they are totally redundant with the new ones. I had not removed the rest of the linked code though but we should think of removing them nonetheless unless this breaks something we can't handle.

jleblond: it has to be there so the migration can work (to transfer the fields from aded to created_at). otherwise it doesnt find the field as references

UlysseFG: This should not be the case ... We've removed many fields in that migration including fields we access (volume_history) to extract values. We should be able to delete those field from the model and still access the fields.

PR #21, opened 2021-03-25
Implement ctrl-k shortcut to search

romgrk:


UlysseFG:
This seem to work well. 
I am not big on keyboard shortcuts, so i do not care much myself, but do you think this could annoy people used to having ctrl-k go to their browser search bar ? 
We can test it with the users and see... There might not be that many users that are used to the shortcut anyway.

romgrk:
@UlysseFG Ctrl+K is the default search shortcuts for many applications, such as slack, so people are probably already using another of the keyboard shortcuts to access their URL bar. Eg, on Chrome, pressing Ctrl+L, Alt+D or F6 will also focus the URL bar.

UlysseFG:
Cool thanks... If you are confident this won't clash with anyone's preference you got my vote ; ).

PR #23, opened 2021-04-01
link moved on processes list page. Cypress test changed.

UlysseFG:
WIP : will add the script to setup the test DB before and clear it after.


romgrk:
You're duplicating work here, we have already set up the test database/server spawning with a command runner. 

Test database setup: https://github.com/c3g/freezeman/blob/master/backend/setup_testing.sh
Test runners: https://github.com/c3g/freezeman/blob/master/package.json#L18-L19

Usage: from the root of the project, `npm run test:frontend:open` for the cypress interface or `npm run test:frontend:run` for the headlesss tests.
The scripts Jasmine and I have contributed assume a virtualenv in the `/backend` folder because both our envs are configured there.

Btw @jleblond, has the [restore_backup.sh](https://github.com/c3g/freezeman/commit/1e130ca3230c51a7e9f5328dc9475c8b3b5a4563) file been deleted? Don't see it in the tree.

jleblond:
Yes Romain it has been deleted. Ulysse had a good point when he told me that these bash scripts, dropping fms db should not be in prod. It's a file we can keep on our local machines.

romgrk:
Seems inconvenient to avoid putting files related to the project in the repo. If the file is really dangerous I'd put a git post-checkout hook to delete it on the server (`echo 'rm -f backend/restore_backup.sh' >> .git/hooks/post-checkout && chmod +x .git/hooks/post-checkout`).

UlysseFG:
I don't see these files really related to the project they are dev tools. Restore backup should never have to be run anywhere but on our local setup not even on the tests servers. Anyway ... i have a question for you on another subject. I dropped the setup from the PR. i am wondering if the decision to keep the env in the backend is motivated by some technical knowledge i might not be aware of ? I am ready to go back to the setup of QC and DEV if there is a good reason. At the moment, the setup script won't work on the servers... I also found that in order to work the script need to install npm on the root ... shouldn't we set all the env including npm (the frontend package.json) at the root ?

jleblond:
i think there are missing some cypress stuff in the individual entities .spec.js files
these are supposedly meant to be run separately (that's what Cypress offers as a possibility when running cypress visually in the browser through 'open' instead of 'run')

also, the tests are not truly independent. Protocols need to have the Sample submission template imported, in order to do manipulations on the Samples. Same goes for Samples, which require Containers creation.

UlysseFG:
yeah you are right ... this was meant as a first step toward making them modular but maybe we don't need that ... I think cypress may not be able to see the inside of the function. This need more work obviously ... Do we want to split tests up ? i thought i read this at some point in a PR ...

romgrk:
> I don't see these files really related to the project they are dev tools. Restore backup should never have to be run anywhere but on our local setup not even on the tests servers. 

Yes, dev tools are part of the project.

> Anyway ... i have a question for you on another subject. I dropped the setup from the PR. i am wondering if the decision to keep the env in the backend is motivated by some technical knowledge i might not be aware of ?

No, on my side it's for the same justification as before, I feel like dependencies should be scoped only where they are used. The virtualenv is a backend dependency.

> I also found that in order to work the script need to install npm on the root ...

Yes, the root of the project uses foreman as a task runner to spawn cypress & the server side by side.

> shouldn't we set all the env including npm (the frontend package.json) at the root ?

No, same reason as the backend. Frontend dependencies do not belong in the root of the project. The npm install at the root only installs `foreman` at the root, and the npm install in `/frontend` installs frontend dependencies only in `/frontend`.

<hr/>

Regarding the cypress thing, I think Ulysse's approach is good. Our test are not independent so they all need to be run inside `all.spec.js` together, but it's nice to be able to have them in separate files. I would keep things as Ulysse put them, and for running them don't try to run any file except `all.spec.js`.
Btw if ever you need to focus on a single test(s) while developing you can also do this to select eg a context or a it block:
```diff
diff --git a/frontend/cypress/integration/all.spec.js b/frontend/cypress/integration/all.spec.js
index 4b86395..616dbe2 100644
--- a/frontend/cypress/integration/all.spec.js
+++ b/frontend/cypress/integration/all.spec.js
@@ -43,7 +43,7 @@ context('All tests', () => {
     cy.url().should('contain', '/dashboard')
   })
 
-  context('Containers', () => {
+  context.only('Containers', () => {
     it('creates single container', () => {
       navigateTo('Container', 'Add')
       const name = 'test-container-add'
@@ -57,7 +57,7 @@ context('All tests', () => {
       cy.get('body').should('contain', `Container ${name}`) // Details title
     })
 
-    it('creates multiple containers (template import)', () => {
+    it.only('creates multiple containers (template import)', () => {
       navigateTo('Container', 'Add Containers')
       cy.get('input[type=file]').attachFile('Container_creation_v0.4.xlsx')
       submit()
```

package.json
```diff
@@ -4,20 +4,17 @@
...
     "//2": "cypress_xxx use sleep 10 to allow DB setup & frontend server to spawn",
-    "proc:cypress_open":      "cd frontend && sleep 10 && npx cypress open",
-    "proc:cypress_run":       "cd frontend && sleep 10 && npx cypress run",
-
+    "proc:cypress_open": "cd frontend && sleep 10 && npx cypress open",
+    "proc:cypress_run": "cd frontend && sleep 10 && npx cypress run",
     "//3": "This runs the backend & the frontend dev-server",
     "test:frontend:open": "nf start backend_runserver,frontend,cypress_open",
-    "test:frontend:run":  "nf start backend_runserver,frontend,cypress_run",
-
```
romgrk, line 20: Nitpick but the alignment of this file doesn't need to be changed, it's more readable the way it was originally written. Aligned characters make it easy to spot the divergences across lines.

UlysseFG:
ok i'll re-revert and add to the script to launch only the main spec. I'll also change the env structure on dev and qc server (as well as mine) to match what is done for the test runner to work.

PR #24, opened 2021-04-06
Change Sample Diff to accomodate user fields

jleblond:
I dont really like how usersByID has to be passed everywhere, but that's how I can get the username instead of user id...


romgrk:
> I dont really like how usersByID has to be passed everywhere, but that's how I can get the username instead of user id...

I know right, so unsatisfying. But we don't seem to have a choice.

LGTM, just one thing, can you assign the array `['created_by', 'updated_by']` to a constant and reuse it? Otherwise the JS engine needs to allocate & deallocate the array for each call to `renderDeltaValue()`.

PR #25, opened 2021-04-07
Add technician names

jleblond:


frontend/src/components/TrackingFieldsContent.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import {Descriptions} from "antd";
+import {withUser} from "../utils/withItem";
+
+export const TrackingFieldsContent = ({usersByID, entity}) => {
+    const displayUser = (user) => `${user.first_name} ${user.last_name} (${user.username})`
+
+    return <>
+        <Descriptions bordered={true} size="small" title="Tracking Details" style={{marginTop: "24px"}}>
+          <Descriptions.Item label="Item created by">
```
UlysseFG, line 10: Include the date or time for creation and modification ?

UlysseFG (commented):
Looks good ... i would include the time to the tracking info section...

frontend/src/components/TrackingFieldsContent.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import {Descriptions} from "antd";
+import {withUser} from "../utils/withItem";
+
+export const TrackingFieldsContent = ({usersByID, entity}) => {
+    const displayUser = (user) => `${user.first_name} ${user.last_name} (${user.username})`
```
romgrk, line 6: Move this one outside of the component, otherwise you're creating a new function instance on every render.

frontend/src/components/TrackingFieldsContent.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import {Descriptions} from "antd";
+import {withUser} from "../utils/withItem";
+
+export const TrackingFieldsContent = ({usersByID, entity}) => {
```
romgrk, line 5: Why not `connect()` this component rather than pass it `usersById` through the parent component? You're repeating the `usersById={usersById}` multiple times.

jleblond:
As discussed with Ulysse, Id rather not include the time for now. The timezone is not standardized yet so we have to come with a solution for that first. I'd rather not display the date yet, rather than having misleading information displayed.

PR #27, opened 2021-04-08
Session expiration

jleblond:
NOTE:  ******time limit to be changed to 12 hours before merging ******* (seconds used for testing purposes)


frontend/src/utils/useUserInputExpiration.js
```diff
@@ -0,0 +1,43 @@
...
+    const resetTimer = () => {
+        if (timer)
+            clearTimeout(timer);
+        startTimer()
+    };
+
+    useEffect(() => {
+        const events = [
+            'mousedown',
+            'click',
```
romgrk, line 23: Unnecessary to have both `mousedown` and `click`. `click` will fire for any device (desktop, tablet, phone).

frontend/src/utils/useUserInputExpiration.js
```diff
@@ -0,0 +1,43 @@
...
+        startTimer()
+        return () => {
+            if (timer)
+                clearTimeout(timer);
+
+            for(let i in events){
+                window.removeEventListener(events[i], resetTimer);
+            }
+        }
+    },[]);
```
romgrk, line 41: Formatting: space after `for`, space between `){`, space before `[]`

frontend/src/utils/useUserInputExpiration.js
```diff
@@ -0,0 +1,43 @@
+/*
+ * useUserInputExpiration.js
+ */
+
+import { useEffect } from 'react';
+
+export default function useUserInputExpiration(expirationFunction, expirationDelay) {
```
romgrk, line 7: I'd leave the JSDoc comment.

frontend/src/components/App.js
```diff
@@ -114,7 +114,8 @@ const App = ({userID, user, logOut, fetchInitialData, fetchAuthorizedData}) => {
 
   useEffect(onDidMount, []);
 
-  useUserInputExpiration(logOut, 10 * 1000); // 1000ms = 1 s
+  // Logout the user after 12 hours in all cases where the tab stays open
+  useUserInputExpiration(logOut, 43200 * 1000); // total 12 hours (1000ms = 1 s)
```
romgrk, line 118: I'd create a file `time.js` with
```js
export const second = 1000;
export const minute = second * 60;
export const hour = minute * 60;
export const day = hour * 24; // Slightly inaccurate
```
Then you can simply write `12 * hour`, simple & readable, no need for the comment.

backend/fms/settings.py
```diff
@@ -208,13 +208,10 @@
...
-
-SESSION_EXPIRE_AT_BROWSER_CLOSE = False
-SESSION_COOKIE_AGE = 60 * 60  # 1 hour
-
+# Mechanism in order to automatically logout the user after 4 hours
 SIMPLE_JWT = {
-'ACCESS_TOKEN_LIFETIME': timedelta(seconds=20),
-'REFRESH_TOKEN_LIFETIME': timedelta(seconds=20),
+'ACCESS_TOKEN_LIFETIME': timedelta(hours=4),
+'REFRESH_TOKEN_LIFETIME': timedelta(minutes=5),
```
romgrk, line 214: Not sure what this does, but is 5 minutes really a good delay?

frontend/src/utils/api.js
```diff
@@ -108,6 +109,7 @@ export function withToken(token, fn) {
 const ongoingRequests = {}
 
 function apiFetch(method, route, body, options = { abort: false }) {
+  refreshAuthToken()
```
romgrk, line 112: This is an async action, not a regular function. You can't call it like that.

backend/fms/settings.py
jleblond: the refresh token is used to get a new access token. even though it is usually longer-lived than the access token, for our purpose (expiring the session in the back-end) i think it should not be a long time...  if the browser is closed, we dont want to have a long refresh token lifetime.. (as we discussed, the front-end expiration is handled by the timer)

romgrk: But won't that have the effect of logging the user out 10 minutes after they close their browser tab?

PR #29, opened 2021-04-13
Logout notifications

romgrk:
Closes https://206.12.92.46/issues/367#change-1326

Test with:
```js
// directly in browser console
s = JSON.parse(localStorage['persist:auth'])
ts = JSON.parse(s.tokens)
ts.access = 'abcdef'
ts.refresh = 'abcdef'
s.tokens = JSON.stringify(ts)
localStorage['persist:auth'] = JSON.stringify(s)
```

### DIFF
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -35,4 +35,4 @@ const renderApp = () =>
     document.getElementById("root")
   );
 
-document.addEventListener("DOMContentLoaded", renderApp);
+renderApp();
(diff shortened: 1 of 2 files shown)


jleblond:
I think there is a need for being logged out if the token expires after the session is timedout, but I also think we should leave some error message to say that we were disconnected due to token expiration. I think David just meant that we should not have like 20 error messages stacking up. 
Also, I encounter a similar issue if the server is not responding (getting like 10 identical error messages)

jleblond:
PR reopened, this was initially merged to https://github.com/c3g/freezeman/pull/27 and closed. This PR will now be independent and merged eventually on master with 3.2

PR #30, opened 2021-04-13
Update Entities List and Summary after template action, entity addition or update

jleblond:


frontend/src/components/TemplateFlow.js
```diff
@@ -177,7 +180,21 @@ STEPS.UPLOAD  = 0
...
+export const mapStateToProps = state => ({
+  containersByID: state.containers.itemsByID,
+  containers: state.containers.items,
+  individualsByID: state.individuals.itemsByID,
+  individuals: state.individuals.items,
+  processes: state.processes.items,
+  processesByID: state.processes.itemsByID,
+  samples: state.samples.items,
+  samplesByID: state.samples.itemsByID,
+});
```
romgrk, line 194: No need to export these.

frontend/src/components/TemplateFlow.js
```diff
@@ -219,7 +236,8 @@ const TemplateFlow = (props) => {
     setIsSubmitting(true)
     submitRequest(actionIndex, file)
     .then(response => {
-      setSubmitResult({ valid: true })
+      setSubmitResult({ valid: true });
+      fetchInitialData();
```
romgrk, line 240: Not sure if this is the behavior you expect, but the `fetchInitialData()` promise isn't awaited on here.

PR #31, opened 2021-04-13
Merging QC into Master

jleblond:
TODO: verify tests. Newly added tests to QC clash with "c3g/interrupt-ressource-100-errors" branch that was merged in master.


backend/fms_core/resources/_generic.py
```diff
@@ -28,7 +28,14 @@ def after_save_instance(self, instance, using_transactions, dry_run):
     def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, raise_errors=False, **kwargs):
         if self.errorCount < self.ERROR_CUTOFF:
             row_result = super().import_row(row, instance_loader, using_transactions, dry_run, raise_errors, **kwargs)
-            error_count = 0 if not row_result.errors else len(row_result.errors[0])
+
+            error_count = 0
+            if row_result.errors:
+                if type(row_result.errors[0]) is list:
```
UlysseFG, line 34: i would use the pythonic if else one liner format but this might fix the errors when there is a single error (what kind of format is this that changes depending on the number of items ...)

jleblond: If there is only one error, we dont have a list, but a Error object, on which len(Error) throws an error (len is not known for Error).. I can do an if not row_result.errors:   .... else: ....    if you prefer

UlysseFG: the comment was to the person in the universe that decided that when you have only one error you get a error object instead of an array...

PR #32, opened 2021-04-15
Date format

romgrk:
Closes https://206.12.92.46/issues/373
Closes https://206.12.92.46/issues/365

Changes date rendering in the sample history timeline.

### DIFF
--- a/frontend/src/utils/humanReadableTime.js
+++ b/frontend/src/utils/humanReadableTime.js
@@ -1,0 +1,59 @@
+/*
+ * date.js
+ */
+
+const weekDays = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday']
+
+function asTime(date) {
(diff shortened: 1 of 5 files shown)


frontend/src/utils/humanReadableTime.js
```diff
@@ -0,0 +1,59 @@
...
+    return 'Yesterday at ' + asTime(date)
+
+  if (isThisWeek(date))
+    return asWeekDay(date) + ' ' + asTime(date)
+
+  return full(date)
+}
+
+export function full(date) {
+  if (isThisYear(date))
```
jleblond, line 55: I would not use that condition. If it was made this year, I would still put the year to be clear and coherent with the other dates. It was not obvious to me that the year was not being displayed because it was this year.

jleblond (approved):
Looks good! Im unsure about removing the year from the date if it is the current year, but besides that, all good!

romgrk:
Moved the condition to `compact()` only, `full()` is now always full date. Can merge now?

jleblond:
Yes it's good for me!

PR #33, opened 2021-04-19
Fixes for 3.1

jleblond:
WIP


backend/fms_core/models/process_sample.py
```diff
@@ -20,7 +20,7 @@
 class ProcessSample(TrackedModel):
     process = models.ForeignKey(Process, on_delete=models.PROTECT, related_name="process_sample", help_text="Process")
     source_sample = models.ForeignKey(Sample, on_delete=models.PROTECT, related_name="process_sample", help_text="Source Sample")
-    execution_date = models.DateField(default=timezone.now, help_text="Date of execution of the process.")
+    execution_date = models.DateField(help_text="Date of execution of the process.")
```
jleblond, line 23: here blank and Nil are at False by default.. by removing default=timezone.now, you now have to manually handle the entry of a default execution_date at each ProcessSample object creation. For now, I see that you added 'update_date' to the SampleUpdate Resource. However, it is still not in the template as a column, and will probably end up with a None value...

jleblond:
I am pushing my changes for the volume_used of a ProcessSample (for the Sample Update action). Havent modified your default execution_date changes, just merged your changes, but please take a look at my comment :)

backend/fms_core/template_paths.py
```diff
@@ -20,5 +20,5 @@
 CONTAINER_RENAME_TEMPLATE = static("submission_templates/Container_rename_v0.1.xlsx")
 SAMPLE_EXTRACTION_TEMPLATE = static("submission_templates/Extraction_v0.11.xlsx")
 SAMPLE_SUBMISSION_TEMPLATE = static("submission_templates/Sample_submission_v0.14.xlsx")
-SAMPLE_UPDATE_TEMPLATE = static("submission_templates/Sample_update_v0.5.xlsx")
+SAMPLE_UPDATE_TEMPLATE = static("submission_templates/Sample_update_v0.6.xlsx")
```
jleblond, line 23: If we are updating the templates, we should also update the csv/xlsx from our unit tests folder (fms_core/example_data) and from the front-end (cypress/fixtures) (and their links)

backend/fms_core/resources/sample.py
```diff
@@ -35,7 +35,7 @@ class SampleResource(GenericResource):
     tissue_source = Field(attribute='tissue_source', column_name='Tissue Source')
     concentration = Field(attribute='concentration', column_name='Conc. (ng/uL)', widget=DecimalWidget())
     depleted = Field(attribute='depleted', column_name='Source Depleted')
-    creation_date = Field(attribute='reception_date', column_name='Reception Date', widget=DateWidget())
+    creation_date = Field(attribute='creation_date', column_name='Reception Date', widget=DateWidget())
```
jleblond, line 38: what about this field in the single Sample creation form? 
if the user must enter it, we have to add the field manually to the form, and collect that information.

backend/fms_core/template_paths.py
UlysseFG: yeah i'll get on it...

backend/fms_core/resources/sample.py
UlysseFG: Not sure what you mean here.

PR #34, opened 2021-04-19
Forward more errors to frontend, cleanup resources, add warnings.

UlysseFG:
WIP


jleblond:
Container rename template & Container move template:
-There is a 'update comment' field in those templates, but this field is not shown in any of the container page (and we probably have the same issue with that field, such as the one we had with Sample... update_comment being replaced by the latest update_comment..)

UlysseFG:
Yeah container update_comment field will need to get a revamp too ...

backend/fms_core/migrations/0018_v3_1_0.py
```diff
@@ -270,5 +270,64 @@ class Migration(migrations.Migration):
             field=models.ForeignKey(null=False, blank=True, on_delete=models.PROTECT,
                                     related_name='fms_core_individual_modification', to=settings.AUTH_USER_MODEL),
         ),
-
+        migrations.AlterField(
```
jleblond, line 273: These are modifications to 3.1, yet this branch will be merged with master for 3.2 ?

backend/fms_core/resources/_generic.py
```diff
@@ -2,14 +2,40 @@
...
+from django.core.exceptions import ValidationError
 from reversion.models import Version
 
 
 class GenericResource(resources.ModelResource):
-    clean_model_instances = True
-    skip_unchanged = True
     ERROR_CUTOFF = 20
     errorCount = 0
+    manuallyExclude = []
```
jleblond, line 12: errorCount and manuallyExclude.. camelCase is for JS, python uses snake_case :P also I would change manuallyExclude to a more meaningful name :)

backend/fms_core/resources/_generic.py
```diff
@@ -2,14 +2,40 @@
...
 from import_export import resources
 from import_export.results import RowResult
+from django.core.exceptions import ValidationError
 from reversion.models import Version
 
 
 class GenericResource(resources.ModelResource):
-    clean_model_instances = True
-    skip_unchanged = True
     ERROR_CUTOFF = 20
```
jleblond, line 10: Would also change the variable name to something like: MAX_NB_ERRORS_ALLOWED or something like this, which would be more explicit imo

UlysseFG:
yeah this was hacked in to prevent random errors from creeping up. as for names ... yeah i come from the camel team... i will snake it up and try to figure an good name in less that 10 words...

backend/fms_core/resources/container_rename.py
```diff
@@ -52,47 +45,40 @@ def before_import(self, dataset, using_transactions, dry_run, **kwargs):
...
-        obj.clean(check_regexes=True)
-
-        if not dry_run:
-            # Append a zero-width space to the barcode / name to avoid triggering integrity errors.
-            # These will be removed after the initial import succeeds.
-            obj.barcode = obj.barcode + TEMPORARY_RENAME_SUFFIX
-            obj.name = obj.name + TEMPORARY_RENAME_SUFFIX
+        errors = {}
+
+        if not obj.id:
```
jleblond, line 50: I dont get this here?

backend/fms_core/resources/container_move.py
```diff
@@ -29,17 +30,41 @@ class Meta:
...
 
+    def import_obj(self, obj, data, dry_run):
+        errors = {}
+
+        try:
+            # Test for container barcode to provide a better error message.
+            container = Container.objects.get(barcode=get_normalized_str(data, "Container Barcode to move"))
+        except Container.DoesNotExist as e:
+            errors["container"] = ValidationError([f"Container Barcode to move was not found."], code="invalid")
+            self.manuallyExclude.extend(["kind", "name", "barcode", "location", "coordinates"])
```
jleblond, line 41: Seems more or less convenient to me to exclude all the remaining fields if the container barcode is not found. Isnt there any other way to skip the cleaning of the fields? 

Also would rename the validation error 'Inexisting source container barcode' or something more concise :)

backend/fms_core/resources/extraction.py
```diff
@@ -126,25 +124,29 @@ def import_field(self, field, obj, data, is_m2m=False):
...
-                # TODO: Currently can only extract into tubes
-                #  - otherwise this logic will fall apart
-                kind=CONTAINER_SPEC_TUBE.container_kind_id,
-                location=parent,
-                coordinates=get_normalized_str(data, "Nucleic Acid Location Coord"),
-            )
-
+        try:
+            obj.container = Container.objects.get(**shared_container_info)
+        except Container.DoesNotExist:
```
jleblond, line 148: I recommend using that approach, much more concise: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create (read the part about defaults which uses attributes only for the create)

backend/fms_core/resources/extraction.py
```diff
@@ -154,7 +156,43 @@ def import_field(self, field, obj, data, is_m2m=False):
...
+                errors["container"] = ValidationError([f"Could not create destination container. " + e.messages.pop()], code="invalid")
+                
+        try:
+            super().import_obj(obj, data, dry_run)
+        except ValidationError as e:
+            errors.update(e.update_error_dict(errors).copy())
+
+        try:
+            # Create a process for the current extraction
+            self.process = Process.objects.create(protocol=Protocol.objects.get(name="Extraction"),
```
jleblond, line 169: Would eventually be better to have constants mapped to the name of protocols, rather than using each time strings like name="Extraction"

backend/fms_core/resources/sample.py
```diff
@@ -135,106 +139,118 @@ def import_obj(self, obj, data, dry_run):
...
-            name=get_normalized_str(data, "Individual ID"),
-            sex=get_normalized_str(data, "Sex", default=Individual.SEX_UNKNOWN),
-            taxon=taxon,
-            **({"pedigree": pedigree} if pedigree else {}),
-            **({"cohort": cohort} if cohort else {}),
-            **({"mother": mother} if mother else {}),
-            **({"father": father} if father else {}),
-        )
-        obj.individual = individual
+        try:
```
jleblond, line 167: I don't really like the logic here. try get_or_create, if exception then individual_created false..  later on if individual_created is false then.. I think if you want to check whether the individual already exists you should split up here the get_or_create into a get individual, if IndividualDoesNotExist then add warning message, try create individual if exception then error individual.. Seems more straightforward to me

backend/fms_core/resources/sample.py
```diff
@@ -135,106 +139,118 @@ def import_obj(self, obj, data, dry_run):
...
-            # If needed, create a sample-holding container to store the current
-            # sample; or retrieve an existing one with the correct barcode.
-            # This will throw an error if the kind specified mismatches with an
-            # existing barcode record in the database, which serves as an
-            # ad-hoc additional validation step.
-
-            container, _ = Container.objects.get_or_create(**container_data)
-            obj.container = container
+        elif field.attribute == "sample_kind_name":
+            obj.sample_kind = SampleKind.objects.get(name=data["Sample Kind"])
```
jleblond, line 204: should we rely only on the data validation provided from the excel sheet?

backend/fms_core/resources/sample.py
```diff
@@ -135,106 +139,118 @@ def import_obj(self, obj, data, dry_run):
...
+                # within the container itself. Otherwise, it specifies the location
+                # of the container within the parent container.
+                # TODO: Ideally this should be tweaked
+
+                location_barcode = get_normalized_str(data, "Location Barcode")
+
+                try:
+                    container_parent = Container.objects.get(barcode=location_barcode)
+                except Container.DoesNotExist:
+                    if location_barcode:
```
jleblond, line 220: the logic here seems a bit over complicated or not straightforward? I would get the location_barcode out of the try/except loop. If there is no parent container and it is ok, we should not even attempt to create a container with an empty string location barcode

backend/fms_core/resources/sample.py
```diff
@@ -135,106 +139,118 @@ def import_obj(self, obj, data, dry_run):
...
+                    # Case where the container gets coordinates within the parent
+                    # (e.g. a tube in a rack).
+                    container_data["coordinates"] = normalized_coords
+
+                # If needed, create a sample-holding container to store the current
+                # sample; or retrieve an existing one with the correct barcode.
+                # This will throw an error if the kind specified mismatches with an
+                # existing barcode record in the database, which serves as an
+                # ad-hoc additional validation step.
+                container, _ = Container.objects.get_or_create(**container_data)
```
jleblond, line 250: There is no check here (try/except) whether the Container is indeed successfully created and the associated attributes are ok. In the previous fields, all that we were doing was adding the data to a container_data dictionary.

backend/fms_core/resources/transfer.py
```diff
@@ -60,116 +59,139 @@ class Meta:
...
+        if parent:
+            shared_container_info['location'] = parent
+        if get_normalized_str(data, "Destination Container Name"):
+            shared_container_info['name'] = get_normalized_str(data, "Destination Container Name")
+        if get_normalized_str(data, "Destination Container Kind"):
+            shared_container_info['kind'] = get_normalized_str(data, "Destination Container Kind")
+        if get_normalized_str(data, "Destination Parent Container Coord"):
+            shared_container_info['coordinates'] = get_normalized_str(data, "Destination Parent Container Coord")
+        try:
+            obj.container, container_created = Container.objects.get_or_create(
```
jleblond, line 127: it is not clear that container_created is a bool, just learned it from the documentation. also it is not used anywhere else, therefore i would replace it with _ otherwise it is an unused variable

backend/fms_core/resources/transfer.py
```diff
@@ -60,116 +59,139 @@ class Meta:
...
+                                                   f"[Destination Container Name, Destination Container Kind, Destination Parent Container Barcode, Destination Parent Container Coord] "], code="invalid")
+
+        try:
+            # Create a process for the current transfer
+            self.process = Process.objects.create(protocol=Protocol.objects.get(name="Transfer"),
+                                                  comment="Sample Transfer (imported from template)")
+            self.process_sample = ProcessSample.objects.create(process=self.process,
+                                                               source_sample=self.transferred_from,
+                                                               execution_date=obj.creation_date,
+                                                               volume_used=obj.volume,
```
jleblond, line 143: obj.volume is confusing, I thought by mistake you had put the volume of the obj, so I had to go check the code where it was defined. ideally the name would not be misleading :)

jleblond:
Seems good overall, I requested a few changes in the code stylistically to make things more clear. I think adding warning and errors and so, just make our resources a bit more 'messy', so I would use variable name, loops, and Python convention that make it easier for the person reading (while doing the code review, I had to constantly check the code references to see the meaning of some variables and some operations)

UlysseFG:
Most of your comments are not related directly to what was done in the branch some of them i agree with but maybe we should resolve some of them at a later date ... But maybe there is no better time than the now as would say patate poil the third... 

container_rename: obj.id not null skip the clean on those columns that rely directly on finding the source container.

container_move: When it comes to clean the out of the box solution is all or nothing ... As to the validation message, i like mine better. It uses the name of the template column to say it was not found it can hardly be more precise or concise.

extraction: The main trouble with get_or_create is that it confound all errors into a single one ... if it does not exist it creates it if it cant create it complains about what it know which is not much in this case ... In the end you have very little place to insert you error handling... for code compactness and concurrency issues but for handling errors its very poor, but again both get and create are already poor ... we'll see. 
As for the constants for protocols ... yeah this would be better that what we have.

sample: yeah you summarize my dislike of get_or_create when it come to error handling ... i was attempting to fit the new structure into the code without rewriting all of it ... the test was already there but the warning did not work outside admin... we could split the get_or_create but i am not sure it is any nicer with the get first and the create after at least in this case ... more readable for sure though...
Not sure what you mean about relying on data validation from excel. The answer is most assuredly: no we don't rely on excel validation ;) now what was the question ?
location_barcode logic ... i totally agree with you ;) this was like that and i just left it that way ... this is convoluted...
container creation without try/except this is stuff i did not get to touch... this is part of a future correction that could better deal with failure to create the container...

transfer: the container_created variable is not useful and should be removed.
obj.volume is the volume of the object, it is not misleading ... for transfer, the volume_used is the volume of the resulting sample. The obj being available here it is simpler (you don't have to clean it from the dataset).

Closing comment: "variable names are mostly only clear to the people that create them ..." another citation for patate poil the third...

PR #37, opened 2021-04-20
Use FREEZEMAN_SECRET_KEY when available

romgrk:
Closes https://206.12.92.46/issues/368

Uses $FREEZEMAN_SECRET_KEY when availabe.

@UlysseFG this will require you to update the dev/prod env to add a secret key in the environment variables.

### DIFF
--- a/backend/fms/settings.py
+++ b/backend/fms/settings.py
@@ -20,7 +20,10 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 # See https://docs.djangoproject.com/en/3.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = '*c##1@2jo)b*_jk5+rdq%4r*sst+r&vhc^43ck900h-35fb-ly'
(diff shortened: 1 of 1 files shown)


UlysseFG:
Noted ...

jleblond:
When available? Shouldnt it always be available?

romgrk:
In development.

romgrk:
It's not necessarily PROD vs DEV. Some people may wish to change the secret key in dev.

UlysseFG:
For the deployment checklist, yeah this would be an interesting thing to add. At the moment, we rely on the READMEs but it may be better to get all this information in one place.

romgrk:
Do you wanna create one? For node I have a good idea of the things to do (eg create an unprivileged nodejs user, set NODE_ENV to production, etc) but I've never put a django app in prod.

UlysseFG:
I would let you start with what you think is a good format and with the Node stuff... I'll add what i know about Django and see with IT if they can complete with the part about nginx/uwsgi. I have never installed the whole thing myself. I've been updating the code mostly, but all the server configuration has been done by IT.

PR #38, opened 2021-04-20
Interface template check

jleblond:
WIP


frontend/src/components/TemplateFlow.js
```diff
@@ -42,9 +36,43 @@ function renderResult(checkResult) {
...
+        {checkResult.error.message}
+      </pre>
+      {errors}
+    </>
+  }
+  else {
+    return <>
+      {errors}
+    </>
+  }
```
romgrk, line 51: Inline the conditional inside JSX rather than use if/else.

frontend/src/components/TemplateFlow.js
```diff
@@ -42,9 +36,43 @@ function renderResult(checkResult) {
...
+  const columns = []
+
+  checkResult.diff_headers?.forEach((diff_header, index) => {
+    columns.push({title: diff_header, dataIndex: `${index}`, key: `${index}`})
+  })
+
+  checkResult.rows.forEach((row, index) => {
+    const row_data = {}
+    row.diff.forEach((diff, diff_index) => {
+      row_data[`${diff_index}`] = stringToHTML(diff)
```
romgrk, line 65: `row_data[diff_index]`? or `row_data[String(diff_index)]`?

frontend/src/utils/stringToHTML.js
```diff
@@ -0,0 +1,5 @@
+export default function stringToHTML(str) {
+  const div = document.createElement("div");
+  div.innerHTML = str;
+  return div.innerText;
+}
```
romgrk, line 5: This looks so wrong, from a security perspective. Is it possible to get clean text from the server? If not is it possible to use simply `str.replace(/<[^>]*>/g, '')`?

Also the name shouldn't be stringToHTML, it's converting from HTML to plain text.

frontend/src/components/TemplateFlow.js
jleblond: String, I just changed it to `column-${diff_index}` to clarify what it is

PR #40, opened 2021-04-22
Tissue source optional

jleblond:
Tested it with templates and in the backend where the tissue source is optional.
It appears that it only shows as required for DNA/RNA in the front-end form.


UlysseFG (approved):
looks good...

PR #42, opened 2021-04-26
fixed validation formula in sample submission template

UlysseFG:


jleblond:
Can you check whether all the example data templates also work? (i.e. backend/fms_core/example_data, and frontend/cypress/fixtures)

UlysseFG:
Those are prefiled example ... if they pass they are not an issue since its the data validator of the field that was faulty ... I will check them but i think most of them are not even using the current template ...

jleblond:
Yup it is mostly a concern of having the same version for example templates (used for tests) and for our current templates

UlysseFG:
Mmm they are all to be changed nearly and will likely be updated again in 3.2. I would push this review to 3.2 maybe ? not sure it's work the additional delay...

PR #43, opened 2021-04-26
Template updates

UlysseFG:
- [x] Remove the xlsx templates and rename and move the csv templates for backend tests (coverage).
- [x] Rename and update the xlsx fixtures for frontend tests (cypress).


backend/fms_core/models/container.py
```diff
@@ -73,10 +73,6 @@ def add_error(field: str, error: str):
 
         self.normalize()
 
-        if check_regexes:
```
UlysseFG, line 76: Not sure this should be removed ...

UlysseFG:
Tried to take a uniform naming pattern for our test templates.
{BASE_TEMPLATE_NAME_AND_VERSION}_{[B, F] for frontend or backend}_{LETTER for each test set required}_{NUMBER for each iteration of the template when the base template do not change}.

romgrk:
Hard to see what this does, there are a lot of commits here that have already been merged on master. Maybe do a rebase or merge from master?

UlysseFG:
yeah this is fuggly... this actually came from rebasing on master. The changes i made here are only the ones about the templates and the corresponding tests.

romgrk:
Merge master into this branch then? Rebase works better for simple transactions, merge is more reliable.

frontend/cypress/integration/containers.spec.js
```diff
@@ -49,6 +49,41 @@ export const containersTests = () => {
...
+        cy.get('body').should('contain', '0-0 of 0 items')
+        cy.get('button').contains('Clear Filters').click()
+        cy.get('.ant-table-thead .ant-table-filter-trigger').eq(1).click().type(renameBarcodeDst) // filter by barcode (column 1)
+        cy.get('body').should('contain', '1-1 of 1 items')
+        cy.get('button').contains('Clear Filters').click()
+        cy.get('.ant-table-thead .ant-table-filter-trigger').eq(0).click().type(renameNameDst) // filter by name (column 0)
+        cy.get('body').should('contain', '1-1 of 1 items')
+        cy.get('button').contains('Clear Filters').click()
+      })
+
```
romgrk, line 86: Unless the variables are shared, they should be in the smallest scope required.

UlysseFG: ok... i moved them inside the inner scope.

jleblond:
The Sample Update templates (test and submission template) does not contains all the sample kind types in the Index page ?

UlysseFG:
I kind of left this fall by the wayside because there is no reference to this information on the actual template (first page). I could remove it ... maybe this would be cleaner.

jleblond (approved):
Looks good to me. Only point to mention was regarding the Sample update template (see my comment above)

jleblond:
> I kind of left this fall by the wayside because there is no reference to this information on the actual template (first page). I could remove it ... maybe this would be cleaner.

Yes, either remove it, or correct it :)

UlysseFG:
> Looks good to me. Only point to mention was regarding the Sample update template (see my comment above)

ok i removed the unrequired column from sample update.

PR #46, opened 2021-04-27
About page

jleblond:


romgrk:
I would have inserted this kind of information at build time rather than runtime. Something like [this](https://stackoverflow.com/a/46679785/6303229).

jleblond:
Is it much more advantageous having it at build time rather than runtime? your solution sounds better, but it requires installation of a package, and the commit information in info.py was a functionality already there, and already sent to the api... so that was simpler because i only had to display that information in the front-end

romgrk:
Installing one more package doesn't feel like an issue, even more given the fact that it's a build-time package. But if you're reticent to do it you can also replicate it's functionality with the standard `child_process` module:

Screenshot from 2021-04-29 11-44-51

Your solution feels easier to implement, but it adds runtime overhead. I usually prefer to avoid runtime overhead if it's not necessary.

romgrk:
I'd use `FREEZEMAN_ENV`, for consistency with `FREEZEMAN_SECRET_KEY`.

jleblond:
> I'd use `FREEZEMAN_ENV`, for consistency with `FREEZEMAN_SECRET_KEY`. And for the other variables maybe give them a common prefix `GIT_` or somehow mark them as comming from the build system. Usually the eslint notation `/* global a, b, c */` is understood by most tooling.

Good point to add 'GIT_' in front of the other variables (some of them do not come from GIT though, so I will add the freezeman prefix).. Btw for the freezeman prefix, last week I saw there are some variables set up in prod that are 'FMS_HOST' and so on... so I was thinking of keeping that... maybe we should modify the FREEZEMAN_SECRET_KEY to FMS_SECRETKEY or so ?

UlysseFG:
Yeah the current used prefix for Freezeman for env variables is FMS_ in the uwsgi config that set them on the servers.

jleblond:
> It might not be of interest to the users but i would still display these information on the PROD server since this might help track trouble with a release (if we checkout the wrong tag or forget to checkout before releasing ... unlikely but i think there is no real issue with displaying these informations).

Yeah the PR explicitely mentioned displaying only the FM version , copyright and github repo. But I agree that it might be useful to have in prod as well, and I dont think anyone will complain about it so I will change it hehe

romgrk:
Ok for the `FMS_` prefix, you can change it here for everything.

frontend/src/components/About.js
```diff
@@ -0,0 +1,65 @@
...
+                <Col {...COL_LAYOUT}>
+                    <Card title="Information" size="large">
+                      FreezeMan
+                      <br/>
+                      {(version ? `Version ${version}` : '')}
+                      <br/>
+                      Released under GNU LGPL version {version} © C3G, McGill University
+                      <br/>
+                      { REPOSITORY &&
+                            <a target='_blank' href={REPOSITORY}>{REPOSITORY} </a>
```
romgrk, line 41: Inconsistent spacing: space before `REPOSITORY &&`, and space before `</a>`.

frontend/src/components/About.js
```diff
@@ -0,0 +1,65 @@
...
+
+    return <PageContainer>
+        <AppPageHeader title="About" />
+        <PageContent style={{padding: "0 24px 24px 24px"}}>
+            <Row>
+                <Col {...COL_LAYOUT}>
+                    <Card title="Information" size="large">
+                      FreezeMan
+                      <br/>
+                      {(version ? `Version ${version}` : '')}
```
romgrk, line 36: Superfluous parenthesises, and use `{condition && value}`, not `{condition ? value : ''}`.

frontend/src/components/About.js
```diff
@@ -0,0 +1,65 @@
...
+                      <br/>
+                      Released under GNU LGPL version {version} © C3G, McGill University
```
romgrk, line 38: `{version}` comes from the freezeman version, but this is license version (`GNU LGPL v3`) and is constant.

romgrk (commented):
What about we remove `/backend/VERSION` and just set the version in `/package.json`?

frontend/src/components/About.js
romgrk: Also your link should have `rel="noopener noreferer"` [explanation](https://cccg.slack.com/archives/CCD01V37E/p1620340755045400)

romgrk:
Just switch it to read the package.json?

Something like this: (untested)
```diff
diff --git a/backend/fms/info.py b/backend/fms/info.py
index 67ebb3b..7533b24 100644
--- a/backend/fms/info.py
+++ b/backend/fms/info.py
@@ -1,5 +1,6 @@
 import subprocess
 import datetime
+import json
 from pathlib import Path
 
 __all__ = [
@@ -31,7 +32,7 @@ CONTACT_EMAIL = "info@computationalgenomics.ca"
 COPYRIGHT_YEARS = str(datetime.datetime.now().year)
 REPOSITORY = "https://github.com/c3g/freezeman_server"
 
-VERSION_PATH = Path(__file__).parent.parent / "VERSION"
+PACKAGE_PATH = Path(__file__).parent.parent.parent / "package.json"
 
-with open(VERSION_PATH, "r") as vf:
-    VERSION = vf.read().strip()
+with open(PACKAGE_PATH, "r") as packageFile:
+    VERSION = json.loads(packageFile.read()).get("version")
```

PR #47, opened 2021-04-28
Modify Test Resource for Container Rename with unique barcodes in tem…

jleblond:
…plate

### DIFF
--- a/backend/fms_core/example_data/csv/container_rename.csv
+++ b/backend/fms_core/example_data/csv/container_rename.csv
@@ -7,6 +7,5 @@ Naming Rules,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
 1,box001,box0010,original_box_2,added 0
 2,tube001,tube0010,sample1_tube_2,added 0
-3,tube002,tube0020,sample2_tube_2,added 0
+3,,,,
 4,,,,
-5,,,,
--- a/backend/fms_core/example_data/csv/container_rename.csv
+++ b/backend/fms_core/example_data/csv/container_rename.csv
@@ -5,8 +5,8 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box0001,original_box_2,added 0
-2,tube001,tube002,sample1_tube_2,swapped
-3,tube002,tube001,sample2_tube_2,swapped
+1,box001,box0001b,original_box_2,added 0
+2,tube001,tube002b,sample1_tube_2,swapped
+3,tube002,tube001b,sample2_tube_2,swapped
 4,,,,
 5,,,,
--- a/backend/fms_core/example_data/csv/container_rename.csv
+++ b/backend/fms_core/example_data/csv/container_rename.csv
@@ -5,8 +5,8 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box0001b,original_box_2,added 0
-2,tube001,tube002b,sample1_tube_2,swapped
-3,tube002,tube001b,sample2_tube_2,swapped
+1,box001,box0010,original_box_2,added 0
+2,tube001,tube0010,sample1_tube_2,added 0
+3,tube002,tube0020,sample2_tube_2,added 0
 4,,,,
 5,,,,
--- a/backend/fms_core/tests/invalid_templates/rename_invalid.csv
+++ b/backend/fms_core/tests/invalid_templates/rename_invalid.csv
@@ -5,4 +5,4 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box001$,original_box_2,dollars
+1,box001,rack001,original_box_2,dollars
--- a/backend/fms_core/tests/test_resources.py
+++ b/backend/fms_core/tests/test_resources.py
@@ -269,17 +269,17 @@ class ResourcesTestCase(TestCase):
             r = Dataset().load(rf.read())
             self.rr.import_data(r, raise_errors=True)
 
-        ci = Container.objects.get(barcode="box0001b")
+        ci = Container.objects.get(barcode="box0010")
         self.assertEqual(ci.name, "original_box_2")
         self.assertEqual(ci.update_comment, "added 0")
 
         # Samples "swapped" due to rename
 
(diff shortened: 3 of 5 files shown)


backend/fms_core/tests/invalid_templates/rename_invalid.csv
```diff
@@ -5,4 +5,4 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box001$,original_box_2,dollars
\ No newline at end of file
+1,box001,rack001,original_box_2,dollars
```
UlysseFG, line 8: i would change the test from testing $ in barcode, to testing () in name.

backend/fms_core/example_data/csv/container_rename.csv
```diff
@@ -5,8 +5,8 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box0001,original_box_2,added 0
-2,tube001,tube002,sample1_tube_2,swapped
-3,tube002,tube001,sample2_tube_2,swapped
+1,box001,box0001b,original_box_2,added 0
```
UlysseFG, line 8: I do not seen a need to change box0001 to box0001b

jleblond: It's a rename.. it could be anything like 'box342423', the only thing is that it must be a unique barcode

backend/fms_core/tests/invalid_templates/rename_invalid.csv
jleblond: box001() ? it tests for invalid names (rack001 already exists and is not the same type of container) box001() would be an acceptable name, or () are still not permitted ?

backend/fms_core/tests/test_resources.py
```diff
@@ -298,7 +298,7 @@ def test_invalid_container_rename(self):
             ("rename_invalid.csv", ValidationError),
             ("same_rename.csv", ValidationError),
             ("same_rename_2.csv", ValidationError),
-            ("double_rename.csv", ValueError),
+            ("double_rename.csv", (ValidationError, ValueError)),
```
UlysseFG, line 301: Does this ValidationError have to do with the double rename or this is to catch another unrelated error ? As far as i see the double rename still raise a ValueError.

jleblond: It will also raise a Validation error, that's why I have to catch both.

backend/fms_core/example_data/csv/container_rename.csv
```diff
@@ -5,8 +5,8 @@ Naming Rules,,,,
...
+2,tube001,tube002b,sample1_tube_2,swapped
+3,tube002,tube001b,sample2_tube_2,swapped
```
UlysseFG, line 10: this is not a useful test anymore. I would either test renaming twice the same container barcode or test that we get an error while attempting the swap ... Ideally both.

backend/fms_core/example_data/csv/container_rename.csv
UlysseFG: well it's a test that is supposed to add a 0 and now it adds a b too...

backend/fms_core/example_data/csv/container_rename.csv
jleblond: this is the only resource test that checks a container rename template that is valid.. all the other tests for container rename are invalid templates..

backend/fms_core/tests/invalid_templates/rename_invalid.csv
UlysseFG: yeah i think you are right that this is a good test ... but maybe you can test the name validator ... () in the name is still not allowed as far as i know...

backend/fms_core/example_data/csv/container_rename.csv
jleblond: right, didnt notice it. lol, maybe we should make it more visible? from box001 to box0010 ?

backend/fms_core/tests/test_resources.py
UlysseFG: Yeah but my question is should it raise a validationError ... i am not sure it should anymore than before so it might actually be testing an error with our code ... just make sure you know where this error comes from or at least ensure it's not from an undesirable behaviour.

jleblond: import throws a ValidationError if the container barcode already exists when cleaning..

UlysseFG: mmm ok so the test actually tests 2 errors... maybe we can split those error cases change the barcode of the same barcode twice and in another test give it an already existing barcode name ?

backend/fms_core/example_data/csv/container_rename.csv
UlysseFG: yeah ok ... that would actually help ;)

backend/fms_core/tests/invalid_templates/rename_invalid.csv
jleblond: () is allowed..

UlysseFG: it should not : name_validator = RegexValidator(re.compile(r"^[a-zA-Z0-9.\-_]{1,200}$"))

jleblond: we should open a different ticket regarding this. I think my PR still holds for what it is, and we could then later on add test with  () in container barcodes

UlysseFG: yeah of course ... this was just to fix the issues with the change in rename... i was merely proposing adding the test to name but if we are to change that too ... might as well wait for a final decision on the subject.

backend/fms_core/example_data/csv/container_rename.csv
```diff
@@ -5,8 +5,8 @@ Naming Rules,,,,
 ,,,,
 ,,,,
 #,Old Container Barcode,New Container Barcode,New Container Name,Update Comment
-1,box001,box0001,original_box_2,added 0
-2,tube001,tube002,sample1_tube_2,swapped
-3,tube002,tube001,sample2_tube_2,swapped
+1,box001,box0010,original_box_2,added 0
+2,tube001,tube0010,sample1_tube_2,added 0
```
UlysseFG, line 9: You can remove the 2 lines that used to test the swap if they are not testing something.

PR #49, opened 2021-04-30
Test one process per template

jleblond:


UlysseFG:
We should include a test for updates too.

jleblond:
It is there, it's just that update has only one test for both updates, didnt want to change it much, so I just added the one-line self.assertEqual(ps1.process.id, ps2.process.id)

UlysseFG (approved):
good this will help

PR #50, opened 2021-04-30
same fix to process creation done on qc

UlysseFG:


backend/VERSION
```diff
@@ -1 +1 @@
-3.1.0
+3.2.0
```
jleblond, line 1: I wouldnt change the version to 3.2 already.. those modifications are still part of 3.1.1 just to be clear and it's in QC too

jleblond (approved):
Apart from my comment on the version number, all good!

backend/VERSION
UlysseFG: This is master... it won't make its way to prod before 3.2... It was mainly not to forget (i tend to forget to update this).

jleblond: right, my suggestion was mainly to keep the commits more organized, but since it's not a critical issue, that works for me too

PR #52, opened 2021-05-03
Fix multiple processes with migration

jleblond:
Havent verified in details the data generated yet, but on a first look it should be ok. Will make more verifications tomorrow :)


backend/fms_core/migrations/0019_v3_1_2.py
```diff
@@ -0,0 +1,65 @@
...
+
+    affected_revisions_ids = Revision.objects.filter(date_created__range=["2021-04-29", "2021-05-01"]).values_list('id', flat=True)
+
+    with reversion.create_revision(manage_manually=True):
+        reversion.set_comment("Curation of multiple processes into one process per template")
+        reversion.set_user(User.objects.get(username=ADMIN_USERNAME))
+
+        for revision_id in affected_revisions_ids:
+            # Getting all ProcessSample created in this revision
+            affected_process_samples_ids = Version.objects.filter(serialized_data__contains='"model": "fms_core.processsample"',
```
UlysseFG, line 24: Likely much faster to use the id of the content_type_id/model instead of matching the serialized data to a string.

backend/fms_core/migrations/0019_v3_1_2.py
```diff
@@ -0,0 +1,65 @@
...
+                    first_process_id = min(process_ids)
+
+                    # Change process_id for ProcessSample now pointing to the wrong processes
+                    for ps in process_samples:
+                        ps.process_id = first_process_id
+                        ps.save()
+                        reversion.add_to_revision(ps)
+
+                    # Delete Processes unused (should not have been created in first place)
+                    Process.objects.filter(pk__in=process_ids).exclude(pk=first_process_id).delete()
```
UlysseFG, line 44: This is likely to call a bulk_delete, not sure this will call our delete (not saving a version for processes). I am not sure we want to save a deleted version of those processes, but we might want to since we had an initial version. It would be better for traceability too. Iterating over the processes instead to access the model delete instead of the queryset delete might work better.

backend/fms_core/migrations/0019_v3_1_2.py
jleblond: Even though it might be slightly faster (btw the request also filters the revision_id), I think it's safer to filter by serialized data content, than by giving it a fixed content_id (i.e. '6'). Also, we dont have performance issues here at all, so I dont think it's necessary.

backend/fms_core/migrations/0019_v3_1_2.py
jleblond: Right.. I was wondering whether we should version those processes to be deleted, it might be confusing when we have new valid processes with the same ids as those that were deleted.. But I guess a way to make a compromise would be to manually change the comment of those processes before deleting them manually too..

backend/fms_core/migrations/0019_v3_1_2.py
UlysseFG: you can also use the name of the model to filter using the same id (you can see references in past migrations) ... I agree that performance is not a problem here, but you should still use the most efficient way whenever you can. We tend to get used to using a certain way of doing things ... at some point it might matter.

jleblond: Referring the model name seems good to me! Just didnt want to use the plain id number hardcoded :P

UlysseFG: > content_type__model='processsample'

backend/fms_core/migrations/0019_v3_1_2.py
```diff
@@ -0,0 +1,76 @@
...
+ADMIN_USERNAME='biobankadmin'
+
+
+def fix_process_data(apps, schema_editor):
+    Version = apps.get_model("reversion", "Version")
+    Revision = apps.get_model("reversion", "Revision")
+    ProcessSample = apps.get_model("fms_core", "ProcessSample")
+    Process = apps.get_model("fms_core", "Process")
+
+    affected_revisions_ids = Revision.objects.filter(date_created__range=["2021-04-29", "2021-05-01"]).values_list('id', flat=True)
```
UlysseFG, line 16: Django complains about these dates not being timezone aware (warning), not sure if this is worth acting on it though.

backend/fms_core/migrations/0019_v3_1_2.py
```diff
@@ -0,0 +1,76 @@
+from django.conf import settings
+from django.db import migrations
+from django.contrib.auth.models import User
+import reversion
+import datetime
+
```
UlysseFG, line 6: I would add at the top a comment that explains the issue that is fixed by this data migration. The migration if applied to another DB running the same app would be useless. We may also like to get a clearer idea of what is happening here 2 years from now when we can to do some migration cleanup. Just a short comment explaining that between version 3.1 and 3.1.1 (which correspond the dates below) a bug was generating a process for each line of the submitted templates for extraction, transfer and update.

UlysseFG (commented):
I checked the results and compared them to the templates we got from the lab. Everything seem to fit. Just added a few details to check.

backend/fms_core/migrations/0019_v3_1_2.py
```diff
@@ -0,0 +1,76 @@
...
+                    # Change process_id for ProcessSample now pointing to the wrong processes
+                    for ps in process_samples:
+                        ps.process_id = first_process_id
+                        ps.save()
+                        reversion.add_to_revision(ps)
+
+                    # Delete Processes unused (should not have been created in first place)
+                    for process in processes_to_delete:
+                        process.comment = f'DELETE in curation {datetime.datetime.now().strftime("%Y-%m-%d")} v3.1.2 combined multiple processes - ' + process.comment
```
UlysseFG, line 48: DELETED instead of DELETE

UlysseFG, line 48: Set the delete flag on process to true before saving. The version created is not flagged as deleted.

backend/fms_core/migrations/0019_v3_1_2.py
jleblond: Dont think it's worth the trouble of making it timezone aware. The date filter does not have to be super precise, it was mostly just to not have to go through all the ever existing revisions :P  and it succeeds at filtering the right data  :)

UlysseFG (approved):
This looks good. Can you edit the VERSION file to set it to 3.1.2 as a last request ? Do not forget to make a corresponding PR to update master.

PR #53, opened 2021-05-04
Container hierarchy issues

romgrk:
Closes https://206.12.92.46/issues/346

Fixes UX issues with the container hierarchy component.

Issues:
 - [x] Clicking on the container expand button should list the number of containers or samples, plus the list, right away.
 - [x] Children containers should be organized by container coordinate (e.g. A01, A02, ..., A12, B01, B02, etc.)
 - [x] It's good to have to click to get the list of other containers in the parent container, but we should have to click only once. (currently twice)

Related:
 - https://github.com/ant-design/ant-design/issues/26926


romgrk:
Third issue is caused by ant design issue and might not be solvable here, might require fixing upstream.

frontend/src/components/containers/ContainerHierarchy.js
```diff
@@ -102,31 +127,43 @@ const buildContainerTreeFromPath = (context, path) => {
...
+          title: renderEntry(
             <Link to={`/samples/${sample.id}`}>
               <strong>{sample.name}</strong>{' '}
               sample ({sampleKind})
             </Link>
-          </span>,
+          ),
           icon: <CheckOutlined />,
-          key: sampleId,
+          key: `sample-${sampleId}`,
```
UlysseFG, line 166: This breaks the link to the sample when clicked directly (which uses this key instead of the link defined above). right-click  open in new tab works.

frontend/src/components/containers/ContainerHierarchy.js
```diff
@@ -102,31 +127,43 @@ const buildContainerTreeFromPath = (context, path) => {
...
           type: 'sample',
```
UlysseFG, line 167: Maybe this could be the PR to address the ctrl-click problem in the tree. I do not know if it's even possible, but at the moment we have the same information duplicated on the container page for the sole purpose of being able to ctrl-click links. The sample(s) listed in content are also listed in the tree below. I think we could get rid of the content section if we could allow ctrl-click to work in the tree.

UlysseFG: Also, i think the samples listed in the tree should have their coord listed. This may require a lot of loading but i think this information is likely to be desirable.

frontend/src/components/containers/ContainerHierarchy.js
jleblond: As Ulysse said it breaks the link (the link becomes /sample/sample-23 instead of /sample/23), but I dont think the key should be used to generate the url anyways.. The key is a React element that defines a unique HTML DOM element. Could that be fixed without using directly the key?

frontend/src/components/containers/ContainerHierarchy.js
jleblond: Totally agree with Ulysse on these points :P

jleblond (changes requested):
There is the bug with the link as mentioned above, and displaying the coordinates of the Samples in a Container Hierarchy would be good to have too

romgrk:
All changes have been implemented. The Contents section has been removed, the hierarchy links now allow to open items in new tabs.

UlysseFG:
Very nice... There seem to be an issue remaining with the ordering of the samples in the tree when you have a plate instead of a tube-rack. For samples in plates the coordinate of interest is at the level of the sample (coordinates attribute). This field is left empty for sample in tubes (for which the coordinates of interest are on the container).

frontend/src/utils/platform.js
```diff
@@ -0,0 +1,65 @@
+/*
+ * platform.js
```
romgrk, line 2: If ever you need, this file adds useful stuff for platform specific features.
 - `platform().isTouch` tells you if the platform has touch-only interactions (mobile, tablet). This can be useful to know if onMouseOver behaviors should be implemented otherwise, eg show tooltips when the user holds an element pressed rather than when the user hover the element.
 - body.className gets classes for OS & type, eg `.platform-linux .platform-desktop`; we can also add the browser if needed, eg when a browser needs specific styling corrections.

PR #54, opened 2021-05-04
Forgot password

romgrk:
Closes https://206.12.92.46/issues/371

Implements a password reset feature.


romgrk:
This is ready for review.

If `FMS_EMAIL_HOST` is not set, the password-reset emails will be logged to the console. If you want to test this with real emails, export these variables before running django:

```
FMS_EMAIL_HOST=smtp.office365.com
FMS_EMAIL_PORT=587
FMS_EMAIL_FROM=romain.gregoire@mcgill.ca 
FMS_EMAIL_USER=romain.gregoire@mcgill.ca 
FMS_EMAIL_PASSWORD=secret
FMS_EMAIL_TLS=False
```

You will need to install the new django dependency and run `python manage.py migrate`.

This PR introduces many new environement variables. I've documented them in `/backend/example.uwsgi.ini`.

UlysseFG:
From what i understand this still leaves us in need of an email account (or server) to serve these emails.

romgrk:
Indeed, but that's independent from this PR. Is this OK for merging?

jleblond:
I would add the Freezeman Logo to the new homepage. Or something just to identify clearly the app :) 

Otherwise, seems ok. I did some tests on my end, and its not sending the email though just printing it in the console? (I installed the back-end package, setup the environment variables)

UlysseFG:
I like the new login page look, but i have still to find way to have it send me emails. We should first confirm this works with the email setup we have. I used the account provided by David last week. I think google needs TLS, but even when activating TLS i do not receive emails. I don't know if its the firewall at the center that blocks them or googles anti-spam filters that catch this, but we should resolve this before integrating this.

romgrk:
Email sending works for me with my mcgill.ca account. To debug your installation, I'd need details. If it's still printing to the console, it's likely that `FMS_EMAIL_HOST` is empty; I'd recommend to make sure your environment variables are accessible to the python app.

I can try to test with the account from David, send me the credentials through slack I'll test it.

UlysseFG:
ok after a few stumbling around this works ... The little eye to see the password you enter would be sufficient for a work from home app (where no one is peeping at your screen), but it might be safer to add the confirmation field to make sure people can reset their password without having to display it.

romgrk:
If the user messes their password, they can just reset their password again, it's not very hard, and it also optimizes the happy path for most users. Can't remember where but I remember seeing a few big websites using this pattern and just skip the password confirmation. For most users, the password is also stored by the browser so it doesn't matter much.

UlysseFG:
eheh happy path ;) Ok well you are right that the user can reset it 10 times if they can't type (like me).

UlysseFG (approved):

PR #55, opened 2021-05-04
Processes in SampleDetail page

jleblond:


frontend/src/utils/withItem.js
```diff
@@ -80,9 +82,12 @@ export const withContainer = createWithItem(Container, api.containers)
 export const withSample = createWithItem(Sample, api.samples)
 export const withIndividual = createWithItem(Individual, api.individuals)
 export const withUser = createWithItem(User, api.users)
+export const withProcess = createWithItem(Process, api.processes)
+export const withProtocol = createWithItem(Protocol, api.protocols)
```
UlysseFG, line 86: Not sure this will ever be useful since we load all protocols each time we reload the site... If you see a use for it, you could add it to the export below too.

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -196,6 +210,26 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, ind
           </div>
         </Col>
       </Row>
+
+       <Title level={2} style={{ marginTop: '1em' }}>Processes</Title>
```
UlysseFG, line 214: i would put this above the version section if this cannot go side by side with it (i agree with the argument of the monitor size but this is starting to be a long page ;) ) ... I think this information is of more interest for the lab during day to day operation than the versions... The version section could also push this down and people may not realize it is there.

frontend/src/components/samples/SampleDetailProcess.js
```diff
@@ -0,0 +1,77 @@
...
+
+const SampleDetailProcess = ({processes, usersByID, protocolsByID,}) => {
+    const columns = [
+      {
+        title: '',
+        dataIndex: 'id',
+        key: 'id',
+        render: (id, _) =>
+          <Link to={`/processes/${id}`}>
+            Process #{id}
```
UlysseFG, line 23: The process ID provided to the user is the process model id, the process_sample model id is kept internal (you use it to link to the right processes, but do not display that id) and it is not provided to the user since it has no use to the user and would just confuse him (what is the difference with process and process_sample what is that id is it process or process_sample ?).

jleblond: We are back to the argument of how do we identify a 'process-sample' . I thought on your end you decided to name it Process #23424  as a mean to represent ProcessSample #23424. As I have said, this might get confusing very quickly, so we better figure out a way to represent those concepts in our interface, and not confuse it with our model names :)

frontend/src/components/samples/SampleDetailProcess.js
```diff
@@ -0,0 +1,77 @@
...
+          </Link>,
+      },
+      {
+        title: 'Execution Date',
```
UlysseFG, line 27: Just to confuse you some more ... I would call it "Date Processed", like in the processSampleList. Execution Date does not sound as "processy".

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -73,6 +77,15 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, ind
   if (isLoaded && !sample.versions && !sample.isFetching)
     listVersions(sample.id);
 
+  if (isLoaded && !isProcessesEmpty && !sample.isFetching) {
+    sample.processes_samples.forEach((id, i) => {
+      withProcess(processesByID, id, process => process.id);
+    })
+    sample.processes_samples.forEach((id, i) => {
+      sampleProcesses.push(processesByID[id])
+    })
```
romgrk, line 86: `const sampleProcesses = sample.processes_samples.map(id => processesByID[id])`?

Btw why is that thing called `processes_samples`? I don't understand the double plural.

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -164,6 +177,25 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, ind
...
+      <Title level={2} style={{ marginTop: '1em' }}>Processes</Title>
+      <Row>
+        <Col sm={24} md={24}>
+          <div>
+            <Card>
+              {
+                isProcessesEmpty ?
+                  <Empty image={Empty.PRESENTED_IMAGE_SIMPLE} /> :
+                  <div>
+                      {sampleProcesses &&
```
romgrk, line 189: Why `sampleProcesses && ...`? First, that thing is an array so it's always truthy. Second, aren't you already checking that it's not empty?

frontend/src/components/samples/SamplesDetailContent.js
jleblond: Its the ProcessSample objects associated to a Sample. The model name is ProcessSample.. I guess I could just rename it 'ProcessSamples' if you prefer

romgrk: But are those things processes or samples? If those are processes I'd just call them processes.

jleblond: There are 3 models: Process, Sample and ProcessSample. Here we want to represent ProcessSample objects. ( I know this is confusing, even now with the Process page listing rows of ProcessSample, thats definitely an issue imo ...)

frontend/src/components/samples/SampleDetailProcess.js
UlysseFG: This is more of a stopgap solution. The idea is to eventually have this process ID lead to a process"batch" ID page where you can see the meta-data at that level. There is no real meta-data there at the moment and since we have not found/decided how to link the sample level data this is as good a solution as any without introducing a new "entity" which is not really one ;) (processSample). The eye / "details..." / repr of processSample still has my vote.

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -67,8 +67,9 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, pro
   const experimentalGroups = sample.experimental_group || [];
   const versions = sample.versions;
   const isVersionsEmpty = versions && versions.length === 0;
-  const isProcessesEmpty = sample.processes_samples && sample.processes_samples.length === 0;
-  let sampleProcesses = []
+  const sampleProcessSamples = sample.process_samples
+  const isProcessesEmpty = sampleProcessSamples && sampleProcessSamples.length === 0;
+  let processSamples = []
```
romgrk, line 72: `const`

frontend/src/components/samples/SamplesDetailContent.js
```diff
@@ -78,11 +79,11 @@ const SamplesDetailContent = ({samplesByID, sampleKindsByID, containersByID, pro
...
 
   if (isLoaded && !isProcessesEmpty && !sample.isFetching) {
-    sample.processes_samples.forEach((id, i) => {
+    sampleProcessSamples.forEach((id, i) => {
       withProcess(processesByID, id, process => process.id);
     })
-    sample.processes_samples.forEach((id, i) => {
-      sampleProcesses.push(processesByID[id])
+    sampleProcessSamples.forEach((id, i) => {
+      processSamples.push(processesByID[id])
```
romgrk, line 86: Loop once and do both tasks.

Can this put `undefined`s in `processSamples` if the process data hasn't been completed yet?

PR #58, opened 2021-05-10
Curation entity attribute

UlysseFG:


jleblond:
Took a quick look, and I would recommend being able to do these in 'batches' under a Revision. Would be easier to track :)

UlysseFG:
This is exactly that ;)...

romgrk:
Can you guys always include a sentence to describe your PRs? I read them as much as possible to keep up with the changes in the backend but sometimes I don't have enough context to understand what it does or why it's needed, such as this one.

PR #60, opened 2021-05-12
New curation script to delete individuals.

UlysseFG:
This is a curation script that deals with individual deletions. It confirms the individuals are not linked as parents to another individual and they do not have samples linked to them. This PR also move the actual deletion outside the Revision block to make sure a last version of the deleted item is saved.

### DIFF
--- a/backend/fms_core/management/commands/_delete_individual.py
+++ b/backend/fms_core/management/commands/_delete_individual.py
@@ -63,10 +63,7 @@ def delete_individual(params, objects_to_delete, log):
                 else:
                     log.info(f"Deleted [Individual] name [{individual.name}] id [{individual.id}]].")
                     individual.deleted = True
-                    if user_id:
-                        individual.save(requester_id=user_id) # save using the id of the requester
-                    else:
-                        individual.save()# save using the default admin user
+                    individual.save(requester_id=user_id) # save using the id of the requester (using the default admin user if None)
                     objects_to_delete.append(individual)  # Delay deletion until after the revision block so the object get a version
                     count_deleted += 1
             except individual_model.DoesNotExist:
--- a/backend/fms_core/management/commands/_delete_individual.py
+++ b/backend/fms_core/management/commands/_delete_individual.py
@@ -1,0 +1,86 @@
+from django.apps import apps
+import reversion
+import json
+import logging
+from reversion.models import Version
+from fms_core.models import Individual
+
+# Parameters required for this curation
+ACTION = "action"                     # = delete_individual
+CURATION_INDEX = "curation_index"     # Number indicating the order in which this action was performed during the curation.
+COMMENT = "comment"                   # An optional comment to be stored in the logs
+INDIVIDUAL_NAMES = "individual_names" # An array of individual name that are to be deleted.
+USER_ID = "requester_user_id"         # The user id of the person requesting the curation. Optional. If left empty, uses biobankadmin id.
+
+# Curation params template
+# { CURATION_INDEX: 1,
+#   ACTION: "delete_individual",
+#   COMMENT: "Dr. No asked to delete these individuals that are not used.",
+#   INDIVIDUAL_NAMES: ["Roger", "TiPaul", "Gertrude"], # List of individual names (unique) to be deleted
+#   USER_ID: 5
+# }
+
+# This curation deletes only individual that are not referenced. Remove individual from samples references using "update_field_value" first.
+# To delete parents, a first curation need to be run to delete the child individual (or an update), then parents will be unreferenced.
+
+# function that checks the references to an individual and list them.
+def references(individual):
+    links = []
+    children = list(individual.mother_of.all())
+    if children:
+        links.append(f"Mother of {[child.name for child in children]}.")
+        children = None
+    children = list(individual.father_of.all())
+    if children:
+        links.append(f"Father of {[child.name for child in children]}.")
+        children = None
+    samples = list(individual.samples.all())
+    if samples:
+        links.append(f"Has samples {[sample.id for sample in samples]}")
+    return links
+
+
+def delete_individual(params, objects_to_delete, log):
+    log.info("Action [" + str(params[CURATION_INDEX]) + "] Update Field Value started.")
+    log.info("Comment [" + str(params.get(COMMENT, "None")) + "].")
+    log.info("Individual names : " + str(params[INDIVIDUAL_NAMES]) + ".")
+    log.info("Requester id : " + str(params.get(USER_ID)))
+
+    # initialize the curation
+    curation_code = params.get(CURATION_INDEX, "Invalid index")
+    error_found = False
+    name_array = params[INDIVIDUAL_NAMES]
+    user_id = params.get(USER_ID)
+
+    try:
+        individual_model = apps.get_model("fms_core", "Individual")
+        count_deletes = 0
+        for name in name_array:
+            try:
+                individual = individual_model.objects.get(name=name)
+                links = references(individual)
+                if links:
+                    log.error(f"Individual [{name}] is still referenced. {links}")
+                    error_found = True
+                else:
+                    log.info(f"Deleted [Individual] name [{individual.name}] id [{individual.id}]].")
+                    individual.deleted = True
+                    if user_id:
+                        individual.save(requester_id=user_id) # save using the id of the requester
+                    else:
+                        individual.save()# save using the default admin user
+                    objects_to_delete.append(individual)  # Delay deletion until after the revision block so the object get a version
+                    count_deletes += 1
+            except individual_model.DoesNotExist:
+                log.error(f"No individual found for name [{name}].")
+                error_found = True
+            except individual_model.MultipleObjectsReturned:
+                log.error(f"Multiple individuals found for name [{name}]. Provide a unique identifier.")
+                error_found = True
+    except LookupError:
+        log.error("Model [Individual] does not exist.")
+        error_found = True
+    if not error_found:
+        curation_code = None
+        log.info(f"Deleted [{count_deletes}] individuals.")
+    return curation_code
--- a/backend/fms_core/management/commands/_delete_individual.py
+++ b/backend/fms_core/management/commands/_delete_individual.py
@@ -24,16 +24,14 @@ USER_ID = "requester_user_id"         # The user id of the person requesting the
 # To delete parents, a first curation need to be run to delete the child individual (or an update), then parents will be unreferenced.
 
 # function that checks the references to an individual and list them.
-def references(individual):
+def list_references(individual):
     links = []
-    children = list(individual.mother_of.all())
-    if children:
-        links.append(f"Mother of {[child.name for child in children]}.")
-        children = None
-    children = list(individual.father_of.all())
-    if children:
-        links.append(f"Father of {[child.name for child in children]}.")
-        children = None
+    mother_of = list(individual.mother_of.all())
+    if mother_of:
+        links.append(f"Mother of {[child.name for child in mother_of]}.")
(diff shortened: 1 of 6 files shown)


jleblond:
Some of the individuals in your curation JSON file are still referenced, so the transaction was rolled back.

UlysseFG:
> Some of the individuals in your curation JSON file are still referenced, so the transaction was rolled back.

Yeah this is the intended behaviour. The deletion would be blocked by the DB anyway. You will have to play with the DB and the param file to test a working condition.

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+#   COMMENT: "Dr. No asked to delete these individuals that are not used.",
+#   INDIVIDUAL_NAMES: ["Roger", "TiPaul", "Gertrude"], # List of individual names (unique) to be deleted
+#   USER_ID: 5
+# }
+
+# This curation deletes only individual that are not referenced. Remove individual from samples references using "update_field_value" first.
+# To delete parents, a first curation need to be run to delete the child individual (or an update), then parents will be unreferenced.
+
+# function that checks the references to an individual and list them.
+def references(individual):
```
jleblond, line 27: Most of my comments will be more stylistic/naming/performance related :P they are not bugs but would  improve the style 

references is too vague for a method name

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+    links = []
```
jleblond, line 28: Can you use a more concise logic? reassigning the children variable and rechecking it for both children as mother and father does not look straightforward to me. 

Also, I'd make that a model method that lists the children samples, and another method that collects the samples from the collection of X individuals.

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # initialize the curation
+    curation_code = params.get(CURATION_INDEX, "Invalid index")
+    error_found = False
+    name_array = params[INDIVIDUAL_NAMES]
+    user_id = params.get(USER_ID)
+
+    try:
+        individual_model = apps.get_model("fms_core", "Individual")
+        count_deletes = 0
```
jleblond, line 57: count_deleted ?

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+        for name in name_array:
+            try:
+                individual = individual_model.objects.get(name=name)
+                links = references(individual)
+                if links:
+                    log.error(f"Individual [{name}] is still referenced. {links}")
+                    error_found = True
```
jleblond, line 64: Not sure about using a boolean to stop the transaction/curation.. shouldnt an Error be raised and caught?

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+                else:
+                    log.info(f"Deleted [Individual] name [{individual.name}] id [{individual.id}]].")
+                    individual.deleted = True
+                    if user_id:
+                        individual.save(requester_id=user_id) # save using the id of the requester
+                    else:
```
jleblond, line 70: This logic of verifying whether there is a requester user_id or using the default admin, should not be part of the curation, it should be in the model or TrackedModel. I have seen that there is something there that handles that (might need few improvements)

backend/fms_core/management/commands/_delete_individual.py
```diff
@@ -0,0 +1,86 @@
...
+                log.error(f"No individual found for name [{name}].")
+                error_found = True
+            except individual_model.MultipleObjectsReturned:
+                log.error(f"Multiple individuals found for name [{name}]. Provide a unique identifier.")
+                error_found = True
+    except LookupError:
+        log.error("Model [Individual] does not exist.")
+        error_found = True
+    if not error_found:
+        curation_code = None
```
jleblond, line 84: This also refers to my comment about error handling. It's better not using boolean and curation_index as a way to check if it generates errors..

backend/fms_core/models/sample.py
```diff
@@ -240,11 +240,11 @@ def source_depleted(self) -> bool:
         return self.extracted_from.depleted if self.extracted_from else None
 
     @property
-    def extracted_from(self) -> ["Sample"]:
+    def extracted_from(self) -> "Sample":
```
jleblond, line 243: I think at some point you were talking about referring to multiple samples (array)? Did you change your mind on that?

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: Nah this is part of the curation and the tracked model code supporting this was added for the curation scripts. The model do not care who did what ...

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: This is a way to support both exceptions and "normal" errors ... I will not throw exceptions or define custom errors for all the case where some condition raise an error. This also let us manage error much more nimbly than the exception handling approach you suggest that basically crashes your application at the first error encountered. The way around this will require in any case that kind of error handling ...

backend/fms_core/management/commands/_delete_individual.py
jleblond: I mean the if user_id individual.save(... user_id) else  individual.save()

backend/fms_core/models/sample.py
UlysseFG: Nope i just removed that so it would match the first() in the function code... if (when) we change first() for all() we'll add back the [ ].

backend/fms_core/management/commands/_delete_individual.py
jleblond: With your current logic by passing an argument, you should handle the general cases where there is no requester_id provided, in the TrackedModel

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: I want to be able to process all curations and log the result for each before interrupting processing. Raising exceptions would just force me to catch them and store them in a variable ... so i can continue processing the rest...

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: potato potéto to me ;) i'll change it...

backend/fms_core/management/commands/_delete_individual.py
jleblond: You can raise an exception and catch it by adding a specific message to the console.. you do not have to interrupt the flow of the whole process

backend/fms_core/models/sample.py
jleblond: I am surprised that we can change the return type without much consequences on the rest of our app.. As long as it does not have side effects...

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: list_references then... ? I do not find it vague personally. Give me a better name if you have one.

backend/fms_core/management/commands/_delete_individual.py
jleblond: you were the one who asked me to change DELETE TO DELETED in a comment message :P same should go for variable names ;P

backend/fms_core/management/commands/_delete_individual.py
jleblond: catch => printing ?  anyways you are doing this in an atomic transaction

backend/fms_core/management/commands/_delete_individual.py
jleblond: entities_references or reference_entities or references_to_entities (or something like that) it just doesnt say reference for/from what? 

also for code modularity and responsability it should return what it is (the obj themselves)

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: I hardly see this as a useful model method. This is very specific to this curation needs (the fact that it return human readable references for the logs). Maybe we could think of an alternate version that returns a list in a less human readable form for a method. 
As for being concise, i could save a couple lines by using different variables. The idea here is to give a useful log information. Separating mother and father helps to identify the error. Not sure i can save much there ...

jleblond: Just a suggestion, but something like this will be much more performant querywise and memory-wise, especially if we are to have lot of individuals (millions in the future?)

    reference_logs = []
    if individual.mother_of.count() > 0:
        reference_logs.append(f"Mother of {list(individual.mother_of.values_list('id', flat=True))}.")

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: So you catch an error you throw it up and catch it again right away to perform the same operation ? if you do not raise the error in the end the transaction will complete ... so if you catch => print you have to raise again or store it in a variable and raise later ... error handling with exception handling is not a great way to handle testable cases in my opinion.

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: if you do that your transaction will complete and you'll have a partial curation ...

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: Eheh this was English syntax correction which have much more clear cut rules than variable naming ;) But i agree that it was of very modest importance ;).

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: def list_references_to(individual) ... how does it sound ?

backend/fms_core/management/commands/_delete_individual.py
jleblond: alright i see this specific context you still want to go through all the items, display errors and all.. the boolean didnt sound the way to go to me, but I can live with it :P

UlysseFG: eheh it is not cute but it holds the "error state" across multiple errors without having to juggle all of them.

backend/fms_core/management/commands/_delete_individual.py
jleblond: better :P

backend/fms_core/management/commands/_delete_individual.py
jleblond: Code is self-documenting :P variable names ARE in some way important hehe

backend/fms_core/management/commands/_delete_individual.py
UlysseFG: Not sure this would save much of anything ... unless we have very fertile mothers ... ;)

jleblond: right.. lol an individual should not have many children individuals or many samples... ok.. lol

UlysseFG: It must admit i am not very up to date with the performance tweaking of those inline string builders. The way python convert that string list into a single string may be ineffective, but it might not be worth using less readable code for those curations that will always stay within a reasonable scope. If we have a blanket curation to do on all samples or individuals we'd use a migration or at least design a different script (one that would not list each individual to delete, but instead the condition that need to be met for deletion).

jleblond (approved):

PR #61, opened 2021-05-12
Rename frontend Process to ProcessSample

jleblond:
in the code variables/names only. the interface text remains the same for the user

I tested it manually, but double-check would be welcome since it involves lots of variables and links renaming


frontend/src/components/ActionContent.js
```diff
@@ -20,13 +20,13 @@ const LOADING_ACTION = {
 const checkRequests = {
   sample:    api.samples.template.check,
   container: api.containers.template.check,
-  process:   api.processes.template.check,
+  processSample:   api.processesSamples.template.check,
```
romgrk, line 23: Double plural?

jleblond: We can have it single plural (processSamples) if you prefer. I wanted it to be coherent with our API (processes-samples) but if needed, I can change that too, its not too big

jleblond:
Closed. Will create another PR to rename it to another name in the front-end and back-end

PR #65, opened 2021-05-13
Handle JSON fields in Sample Versions Diff

jleblond:
This is mainly to accomodate the use of the 'Experimental group' field which is a JSON field. It would not render properly in the Sample Detail page, in the Versions box. 

PS: @UlysseFG  I could open another PR to have it merged to QC too if you want..

### DIFF
--- a/frontend/src/utils/renderSampleDiff.js
+++ b/frontend/src/utils/renderSampleDiff.js
@@ -28,7 +28,7 @@ export default function renderSampleDiff(oldVersion, newVersion, usersByID) {
   const items = Object.entries(deltas).map(([key, delta]) => {
     if (Array.isArray(delta))
       return renderArrayDelta(key, delta, oldVersion, newVersion, usersByID);
-    else if(delta.constructor == Object)
+    else if (delta.constructor == Object)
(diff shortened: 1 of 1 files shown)


frontend/src/utils/renderSampleDiff.js
```diff
@@ -25,10 +25,11 @@ export default function renderSampleDiff(oldVersion, newVersion, usersByID) {
   if (deltas === undefined)
     return null;
 
-
   const items = Object.entries(deltas).map(([key, delta]) => {
     if (Array.isArray(delta))
       return renderArrayDelta(key, delta, oldVersion, newVersion, usersByID);
+    else if(delta.constructor == Object)
```
romgrk, line 31: Space after if.

romgrk (approved):

UlysseFG:
I do not think we need another PR. We can try to merge these PR in master then move master over to QC. I'll check this in the morning...

jleblond:
From my understanding they will start using the Experimental groups very soon, and it wont be production-ready. But it's only a display bug, so maybe it can go to 3.2

PR #70, opened 2021-05-19
Filter with batch values

jleblond:
Closes https://206.12.92.46/issues/400

Allows to filter some commons fields (such as container barcode, container name, sample name, individual name) with multiple values in a free text filter.

i.e test for barcodes "EQ00422043,EQ00422044"
or for sample names: "CHUM-2015201670A, RIM-8143272302"


romgrk:
When there is a linked issue, copy paste the actual link, it's faster for us.

The PR seems fine, haven't tested it but do whitespaces affect the search results? If so I'd trim them.

jleblond:
No white spaces dont affect it

romgrk (approved):

frontend/src/utils/serializeFilterParams.js
```diff
@@ -38,11 +38,19 @@ export default function serializeFilterParams(filters, descriptions) {
 
       case FILTER_TYPE.INPUT: {
         const options = filters[field].options
-        key = (options && options.exactMatch) ? (key + "__startswith") : (key + "__icontains")
+
+        if (value.includes(',')) {
```
UlysseFG, line 42: what about fields that can contains ',' ?

frontend/src/utils/serializeFilterParams.js
```diff
@@ -38,11 +38,19 @@ export default function serializeFilterParams(filters, descriptions) {
...
+          key += "__in"
+        } else if (options?.exactMatch) {
+          key += "__startswith"
```
UlysseFG, line 45: We need at some point down the line to have an exact match that actually do what is advertised ... eventually people will complain about that (e.g. i want sample juice_1 and i keep getting juice_12 too!).

jleblond: That was discussed previously in the implementation of the exact match with Romain. Apparently those kind of 'exact matches' are rather starts_with than exact =.
The feature was approved, and to be honest I wouldnt replace it with a '=' neither... If you strongly disagree we can discuss this in private, or at a team meeting.

jleblond: I can understand your point, but we shouldnt make a decision here :P let's bring it back to a team discussion (with David)

frontend/src/utils/serializeFilterParams.js
```diff
@@ -39,20 +39,25 @@ export default function serializeFilterParams(filters, descriptions) {
...
+        if (value.includes(',')) {
+          key += "__in"
+        }
+        else if (options) {
+            if (options.recursiveMatch)
+                key += "__recursive"
+            else if (options.exactMatch)
+                key += "__startswith"
+            else
+                key += "__icontains"
```
romgrk, line 51: Inconsistent indentation here

sebastianballesteros:
Ignore the first commits. I made them a few weeks ago when I was starting to play with it and I had to commit them cause otherwise they would mess up other branches in my local changes (my bad).

jleblond:
Would there be a way for you to close this previous PR, rebase your branch (or merge), and re-open a new one? That'd be easier for review, there is a bunch of unrelated changes here when I look at 'Files changed'.

On a quick look it looks like a great solution you found! I am wondering if there is any way for you to make it a bit more modular. Those methods that parse the query and split it, they seem very similar. Also, does it still work with the exact match and recursive match options?

sebastianballesteros:
Yes, I think if I rebase and open a new one will be easier for you to to review it. I actually tried to made it modular by referencing a single method outside the classes but Django complained. Let me see what I can do.

backend/fms_core/filters.py
```diff
@@ -0,0 +1,58 @@
...
+    name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+    class Meta:
+        model = Container
+        fields = _container_filterset_fields
+
+    def batch_barcode_filter(self, queryset, name, value):
+        query = Q()
+        for barcode in value.split(" "):
+            query |= Q(barcode__contains=barcode)
```
UlysseFG, line 24: This is a promising avenue ... To keep thing simple maybe you can load the "batch" filterset just when you have a batch request ... that way you may just use __in instead of __contains and ignore the other type of comparison in the batch filterset. The rest of the logic would remain the same for other fields. This would be a patch but this might prevent you from having to recode the whole filterset_field logic in your filterset.

backend/fms_core/filters.py
```diff
@@ -0,0 +1,58 @@
...
+    _individual_filterset_fields,
+)
+
+class ContainerBatchFilter(django_filters.FilterSet):
+    barcode = django_filters.CharFilter(field_name="barcode", method="batch_barcode_filter")
+    name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+    class Meta:
+        model = Container
+        fields = _container_filterset_fields
```
UlysseFG, line 19: are these fields already filled by the call to filterset_fields in the viewset ?

backend/fms_core/filters.py
UlysseFG: Mmm this may already be doing that for those 2 fields ... You'll still have to use a better comparison than contains for those batch fields.

PR #71, opened 2021-05-19
applying conversion to decimal to volume computation using delta volume.

UlysseFG:
When we compute the new volume with a delta volume we need to convert the resulting volume to a Decimal of 3 positions.

### DIFF
--- a/backend/fms_core/resources/sample_update.py
+++ b/backend/fms_core/resources/sample_update.py
@@ -126,7 +126,7 @@ class SampleUpdateResource(GenericResource):
             # Manually process volume history and don't call superclass method
             delta_vol = blank_str_to_none(data.get("Delta Volume (uL)"))  # "" -> None for CSVs
             if delta_vol is not None:  # Only update volume if we got a value
-                obj.volume = obj.volume + Decimal(delta_vol)
+                obj.volume = float_to_decimal(obj.volume + Decimal(delta_vol))
             return
 
         if field.attribute == 'depleted':


jleblond:
It's fine for me, please also do a commit to master afterwards :)

romgrk:
> When we compute the new volume with a delta volume we need to convert the resulting volume to a Decimal of 3 positions.

I don't see the 3 decimals thing in this PR, is it 3 decimals by default?

> It's fine for me, please also do a commit to master afterwards :)

Should target master first by default, and backport it to qc if required.

UlysseFG:
> > When we compute the new volume with a delta volume we need to convert the resulting volume to a Decimal of 3 positions.
> 
> I don't see the 3 decimals thing in this PR, is it 3 decimals by default?

Yeah the call to float_to_decimal has a default of 3 decimals places.
 
> > It's fine for me, please also do a commit to master afterwards :)

> Should target master first by default, and backport it to qc if required.

Can you explain to me why we should do the port that direction specifically? I'm not sure i see an advantage to doing it that way. I don't mind either way i am just curious that i may be missing something.

romgrk:
> Can you explain to me why we should do the port that direction specifically? I'm not sure i see an advantage to doing it that way. I don't mind either way i am just curious that i may be missing something.

It's conceptually easier to have a single development branch, from which the commits are merged into the production (or qc) branches. It creates a commit flow always in the same direction: feature branch => master => qc => prod. Merging sometimes on qc, sometimes on master, makes it unclear which one is the development branch.

branching

UlysseFG:
I see ... i just fear that merging master into qc will propagate changes that should not apply to the production branch. But as i watch you also imply that we should create the branch from the last released tag instead of the head ... that would solve the problem ... I'll do that in the future...

romgrk:
Yes, and prod should pull from a branch that's always behind qc, containing only the commits that have passed the qc checks. The only case where I would commit directly on qc/prod would be if there is a security hotfix.

image

jleblond:
@romgrk Indeed, usually in agile development (with very often small and quick releases) it's better to have the prod branch on master. However in our case, we go through relatively longer cycles of testing on qc which may add up a month before being released to prod, and we have fixed versions. Also, we have to be very careful because any small change that is made inadvertely may have a big impact on all the data we have (requiring a lot of data curation). Although unconventional, I think in our case it's best to have dev pull from prod, rather than the inverse.

PR #75, opened 2021-05-26
Revision endpoint for users

jleblond:
More efficient way of loading versions in a User details page. 
On page loading, it loads the Revision objects. And it only loads the corresponding Version objects when you click on a Revision (+) button. 

@romgrk not a big deal, but I'm not sure about having those Versions still stored in the Redux store (under Users/itemsByID/Versions)..

### DIFF
--- a/frontend/src/components/users/UsersDetailContent.js
+++ b/frontend/src/components/users/UsersDetailContent.js
@@ -68,6 +68,7 @@ const ReportsUserContent = ({canWrite, isFetching, usersError, usersByID, groups
   const history = useHistory();
   const {id} = useParams();
   const [expandedGroups, setExpandedGroups] = useState({});
+  const [isLoadRevisions, setIsLoadRevisions] = useState(false);
 
   const user = usersByID[id];
 
@@ -75,7 +76,8 @@ const ReportsUserContent = ({canWrite, isFetching, usersError, usersByID, groups
     get(id)
   }
 
-  if (user && !user.revisions && !user.isFetching) {
+  if (user && !user.isFetching && !isLoadRevisions) {
+    setIsLoadRevisions(true)
     setTimeout(() => listRevisions(user.id), 0);
   }
 
--- a/frontend/src/components/users/UsersDetailContent.js
+++ b/frontend/src/components/users/UsersDetailContent.js
@@ -219,7 +219,6 @@ function TimelineEntry({ revision, expandedGroups, setExpandedGroups, listVersio
     listVersions(revision.user, revision.id)
         .then(response => {
           setGroup(response.results)
-          console.log(response)
           setIsLoading(false)
         })
   }
@@ -261,9 +260,7 @@ function TimelineEntry({ revision, expandedGroups, setExpandedGroups, listVersio
 
 function renderTimelineLabel(revision) {
   return (
-      <div>
-        <Text type="secondary">{dateToString(revision.date_created)}</Text>
-      </div>
+      <Text type="secondary">{dateToString(revision.date_created)}</Text>
   )
 }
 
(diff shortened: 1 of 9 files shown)


frontend/src/modules/users/actions.js
```diff
@@ -73,19 +74,31 @@ export const listTable = ({ offset = 0, limit = DEFAULT_PAGINATION_LIMIT } = {},
...
 }
 
+export const listVersions = (userId, revisionId) => (dispatch, getState) => {
+    const user = getState().users.itemsByID[userId];
+    if (user.isFetching) return Promise.resolve();
+    const meta = { id: userId };
+    return dispatch(
+        networkAction(
+            LIST_VERSIONS,
+            api.users.listVersions(userId, {revision__id: revisionId, limit: 10000}),
```
UlysseFG, line 97: Is it necessary to maintain this limit ? I would still go for something less extreme like 400-500 or does it still causes issues ?

jleblond: yes it is necessary. otherwise the limit by default is 100 (and most of Ariane's largest modifications contain about 4K Versions per Revision)

jleblond:
Display remains the same as the version in production. I did not change any of the content. If you feel like there should be changes, I would open a new ticket for that.

frontend/src/modules/users/actions.js
UlysseFG: cant they be loaded with "load more" 500 each time ? i think if someone really need to dig into a 4000 versions revision they can tolerate to click a few times. Most likely they do not need to list all versions to find what they need and since there is little info at the revision level it is likely not even the revision they are looking for (*jedi hand waving*).

UlysseFG:
> The issues you mentioned appear in that screenshot from the biobank .. website
Yeah but since this is extra backlog we are doing i think we can squeeze in a little improvement to the looks ; ). Totally optional.

frontend/src/modules/users/actions.js
jleblond: They could but I am not sure it is worth implementing it. It happens for like 3-4 revisions, among the 95 000 existing revisions up to now..

jleblond:
Yes but as I have said previously, I'd rather have PRs for different features. And we didnt decide on what information to display there. This would require more discussion, and I'd rather invest this time discussing Experiments :P

UlysseFG (approved):
Well done for the left aligned timeline it really gives much more space for the expanded display. We'll revisit this some other time ... maybe.

frontend/src/components/users/UsersDetailContent.js
```diff
@@ -239,9 +208,21 @@ const tableStyle = {
...
+  const [group, setGroup] = useState([]);
+  const [isLoading, setIsLoading] = useState(undefined);
+
+  const onClick = () => {
+    setIsLoading(true)
+    setExpandedGroups(set(expandedGroups, revision.id, !isExpanded))
+    listVersions(revision.user, revision.id)
+        .then(response => {
+          setGroup(response.results)
+          console.log(response)
```
romgrk, line 222: Remove console.log before merging

frontend/src/components/users/UsersDetailContent.js
```diff
@@ -279,7 +261,9 @@ function TimelineEntry({ group, expandedGroups, setExpandedGroups }) {
 
 function renderTimelineLabel(revision) {
   return (
-    <Text type="secondary">{dateToString(revision.date_created)}</Text>
+      <div>
+        <Text type="secondary">{dateToString(revision.date_created)}</Text>
+      </div>
```
romgrk, line 266: Necessary to wrap in a div?

romgrk (commented):
Wdym about storing revisions in the store? They're not essential but it's nice to have things like the rest of the items.

PR #78, opened 2021-05-31
Rename secret key env variable

UlysseFG:
Renaming the secret key env variable to match the pattern used across the app.


jleblond (approved):

jleblond:
Just dont forget to do the changes in Prod too!

PR #79, opened 2021-05-31
Set empty warning when too many errors.

UlysseFG:
The template fails when there is too many errors because the warnings are not set. Setting the warning to an empty list in that case.

### DIFF
--- a/backend/fms_core/resources/_generic.py
+++ b/backend/fms_core/resources/_generic.py
@@ -70,6 +70,7 @@ class GenericResource(resources.ModelResource):
         else:
             row_result = self.get_row_result_class()()
             row_result.import_type = RowResult.IMPORT_TYPE_SKIP
+            row_result.warnings = []
         return row_result
 
     def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):


romgrk:
Is there a reason why you keep creating PRs for both branches? Is this urgently needed on qc? And wouldn't it be possible to just cherry-pick the wanted commits or just merge the same branch in qc & master rather than creating 2 feature branches?

UlysseFG:
I do this because it's faster atm. I will have to read and try to figure out if you are pushing forward is actually working. Each attempt i've made i get more commit that i care to deal with and some changes that are not desired. This complicate the job for a one liner... I'm not sure that cherry picking won't just lead to more cherry picking as the branch move appart...

romgrk:
What about merging the same feature branch in both qc and master?

UlysseFG:
There is code in qc that is not on master and there is code on master that is not on qc ... So unless you make cherry-picking and prune stuff out of your merges you'll introduce errors. Unless you are careful and take the time to validate each commit and change this is not an option. I am still unconvinced that there is a way to do that with git more efficiently than making one branch for each version.

romgrk:
What code is only in qc and why does it need to be only in qc?

PR #81, opened 2021-05-31
Recursive container search

jleblond:
Description: This allows the user to lookup samples by containers recursively. Let's say rack 1 (barcode: rack1) contains tube A (sample A), tube B (sample B), tube C (sample C). When looking up in the Sample list page by applying filter container barcode 'rack1', you will get a list of samples with sample A, sample B, sample C. Previously you would not get any sample in the list, because rack1 does not contain any sample directly.

Name of recursive option and variable are open for discussion :)

### DIFF
--- a/backend/fms_core/viewsets.py
+++ b/backend/fms_core/viewsets.py
@@ -553,14 +553,24 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
     ]
 
     def get_queryset(self):
-        recursive = self.request.query_params.get("recursive", False)
+        container_barcode = self.request.query_params.get('container__barcode__recursive')
+        container_name = self.request.query_params.get('container__name__recursive')
+        recursive = container_barcode or container_name
+
         if recursive:
-            container = self.request.query_params.get('container_id')
+            containers = Container.objects.all()
+            if container_barcode:
+                containers = containers.filter(barcode=container_barcode)
+            if container_name:
+                containers = containers.filter(name=container_name)
+
+            container_ids = tuple(containers.values_list('id', flat=True))
+
 
             parent_containers = Container.objects.raw('''WITH RECURSIVE parent(id, location_id) AS (
                                                                SELECT id, location_id
                                                                FROM fms_core_container
-                                                               WHERE id = %s
+                                                               WHERE id IN %s
 
                                                                UNION ALL
 
@@ -568,7 +578,7 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
                                                                FROM fms_core_container AS child, parent
                                                                WHERE child.location_id = parent.id
                                                            )
-                                                           SELECT * FROM parent''', [container])
+                                                           SELECT * FROM parent''', params=[container_ids])
 
             return self.queryset.filter(container__in=parent_containers)
 
--- a/backend/fms_core/viewsets.py
+++ b/backend/fms_core/viewsets.py
@@ -566,6 +566,8 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
 
             container_ids = tuple(containers.values_list('id', flat=True))
 
+            if not container_ids:
+                container_ids = tuple([None])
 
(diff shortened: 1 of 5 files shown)


backend/fms_core/viewsets.py
```diff
@@ -552,6 +552,42 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
...
+
+                                                               UNION ALL
+
+                                                               SELECT child.id, child.location_id
+                                                               FROM fms_core_container AS child, parent
+                                                               WHERE child.location_id = parent.id
+                                                           )
+                                                           SELECT * FROM parent''', params=[container_ids])
+
+            return self.queryset.filter(container__in=parent_containers)
```
UlysseFG, line 585: I would set the self.queryset to the new value instead of having a second return statement.

jleblond: What do you mean?

UlysseFG: i would set the viewset internal queryset to the correct one instead of just returning it. In case somewhere else django actually uses it for one thing or another without going through get queryset. More than a single return statement in a function need to be justified in some way in my eyes also (readability, performance, ...).

romgrk: > More than a single return statement in a function need to be justified in some way

I have the opposite opinion, but I feel the the early-return style is more readable. E.g., first is the current style, second is early-return style:

```python
if a:
  # ...
  return a
return b
```

```python
if not a:
  return b
# ...
return b
```

It allows to filter out easy cases early, and the end of the function is where the hard cases are handled. Otherwise you need to jump many lines of code to make the connection between the `if` and the `return`, and that is harder to understand.

frontend/src/components/filters/getFilterProps.js
```diff
@@ -48,31 +48,52 @@ function getInputFilterProps(column, descriptions, filters, setFilter, setFilter
       confirm()
   }
 
-  const onToggleSwitch = checked => {
+  const onToggleSwitch = (checked, option )=> {
```
romgrk, line 51: Formatting typo around `)=>`. Also usually `(key, value)` is a more common order for those arguments.

frontend/src/components/filters/getFilterProps.js
```diff
@@ -48,31 +48,52 @@ function getInputFilterProps(column, descriptions, filters, setFilter, setFilter
...
+    setFilterOption(dataIndex, option, checked)
+  }
+
+  const onChangeRecursive = checked => {
+    onToggleSwitch(checked, 'recursiveMatch')
     setFilterOption(dataIndex, 'exactMatch', checked)
```
romgrk, line 57: Is it normal that `onChangeRecursive` sets `exactMatch` to the same value?

frontend/src/components/filters/getFilterProps.js
```diff
@@ -48,31 +48,52 @@ function getInputFilterProps(column, descriptions, filters, setFilter, setFilter
...
-          <Switch
-            size='small'
-            checked={options?.exactMatch ?? false}
-            onChange={e => onToggleSwitch(e, dataIndex)}
+      <div style={{ padding: 8, alignItems: 'center' }}>
+          <Input
+            ref={inputRef}
+            allowClear
+            placeholder={`Search ${description.label}`}
+            style={{ marginRight: 8}}
```
romgrk, line 68: Typo, space after `8`

frontend/src/components/filters/getFilterProps.js
jleblond: Yes. You cant do recursive search with a non-exact value, so it will change automatically

romgrk (commented):
LGTM overall

frontend/src/components/filters/getFilterProps.js
romgrk: And what if the user unchecks `exactMatch` after having checked `recursiveMatch`? I would not fix the issue this way, I would make sure that `exactMatch` is ignored in the query params creation function, and maybe disabled `exactMatch` if `recursiveMatch` is checked.

jleblond: yes it is disabled

PR #83, opened 2021-06-08
Addition of an env variables file and upgrade of the webpack to fix some vulnerabilities

UlysseFG:
The .env files is now fetched to load the FMS_ENV variable.
Fixing some vulnerabilities led to upgrading the webpack to v5 which had to be fixed by some changes (removed a call to webpack.NamedModulesPlugin, added a rule to prevent the enforcement of fully specified module names. upgraded html-webpack-plugin to v5).


jleblond (approved):
Great, thanks!

PR #84, opened 2021-06-08
Improve container hierarchy with preloading

romgrk:


jleblond:
Closed. Will solve the problem with a different approach: https://github.com/c3g/freezeman/pull/85

PR #85, opened 2021-06-09
Container hierarchy samples

jleblond:
Alternate approach to the problem in our container hierarchy tree, where we would need to explode the contains with direct children samples. 

I got rid of the Ant tree nodes for the samples, instead I am directly listing the Samples under its container, in the same tree node. 

TODO: re-order the samples according to their tube coordinates and the samples coordinates (in a plate for example)


frontend/src/components/containers/ContainersDetailContent.js
```diff
@@ -32,8 +32,9 @@ const ContainersDetailContent = ({containersByID, samplesByID, usersByID, get, l
   if (!isLoaded)
     get(id);
 
-  if (isLoaded && !container.parents)
+  if (isLoaded && !container.parents) {
     listParents(id);
+  }
```
romgrk, line 37: Inconsistent style, if you're going to add braces to single-line ifs add them everywhere (eg above).

romgrk:
Haven't seen the recursive logic in action as it's not used yet, but for the style, I'd suggest to display the sampe inline if there is only one of them. If there is more than one of them, I'd suggest to display it as another level in the tree.

Screenshot from 2021-06-13 15-40-05
Screenshot from 2021-06-13 15-41-58

jleblond:
Ulysse had the same suggesiton. However I want to keep it the same for all type of containers (some of them being plates of 96 or 384 samples), in which case it would be better to have them listed all underneath the container. 

I decided to get rid of the recursive actions, as it is more efficient using withSamples for each of the samples under a container. I will clean those actions up from the actions/reducers.

UlysseFG (approved):
Looks good...

PR #86, opened 2021-06-14
complete the migration for processmeasurement rename.

UlysseFG:
There was a few fields that were not adjusted to the new name of the model.


jleblond (commented):
The difference is related to the related name? Are those the related names that we want?

UlysseFG:
> The difference is related to the related name? Are those the related names that we want?

yeah the related name is the equivalent of the old one but we could always have something else... This just match what is used in the model...

PR #92, opened 2021-06-21
PropertyType and PropertyValue

jleblond:


backend/fms_core/models/property_value.py
```diff
@@ -0,0 +1,59 @@
...
+__all__ = ["PropertyValue"]
+
+@reversion.register()
+class PropertyValue(TrackedModel):
+    value = models.JSONField("Property value", validators=[PROPERTY_VALUE_VALIDATOR],
+                                      help_text="Property value")
+    property_type = models.ForeignKey(PropertyType, on_delete=models.PROTECT, related_name="property_type",
+                                  help_text="Property type")
+
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
```
UlysseFG, line 27: on_delete=models.PROTECT instead of CASCADE maybe ? i don't see much case where this would be important but between both i would prefer not to lose all property values related to a content_type ...

backend/fms_core/models/property_type.py
```diff
@@ -0,0 +1,56 @@
...
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH,
+                            help_text="The name of the property",
+                            validators=[name_validator])
+
+    value_type = models.CharField(
+        max_length=20,
+        choices=((type, type) for type in VALUE_TYPE_CHOICES),
+        help_text="Enumerated type to define value type"
+    )
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
```
UlysseFG, line 30: on_delete=models.PROTECT instead of CASCADE maybe ? i don't see much case where this would be important but between both i would prefer not to lose all property values related to a content_type ...

backend/fms_core/tests/test_models/property_type_test.py
```diff
@@ -0,0 +1,77 @@
...
+            except ValidationError as e:
+                self.assertTrue('value_type' in e.message_dict)
+                raise e
+
+    def test_invalid_content_object(self):
+        name = "Test_invalid_value_type"
+        content_object = Container.objects.create(**create_container(barcode='INVALIDPROPERTYTYPE'))
+        with self.assertRaises(ValidationError):
+            try:
+                PropertyType.objects.create(name=name, value_type="invalid_content_object", content_object=content_object)
```
UlysseFG, line 46: There is more that one error here (value_type and content_object) this could come bite us in the future.

backend/fms_core/tests/test_models/property_type_test.py
```diff
@@ -0,0 +1,77 @@
...
+            except ValidationError as e:
+                self.assertTrue('content_object' in e.message_dict)
+                raise e
+
+    def test_missing_name(self):
+        content_object = self.protocol
+        with self.assertRaises(ValidationError):
+            try:
+                PropertyType.objects.create(value_type="invalid_content_object", content_object=content_object)
```
UlysseFG, line 55: value_type is not valid here also ...

backend/fms_core/tests/test_models/property_type_test.py
```diff
@@ -0,0 +1,77 @@
...
+                PropertyType.objects.create(name=name, content_object=content_object)
+            except ValidationError as e:
+                self.assertTrue('value_type' in e.message_dict)
+                raise e
+
+    def test_missing_content_object(self):
+        name = "missing_content_object"
+        with self.assertRaises(ValidationError):
+            try:
+                PropertyType.objects.create(name=name, value_type="invalid_content_object")
```
UlysseFG, line 74: value_type is not valid here also ...

backend/fms_core/tests/test_models/property_value_test.py
```diff
@@ -0,0 +1,98 @@
...
+        self.property_type_with_int = PropertyType.objects.create(name="PropertyNameTestInt",
+                                                                            value_type="int",
+                                                                            content_object=self.protocol)
+        self.process, _ = Process.objects.get_or_create(protocol=self.protocol)
+
+    def test_property_value(self):
+        content_object = self.process
+        value = 10.1
+        property_type_with_float = PropertyType.objects.create(name="PropertyNameTestFloat",
+                                                                            value_type="float",
```
UlysseFG, line 21: Alignement of the params ...

backend/fms_core/tests/test_models/property_value_test.py
```diff
@@ -0,0 +1,98 @@
...
+        self.assertEqual(pv.value, value)
+        self.assertEqual(type(pv.value).__name__, pv.property_type.value_type)
+
+    def test_incorrect_json_value_schema(self):
+        content_object = self.process
+        value = "testtest"
+        with self.assertRaises(ValidationError):
+            try:
+                PropertyValue.objects.create(value=[value, value],
+                                             property_type=self.property_type_with_int,
```
UlysseFG, line 38: property_type do not match the value ...

jleblond: the point here is to have an invalid json schema (such as [value, value]) the property type could be anything that is valid. However i changed it to str, so it is closer to our value, but it is still not accurate and cant be

backend/fms_core/models/property_value.py
jleblond: I prefer CASCADE over PROTECT. If we are to delete a model (which is done after lots of reflexion), we dont want to have any dangling objects related that are broken, nor do we want to go on to manually delete all of these objects in our migration.

UlysseFG: That could be true in some contexts but here we have meta data that we'll never want to delete and even less in bulk cascade like that. If we ever need to move that information we'll remove the PROTECT once the data is safe. I do not see any good reason to use cascade here. This could only lead to mistakenly deleting this information.

backend/fms_core/tests/test_models/property_value_test.py
UlysseFG: I mentioned this because if you remove the schema issue you'll still have a property_type issue with the value. If you have more than one error in these tests if the target error do not fire this might get masked by the other one (not in this case, but as a general danger of leaving other errors in those test cases.

jleblond: What i mean is that in this case the type array still wont match with any of the property type (int, float, bool,str, etc.)

PR #93, opened 2021-06-22
Instrument and protocol creation. Added process.parent_process and experiment_run.start_date.

UlysseFG:
Small additions to the model for the implementation of experiment_run.


backend/fms_core/migrations/0021_v3_3_0.py
```diff
@@ -36,68 +37,69 @@ def create_platforms_and_instrument_types(apps, schema_editor):
...
 
     INSTRUMENT_TYPES_BY_PLATFORM = {
-        "LS454": ["454 GS",
-                  "454 GS 20",
-                  "454 GS FLX",
-                  "454 GS FLX+",
-                  "454 GS FLX Titanium",
-                  "454 GS Junior",
-                  ],
+        "LS454": [ "454 GS",
```
jleblond, line 40: What had to be changed?

backend/fms_core/tests/test_models/experiment_run_test.py
```diff
@@ -17,39 +19,52 @@ def setUp(self):
...
                 raise e
     
     def test_missing_container(self):
-         with self.assertRaises(ValidationError):
+        with self.assertRaises(ValidationError):
             try:
                 er_without_container = ExperimentRun.objects.create(experiment_type=self.experiment_type,
-                                                                    instrument=self.instrument)
+                                                                    instrument=self.instrument,
+                                                                    start_date = self.start_date)
```
jleblond, line 47: Consistency around =

backend/fms_core/migrations/0021_v3_3_0.py
UlysseFG: Only alignments ...

jleblond: I dont think we should add spaces after brackets before the elements.. if that's what leads to changing the alignment

UlysseFG: ok i'll revert those spaces

PR #96, opened 2021-06-23
New container spec for infinium

UlysseFG:


jleblond:
There is a need for a migration too. Its part of the values allowed for the kind of a container

PR #102, opened 2021-07-07
Experiments (v3.3)

jleblond:
New PR as the result of the merging of the back-end WIP https://github.com/c3g/freezeman/pull/94 and the front-end WIP https://github.com/c3g/freezeman/pull/99  

TODO: 
- [x] Ingestion of the template through the front-end
- [x] Display Samples in a ExperimentRun page tab
- [x] Handling errors related to the Sample section of the Experiment template
  - [x] Enough volume left in the source samples (and it is not depleted) for the Experiment
  - [x] Source container barcode is found
  - [x] 2 samples in the same experiment container with the same position
- [x] Handling depletion of sample during ExperimentRun
- [ ] ExperimentRun resource test
- [ ] Cypress integration tests for the template, for the list page, and for different entity pages

TO VERIFY:
- [x] Verify for the volume used for a sample (should it be done in the ProcessMeasurement for the transfer VS the experiment) 
- [ ] The way the date and time properties are stored as JSON in PropertyValue

BONUS TASKS:
- [x] Add the Platform and the Instrument Type as columns and information for ExperimentRuns
- Improve the ExperimentRunResource logic to make it even more modular (i.e. row numbers for sections hardcoded for now)

ASSUMPTIONS:
- Samples related to an Experiment are fetched through their association to their Container. This means, that the Samples should not be transferred over to another Container after being used in an Experiment. Also, Container would not be re-used for another Experiment. If this logic is not true, we should change some of the calls to the API from the front-end.


backend/fms_core/models/property_value.py
```diff
@@ -37,12 +37,17 @@ def add_error(field: str, error: str):
...
             property_type_value_type = self.property_type.value_type
-            if value_type != property_type_value_type:
+
+            # Exception applies to float values which might be entered as ints
+            if property_type_value_type == 'float' and type(self.value).__name__ == 'int':
+                self.value = float(self.value)
+
+            # Check if the property value data type matches the property type 'value_type' attribute
+            if type(self.value).__name__ != property_type_value_type:
                 add_error("value", f"Value type {value_type} does not match property type {property_type_value_type}")
```
UlysseFG, line 48: "... does not match property type's value type ..." maybe... it's more precise though a bit complex to read.

backend/fms_core/resources/_utils.py
```diff
@@ -32,6 +33,19 @@ def skip_rows(dataset: Dataset, num_rows: int = 0, col_skip: int = 1) -> None:
             continue
         dataset.append(tuple(str_normalize(c) if isinstance(c, str) else ("" if c is None else c) for c in r))
 
+def wipe_import_row_result(import_result, row):
```
UlysseFG, line 36: Do not forget row_warnings if you want a thorough cleanup ...

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,279 @@
...
+    float_to_decimal,
+    get_normalized_str,
+)
+
+from ..experiments import PROTOCOLS_BY_EXPERIMENT_TYPE_NAME
+
+__all__ = ["ExperimentRunResource"]
+
+
+process_content_type = ContentType.objects.get(app_label="fms_core", model="process")
```
UlysseFG, line 37: Why is this declared out of scope ? process_content_type is used once ... i would declare this just before usage.

jleblond: its in a loop (the one iterating through the experiment properties) i wanted to save some db call. I guess I could still put it at the top of the import_obj function though

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,279 @@
...
+    experiment_id = Field(column_name='Experiment ID')
+    container_barcode = Field(column_name='Experiment Container Barcode',
+                              widget=ForeignKeyWidget(Container, field='barcode'))
+    container_kind = Field(column_name='Experiment Container Kind')
+    instrument_name = Field(column_name='Instrument Name',
+                              widget=ForeignKeyWidget(Instrument, field='name'))
+    start_date = Field(attribute='start_date', column_name='Experiment Start Date', widget=DateWidget())
+
+
+    ERROR_CUTOFF = 500
```
UlysseFG, line 50: Is this for testing ? What is the reason for changing the cutoff here ? Add a comment to the code with your reason. it is not obvious to me.

jleblond: If there is a problem with only one row, it would stop being processed very early (there is a lot of data in only one row). 
This is why I wanted to increase the cutoff number. 500 is abritrary we can choose another value higher than 20, and I will add a comment if that's ok with you.

backend/fms_core/resources/experiment_run.py
UlysseFG: You are right we could save some hits to the db ... Maybe in the class init ? the import_obj will still get called many times ...

jleblond: I dont see the problem with it hanging there. It is not imported anywhere else. I am unsure about having it declared as a property/attribute of the ExperimentRunResource, but if you think it's still better, I wont argue :P

UlysseFG: I hate global variables mainly ... but yeah its not really a class attribute/property ... it's up to you but i would not define anything out of the scope of the class unless its a constant and even then...

backend/fms_core/resources/experiment_run.py
UlysseFG: yeah we'll need to do some error cleanup here too ... but yeah for now a comment will do just fine and we'll tweak as we cleanup the error messages generated...

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,279 @@
...
+    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, raise_errors=False, **kwargs):
+        import_result = super().import_row(row, instance_loader, using_transactions, dry_run, raise_errors, **kwargs)
+
+        row_id = row.get("#")
+
+        # We are in the Sample section
+        if isinstance(row_id, int) and row_id > 100:
+            sample_row = {
+                "experiment_id": row.get("Experiment ID"),
+                "source_container_barcode": row.get("Experiment Container Barcode"),
```
UlysseFG, line 114: Ugh ... For someone that likes clarity this is truly funky ... I see the why ... but this will likely lead to problems ... what happens if there is more column in the sample section than in the experiment section for some experiment ?

UlysseFG: After our little discussion maybe you should make a comment here explaining that you are mapping to the column names of the experiment section.

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+
+    def import_row(self, row, instance_loader, using_transactions=True, dry_run=False, raise_errors=False, **kwargs):
+        import_result = super().import_row(row, instance_loader, using_transactions, dry_run, raise_errors, **kwargs)
+
+        row_id = row.get("#")
+
+        # We are in the Sample section
+        if isinstance(row_id, int) and row_id > 100:
```
UlysseFG, line 111: This is a bit risky if the users decide to cut unused lines and renumber them...

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,279 @@
...
+        reversion.set_comment("Imported ExperimentRun from template.")
+
+        self.experiments.update({self.temporary_experiment_id: instance.id})
+
+
+    def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
+        super().after_import(dataset, result, using_transactions, dry_run, **kwargs)
+
+        if not dry_run:
+            transfer_protocol = Protocol.objects.get(name="Transfer")
```
UlysseFG, line 240: Tranfer protocol is part of the experiment run protocol.

frontend/src/components/containers/ContainersDetailContent.js
```diff
@@ -47,30 +71,39 @@ const ContainersDetailContent = ({containersByID, samplesByID, usersByID, get, l
...
+            <TrackingFieldsContent entity={container}/>
+
+            <Descriptions bordered={true} size="small" title="Content Details" style={{marginTop: "24px"}}>
+              <Descriptions.Item span={3}>
+                <ContainerHierarchy key={id} container={isLoaded ? container : null} />
+              </Descriptions.Item>
+            </Descriptions>
+          </TabPane>
+
+          <TabPane tab={`Experiments (${container.experiment_runs?.length})`} key="2" style={tabStyle}>
```
UlysseFG, line 101: Make tab optional using container_kind.is_run_container ?

frontend/src/components/experimentRuns/ExperimentRunsListContent.js
```diff
@@ -0,0 +1,165 @@
...
+      <Tag>{experimentTypes.itemsByID[experimentRun.experiment_type]?.workflow}</Tag>,
+  },
+  {
+    title: "Instrument",
+    dataIndex: "instrument",
+    sorter: true,
+    options: instruments.items.map(x => ({ label: x.name, value: x.name })), // for getFilterProps
+    render: (_, experimentRun) =>
+      <div>{instruments.itemsByID[experimentRun.instrument]?.name}</div>,
+  },
```
UlysseFG, line 47: Not sure we need to add this right now (its not required for infinium but will be in the future) but at some point the instrument_type may be required (some instrument have names that do not match their type)

jleblond: It's in the bonus tasks ;P

frontend/src/components/containers/ContainersDetailContent.js
jleblond: I am not a fan of having information displayed or hidden according to some business logic, without much explanations. I'd rather have a compromise between the two where it may be hidden but it is highlighted that it's because it's not a run container.. or have it displayed with a note that it's empty due to it not being a run container... will think about it

frontend/src/components/experimentRuns/ExperimentRunsSamples.js
```diff
@@ -0,0 +1,113 @@
...
+}
+
+const renderProcessMeasurement = (processMeasurement) =>
+    <>{processMeasurement.volume_used} µL</>
+
+const renderListHeader = (container) => {
+  return (
+      <>
+        <Title level={5}>
+          Samples from <Link to={`/containers/${container.id}`}> {container.name} </Link>
```
UlysseFG, line 38: Missing a header for the volume used.

frontend/src/components/experimentRuns/ExperimentRunsProperties.js
```diff
@@ -0,0 +1,53 @@
...
+const ExperimentRunsProperties = ({
+  propertyValuesByID,
+  propertyIDs,
+  protocolName,
+}) => {
+  let properties = []
+  let columns = []
+  let data = []
+
+  if ( propertyValuesByID[propertyIDs[0]]) {
```
UlysseFG, line 22: space after parenthesis and is this a test to check if the protocol has properties ? i would create a intermediate variable hasProperties or hasPropertyValues for readability.

backend/fms_core/resources/experiment_run.py
jleblond: I agree, but in this method you only get an OrderedDict with the content of row. You can situate it within the dataset, you don't even have access to the whole dataset (see image for an example) Do you have a suggestion? 

image

backend/fms_core/resources/experiment_run.py
```diff
@@ -0,0 +1,281 @@
...
+        for i, (property, value) in enumerate(data.items()):
+            # Slicing columns not containing properties
+            if i < self.protocols_starting_idx or i >= self.protocols_ending_idx:
+                pass
+            else:
+                property_type = self.property_types_by_name[property]
+                if value:
+                    process = experiment_run_processes_by_protocol_id[property_type.object_id]
+
+                    if type(value).__name__ in ('datetime','time'):
```
UlysseFG, line 194: time does not seem to differentiate between a time of day which could use a Z and a time delta which would not.

frontend/src/components/experimentRuns/ExperimentRunsSamples.js
UlysseFG, line 38: also i would replace "Samples from " by "Samples inside container " to be clearer.

UlysseFG (commented):
A few things to fix...

PR #103, opened 2021-07-08
Fix error messages

sebastianballesteros:
Added error_messages to the models so the messages are more specific. 
Fixed some exceptions that were thrown in the middle of the import and prevented the attributes that came after them to be processed.


backend/fms_core/resources/sample.py
```diff
@@ -175,7 +175,23 @@ def import_obj(self, obj, data, dry_run):
         except Exception as e:
             individual_created = False
             individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+
+        #Individual field errors
+        if not data["Taxon"]:
```
jleblond, line 180: Would there be a way to make it more modular (or even re-usable in other Resources which need more precise errors too)? :)

backend/fms_core/models/sample.py
```diff
@@ -117,12 +117,19 @@ class Sample(TrackedModel):
...
+
     alias = models.CharField(max_length=200, blank=True, help_text="Alternative sample name given by the "
                                                                    "collaborator or customer.")
-    individual = models.ForeignKey("Individual", on_delete=models.PROTECT, related_name="samples", help_text="Individual associated "
-                                                                                     "with the sample.")
+    individual = models.ForeignKey("Individual",
+                                    on_delete=models.PROTECT,
+                                    related_name="samples",
+                                    help_text="Individual associated with the sample.",
+                                    error_messages={'null': 'Individual could not be processed due to missing information. Taxon and ID are necessary.'})
```
jleblond, line 132: Same as below (my comment related to the container attribute)

backend/fms_core/models/sample.py
```diff
@@ -117,12 +117,19 @@ class Sample(TrackedModel):
     sample_kind = models.ForeignKey(SampleKind, on_delete=models.PROTECT,
                                     help_text="Biological material collected from study subject "
                                               "during the conduct of a genomic study project.")
-    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
-                            help_text="Sample name.")
+
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH,
+                            validators=[name_validator],
+                            help_text="Sample name.",
+                            error_messages={'blank': 'Sample name cannot be blank'})
```
jleblond, line 124: Is it necessary to override the very similar default error message for a blank value?

jleblond (commented):
I just tested it by submitting an incomplete Sample submission template to the front-end. It works well (the error message are more precise). I added a few technical comments on the code below.

Also, fyi, whenever you make changes to a django model, you have to generate the corresponding migration, otherwise your changes won't be applied (i.e.: the custom error_messages you added in the Sample model). You do so by running the django makemigrations command. We have a particular way of organizing our migrations (as much as possible by version number, you can find them in the fms_core/migrations folder)

Lastly, the ticket you got might have focused on some attributes of the Sample resource/submission template. I think such a feature  could eventually benefit some other resource/templates as well, so I'd recommend trying to make the code slightly more re-usable across different resources (for example, you could look at the GenericResource model)

backend/fms_core/resources/sample.py
```diff
@@ -272,7 +296,11 @@ def import_field(self, field, obj, data, is_m2m=False):
             # Ignore importing this, since it's a computed property.
             return
 
-        super().import_field(field, obj, data, is_m2m)
+        try:
+            super().import_field(field, obj, data, is_m2m)
+        except Exception as e:
+            pass
```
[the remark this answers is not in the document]
UlysseFG: ...or at least make sure you are more precise with your except to catch the case you want to ignore as precisely as possible and leave a comment in the code as to what condition this applies to so we can have the information without parsing the whole code for such an unorthodox behaviour...

backend/fms_core/models/sample.py
```diff
@@ -157,7 +164,9 @@ class Sample(TrackedModel):
     # TODO: I would prefer consistent terminology with Container if possible for this heirarchy
     container = models.ForeignKey(Container, on_delete=models.PROTECT, related_name="samples",
                                   limit_choices_to={"kind__in": SAMPLE_CONTAINER_KINDS},
-                                  help_text="Designated location of the sample.")
+                                  help_text="Designated location of the sample.",
+                                  error_messages={'null': 'Container could not be processed due to missing information. Kind, Name and Barcode are necessary.'})
```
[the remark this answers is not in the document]
UlysseFG: At this level, you only know that a container is missing ... This is too much to assume we are in sample submission and that the user did forget to fill in those fields. You could use this error message by inserting it manually at the level of the ressource (maybe on  the generic ressource since this could apply to multiple user submitted templates).

backend/fms_core/resources/sample.py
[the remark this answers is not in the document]
sebastianballesteros: Yes, I looked into it in detail and there's a problem when there's no container_kind. Instead of raising a Blank Exception (which is what is expected) and continuing parsing the row it throws and exception and it stops parsing that row. One option would be: 

```suggestion
            raise ValidationError({"container_kind": ValidationError([f"This field cannot be blank."], code="invalid")})
```

I tested it and it works.

backend/fms_core/resources/sample.py
```diff
@@ -244,8 +264,12 @@ def import_field(self, field, obj, data, is_m2m=False):
                 # This will throw an error if the kind specified mismatches with an
                 # existing barcode record in the database, which serves as an
                 # ad-hoc additional validation step.
-                container, _ = Container.objects.get_or_create(**container_data)
-                obj.container = container
+
+                try:
+                    container, _ = Container.objects.get_or_create(**container_data)
+                    obj.container = container
+                except:
```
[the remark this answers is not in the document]
sebastianballesteros: Definitely, I didn't mean to ignore this exception actually. 

```suggestion
                errors = e.update_error_dict(errors).copy()
```

backend/fms_core/models/sample.py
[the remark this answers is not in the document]
sebastianballesteros: Perfect. Should I just leave "Container could not be processed due to missing information"?
```suggestion
                                  error_messages={'null': 'Container could not be processed due to missing information.'})
```

backend/fms_core/resources/sample.py
sebastianballesteros: I am thinking about creating a function in _utils.py (GenericResource is also an option) and then call that function to validate specific attributes. That way it will be reusable. What do you think?

backend/fms_core/models/sample.py
sebastianballesteros: Same change as the container message. 
```suggestion
                                    error_messages={'null': 'Individual could not be processed due to missing information.'})
```

backend/fms_core/models/sample.py
sebastianballesteros: This particular error message had to be overridden because the default message would be "Name - This field cannot be blank." This ticket: https://206.12.92.46/issues/334 states that it might be confused with container name.

UlysseFG: This might be useful since by default it returns "name cannot be blank" which is not so useful when you have container name, individual name, sample name ... maybe though we could find a more generic approach to this by overloading the models default error_messages to prepend the name of the model as part of the message (%(app_label)s_%(class)s). Not sure of realistic this is though.

backend/fms_core/resources/sample.py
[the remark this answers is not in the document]
UlysseFG: The method you are in parse each field of the template (not only container_kind) if the import_field raise an error for another field this is going to get really annoying to track down...

I would suggest you do the catching of this error in the "field.attribute == "container_barcode" section raising the validation error if the container_kind is missing. 

Additionally, for the exception raised at the level of super.import_field, once the case is tested positively to make sure a validation error is raised if there is a problem you may be able to bypass the exception by setting self.fields_manually_excluded.extend(["container_kind"]) before raising your error...
(and again maybe not)...

This is hacky but at the moment its how we get by at other places (you can check ressource.extraction for an example of what is done elsewhere). If this does not work you'll have to dig a bit deeper ;).

backend/fms_core/models/sample.py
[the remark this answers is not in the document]
UlysseFG: This implies again a user input failure. Since you are at the model level, i would go with something more agnostic like "Sample container cannot be null" which is of very little use to an end user. I would encourage you to work at a more context specific level to be able to provide clearer and more detailed error messages. The field definition will not be able to differentiate whether you create the object from a template submission or programmatically somewhere else. You should work within the ressource for that or alternatively in a module that will be informed of the context somehow.

backend/fms_core/models/sample.py
UlysseFG: Same as above it implies missing user submitted information... this could happen for other reasons (e.g. someone forgot to create an individual altogether inside the app.

backend/fms_core/resources/sample.py
jleblond: I would retract my comment. if we are to change the tool we use, it may not be worth it for now to invest considerable time making it uniform across all the resources, so what's there works for me :)

backend/fms_core/resources/sample.py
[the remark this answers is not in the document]
jleblond: I don't understand the suggested change. Can you explain your intent?

backend/fms_core/resources/sample.py
sebastianballesteros: Great

backend/fms_core/resources/sample.py
```diff
@@ -198,10 +216,19 @@ def import_field(self, field, obj, data, is_m2m=False):
             obj.collection_site = get_normalized_str(data, "Collection Site")
 
         elif field.attribute == "sample_kind_name":
-            obj.sample_kind = SampleKind.objects.get(name=data["Sample Kind"])
+            try:
+                obj.sample_kind = SampleKind.objects.get(name=data["Sample Kind"])
+            except Exception as e:
+                pass
```
jleblond, line 222: Still unsure about passing the exception here too..

backend/fms_core/resources/sample.py
```diff
@@ -175,7 +175,25 @@ def import_obj(self, obj, data, dry_run):
         except Exception as e:
             individual_created = False
             individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+            errors["individual"] = ValidationError("Individual could not be processed due to missing or invalid information.", code="invalid")
+
+        #Individual field errors
+        validate_specific_field(errors, "Taxon", "taxon", data)
```
jleblond, line 181: would there be a less repetitive way of doing it? :)

backend/fms_core/resources/sample.py
```diff
@@ -175,7 +175,25 @@ def import_obj(self, obj, data, dry_run):
         except Exception as e:
             individual_created = False
             individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+            errors["individual"] = ValidationError("Individual could not be processed due to missing or invalid information.", code="invalid")
```
jleblond, line 178: This message is a bit vague to me, but if getting something more specific is much more complex, I would keep it that way.

backend/fms_core/resources/sample.py
sebastianballesteros: `#Individual field errors
        fields = {
            "Taxon": "taxon",
            "Individual ID": "individual_id",
            "Container Kind":"container_kind",
            "Container Name": "container_name",
            "Container Barcode": "container_barcode"
        }
        
        validate_specific_fields(data, errors, fields)`

Then function will traverse the dictionary. What do you think?

backend/fms_core/resources/sample.py
sebastianballesteros: Definitely! my bad. Forgot to raise a ValueError. 

`raise ValueError("Sample kind name cannot be blank.")`

backend/fms_core/resources/sample.py
```diff
@@ -178,16 +178,15 @@ def import_obj(self, obj, data, dry_run):
...
-
-        validate_specific_field(errors, "Individual ID", "individual_id", data)
-
-        #Container field errors
-        validate_specific_field(errors, "Container Kind", "container_kind", data)
-
-        validate_specific_field(errors, "Container Name", "container_name", data)
-
-        validate_specific_field(errors, "Container Barcode", "container_barcode", data)
+        fields = {
```
jleblond, line 181: a suggestion regarding the variable names; fields is a very broad variable name. could be difficult to search through the codebase with a Ctrl + F. I would suggest something more specific and explicit such as attribute_by_field_name or attribute_with_field_name, or field_name_attribute_dict; so anything that hints that the structure is a map/dictionary, and not just a list.

sebastianballesteros: Perfect!

UlysseFG:
I think we should close this PR i don't see this being useful since we are now using the TI. The work on the error message is an ungoing effort but that specific branch is dead i think.

jleblond:
Agree, closing it

PR #104, opened 2021-07-15
Implemented dynamic page size for pagination

sebastianballesteros:
Added a new module to deal with the pageSize state. Modified the action.js of container, individuals, samples and containers to take in account the pageSize state. Finally, added the new reducer to the store.

### DIFF
--- a/frontend/cypress/integration/all.spec.js
+++ b/frontend/cypress/integration/all.spec.js
@@ -4,11 +4,12 @@ import { containersTests } from "./containers.spec"
 import { individualsTests } from "./individuals.spec"
 import { samplesTests } from "./samples.spec"
 import { protocolsTests } from "./protocols.spec"
+import { paginationTests } from "./pagination.spec"
 
 // Tests
 context('All tests', () => {
   containersTests()
   individualsTests()
   samplesTests()
-  protocolsTests()
+  paginationTests()
 })
--- a/frontend/cypress/integration/all.spec.js
+++ b/frontend/cypress/integration/all.spec.js
@@ -11,5 +11,6 @@ context('All tests', () => {
   containersTests()
   individualsTests()
   samplesTests()
+  protocolsTests()
   paginationTests()
 })
--- a/frontend/src/modules/containers/actions.js
+++ b/frontend/src/modules/containers/actions.js
@@ -58,6 +58,7 @@ export const listTable = ({ offset = 0, limit = DEFAULT_PAGINATION_LIMIT } = {},
     if (containers.isFetching && !abort)
         return
 
+    const limit = getState().pagination.pageSize;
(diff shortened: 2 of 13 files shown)


jleblond:
Works very well, good job! Very good idea using the createSlice functionality in this case. 

Just a comment on the technical design: from what I see you added the pagination state in the Redux store, this is what is updated whenever the nb of items is changed by the user. However, I see that there are still the 'limit' (representing the number of items) in each of the states for the entities (ex: in containers, in samples, in individuals, etc.) I think this is duplicate information, and it is not necessary to have both. I would suggest doing a small refactor in order to have only one source of truth :P

sebastianballesteros:
Perfect. I will modify the action of each entity. I will look into the testing too! Indeed that would be a pretty good opportunity for me to get familiar with it :)

frontend/src/modules/containers/actions.js
```diff
@@ -53,11 +53,12 @@ export const list = (options) => async (dispatch, getState) => {
     ));
 };
 
-export const listTable = ({ offset = 0, limit = DEFAULT_PAGINATION_LIMIT } = {}, abort) => async (dispatch, getState) => {
+export const listTable = ({ offset = 0 } = {}, abort) => async (dispatch, getState) => {
```
jleblond, line 56: What I meant is not removing the default pagination limit from the params in the actions. 
It's more that in the store there are 2 places where the information is stored (pagination pageSize AND each of the entities page limit), it could be only stored in the pageSize from the pagination state.

frontend/cypress/integration/all.spec.js
```diff
@@ -4,11 +4,12 @@ import { containersTests } from "./containers.spec"
...
 import { samplesTests } from "./samples.spec"
 import { protocolsTests } from "./protocols.spec"
+import { paginationTests } from "./pagination.spec"
 
 // Tests
 context('All tests', () => {
   containersTests()
   individualsTests()
   samplesTests()
-  protocolsTests()
```
jleblond, line 13: Be careful not to remove lines inadvertedly :P

frontend/cypress/integration/all.spec.js
```diff
@@ -4,11 +4,12 @@ import { containersTests } from "./containers.spec"
 import { individualsTests } from "./individuals.spec"
 import { samplesTests } from "./samples.spec"
 import { protocolsTests } from "./protocols.spec"
+import { paginationTests } from "./pagination.spec"
 
 // Tests
 context('All tests', () => {
```
jleblond, line 10: A suggestion: since FreezeMan is growing, and we're adding more and more tests, I'd be great to be able to run some of the independent tests on their own (pagination test is the perfect example of that, because this test does not rely on other tests for their data, so it can be run alone)

PR #105, opened 2021-07-15
Unit tests for Experiment Run on the backend.

UlysseFG:
Work in progress

Add a section for experiment runs tests with coverage.


backend/fms_core/resources/experiment_run.py
```diff
@@ -91,42 +91,41 @@ def before_import(self, dataset, using_transactions, dry_run, **kwargs):
...
-                "source_sample_volume_used": float_to_decimal(blank_str_to_none(row.get("Instrument Name"))),
-                "experiment_container_position": row.get("Experiment Start Date"),
-             }
-            self.sample_rows.append(sample_row)
-
-            import_result = wipe_import_row_result(import_result, row)
+        row_id = row.get("#", "")
+        self.temporary_experiment_id = row.get("Experiment ID", None)
+        import_result = self.get_row_result_class()()
+        if row_id.isnumeric() and self.temporary_experiment_id: # Uses row ID to identify what are the data rows (compared to title or empty rows)
```
jleblond, line 107: While testing, I noticed that isnumeric only works on string. So this may work for the unit tests in csv file, but for the xlsx files that will be uploaded, row_id is actually an int. Therefore, I suggest checking for a condition such as is_row_id_number = isinstance(row_id, int) or (isinstance(row_id, str) and row_id.isnumeric()) for example

UlysseFG: casting the row_id to str to ensure this works...

PR #106, opened 2021-07-21
Added front end test for experiments

sebastianballesteros:
Addes test for experiments in the front end. I commented out the other tests in all.spec.js just for simplicity.


frontend/cypress/integration/all.spec.js
```diff
@@ -4,11 +4,13 @@ import { containersTests } from "./containers.spec"
...
 import { protocolsTests } from "./protocols.spec"
+import { experimentsTests } from "./experiments.spec"
 
 // Tests
 context('All tests', () => {
-  containersTests()
-  individualsTests()
-  samplesTests()
-  protocolsTests()
+  //containersTests()
```
jleblond, line 11: Don't forget to un-comment those lines before pushing

jleblond:
Looks good overall, however the integration tests for Experiment could be more exhaustive. When adding experiments, they should be displayed in the Experiment list page, but also we should click on a specific Experiment and check whether the details look ok (i.e. details about the experiment, samples tab). Also, a Container or a Sample used for an experiment will  have (in its details page) a tab for the Experiments it was used in.

jleblond (commented):
Be careful not to commit temporary ~$ files generated by your computer (~$Experiment_submission_v3_3_0 .... file)

Also, since we are running those tests on the test database, the container YT something from the Experiment template does not exist in our test database. Make sure we use containers already existing in our test database (i.e. those from the container fixtures, or if none are available, you can add some)

jleblond:
Same recommendation as in https://github.com/c3g/freezeman/pull/111#issuecomment-888354840

Also, there is one last place where the Experiments should be displayed, that is in the Container page, there is an Experiment tab too.

frontend/cypress/integration/experiments.spec.js
```diff
@@ -0,0 +1,119 @@
...
+        cy.get('body').should('contain', 'Sample_plasma1 sample (PLASMA) @ A04')
+        cy.get('body').should('contain', 'Sample_saliva1 sample (SALIVA) @ A05')
+        cy.get('body').should('contain', 'Sample_swab1 sample (SWAB) @ A06')
+        cy.get('body').should('contain', 'Sample_Blood1 sample (BLOOD) @ B03')
+      });
+
+      it('visits a particular sample detailed page to ensure the first experiment is there', () => {
+        cy.navigateTo('Samples')
+        cy.wait(WAIT_TIME)
+        cy.get('[data-row-key="27"] > :nth-child(2) > a > div').click()
```
jleblond, line 81: This one and the second one (with data-row-key="28") didnt work for me. I changed the values to 16 and 15 and it worked well for me. Assuming it did work for you with your keys 27 and 28, I wonder if these are row keys that are generated differently each time?

jleblond (approved):

PR #107, opened 2021-07-26
Validates ExperimentRun from template association to samples

jleblond:


backend/fms_core/resources/experiment_run.py
```diff
@@ -256,11 +264,14 @@ def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
 
             sample_data_errors = []
 
+
             try:
                 experiment_run = ExperimentRun.objects.get(id=self.experiments[data_experiment_id])
+                self.temporary_experiment_ids_without_samples.remove(data_experiment_id)
```
UlysseFG, line 270: I would move this line lower (line 319) where the sample is actually created which verify the condition we are trying to prevent. This would not cause problems to leave it here but it would make more sense later when we read the code and it would also lessen the chance something is added to the code that break the test by adding an alternate path in the code that lead to a state where no sample is created.

PR #109, opened 2021-07-27
Optional property in Experiments

jleblond:


UlysseFG (commented):
i think we should put some validation somewhere taking into account the is_optional field to ensure all non optional properties are entered ... you do it at the resource level but i wonder if this would not be better done at the model level... I don't see any easy way since process need to exist first for a property_value to be created on them. Maybe during the experiment_run clean() ... traverse the process and subprocess linked and make sure all required properties have values...

jleblond:
I dont think it belongs in the experiment_run model clean function... Im open to your suggestion, but we would have to find a more suitable place where to do those validations, and I think for now what we have is the most appropriate place where to put it.

PR #111, opened 2021-07-27
Fixed some tests that were failing due to timeouts or changes in pagi…

sebastianballesteros:
Modified some tests that were failing due to timeouts and pagination changes. 
I tested by running `npm run test:frontend:open` so it should be ok.


jleblond:
Looks good to me. Last recommendation: the wait time (I've seen you added it here and there in this PR and your other PR) I would put it in a variable, and use the same number across the cy.wait functions.

jleblond (approved):

PR #112, opened 2021-07-27
Add some file logging to the QC and DEV server env to track issues not found locally.

UlysseFG:


jleblond:
If I understand correctly, you log a more detailed file (with requests and queries) for environments that are not 'prod'. For prod, you log less information and it only appears in the console ?

UlysseFG:
exactly ... django logging have another filtering done on the DEBUG flag ... if DEBUG is true you get SQL logging else you do not get it. So essentially it will create 3 version of logging one on the console for Warnings and up for PROD one Debug and up on file without SQL for DEV and QC and one Debug and up with SQL for LOCAL...

jleblond:
That's fine for me, if this can help you with the issue :P

PR #118, opened 2021-08-03
Added the project model and its migration file

sebastianballesteros:
Some clarifications: 

1. - Added the project model according to use case ID num. 2 in : https://docs.google.com/spreadsheets/d/1MfdlnuIx0j1brnJZPiQmlibmduIdUSPvieawrdNLgjs/edit#gid=0
2. - Status_choices is just an example of what the choices could be. The real options are going to be given by Andres in the future, after being confirmed by stakeholders. 
3. - I added the Project field to Sample. 
4. - I was debating whether some fields should be mandatory or not but according to the link pasted in point num. 1,  the only 4 mandatory fields are name, creator, creation date and status
5. I added EmailValidator, although I don't know if my implementation makes sense
6. Feel free to add any comments, clarifications or modifications needed :)


backend/fms_core/models/sample.py
```diff
@@ -167,6 +168,9 @@ class Sample(TrackedModel):
     child_of = models.ManyToManyField("self", blank=True, through="SampleLineage",
                                       symmetrical=False, related_name="parent_of")
 
+    project = models.ForeignKey(Project, null=True, on_delete=models.PROTECT,
```
jleblond, line 171: To be removed from this PR; the association between sample and project has not be modelled yet, and it is not needed in this PR yet :)

backend/fms_core/models/project.py
```diff
@@ -0,0 +1,57 @@
...
+    requestor_email = models.CharField(blank=True, max_length=200,
+                                       help_text="The email of the requestor of the project.",
+                                       validators=[email_validator])
+
+    targeted_end_date = models.DateTimeField(blank=True, help_text="Targeted date to conclude the project.")
+
+    IN_PROCESS = "In Process"
+    FINISHED = "Finished"
+
+    STATUS_CHOICES = (
```
jleblond, line 35: I recommend using a similar approach to how choices for value_type in PropertyType are declared. In that way, we do not declare variables for each status that are then not useful on their own, but again this is a personal preference :P However, when there is a list of 'variables', it is often recommended in Python to put them in a dictionary or a list.

backend/fms_core/models/project.py
```diff
@@ -0,0 +1,57 @@
...
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+from ._utils import add_error as _add_error
+from ._validators import name_validator, email_validator
+
+__all__ = ["Project"]
+
+
+@reversion.register()
+class Project(TrackedModel):
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH,
```
jleblond, line 18: We agreed on having a unique name, dont forget to add this

backend/fms_core/migrations/0022_v3_4_0.py
```diff
@@ -1,48 +1,39 @@
...
     ]
 
     operations = [
-        # Migrations related to Project
         migrations.CreateModel(
             name='Project',
             fields=[
-                #Project Model
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('name', models.CharField(help_text='The name of the project.', max_length=200)),
```
sebastianballesteros: As requested, this file was created by running makemigrations. It did not need any modification.

backend/fms_core/models/sample.py
sebastianballesteros: Perfect! Already removed :)

backend/fms_core/models/project.py
sebastianballesteros: Ok, I used the approach used in individual, did not know the approach used in value_type was better. Changes are made ;)

backend/fms_core/models/project.py
sebastianballesteros: My bad

backend/fms_core/viewsets/project.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @action(detail=False, methods=["get"])
+    def summary(self, _request):
+        """
+        Returns summary statistics about the current set of projects in the
+        database.
+        """
+        return Response({
+            "total_count": Project.objects.all().count(),
+            "ongoing_count": Project.objects.all().filter(status='Ongoing').count(),
```
jleblond, line 48: Seems like you defined your status as being 'In Progress' in the model, not 'Ongoing' ;)

backend/fms_core/serializers.py
```diff
@@ -268,3 +271,13 @@ class Meta:
...
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = "__all__"
+
+class ProjectExportSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = "__all__"
```
jleblond, line 283: This is the data that would be in the export file
image

There are probably some irrelevant fields and/or some fields with foreign id's that could display information formatted in a more human-friendly way :P

frontend/src/components/projects/ProjectsFilters.js
```diff
@@ -0,0 +1,43 @@
...
+import FilterGroup from "../filters/FilterGroup";
+import {PROJECT_FILTERS} from "../filters/descriptions";
+
+const mapStateToProps = state => ({
+  filters: state.projects.filters,
+});
+
+const actionCreators = {setFilter};
+
+const ProjectsFilters = ({
```
jleblond, line 16: No need for detached filters for Projects for now. There are detached filters for Samples because there were some fields that we wanted to have filters on, but these fields were not part of any of the list columns. In the case of Projects, all those filters can be applied to the table columns directly :)

backend/fms_core/serializers.py
```diff
@@ -273,8 +273,6 @@ class Meta:
         depth = 1
 
 class ProjectSerializer(serializers.ModelSerializer):
-    targeted_end_date = serializers.DateTimeField(read_only=True, format="%Y-%m-%d")
```
sebastianballesteros: This was the line that was giving problems.

jleblond: Ok, then I would advise to format it properly in the front-end (I think there are examples in the front-end where dates are formatted to be more human-friendly) otherwise the date is displayed with Z 
image

frontend/src/components/projects/ProjectsPage.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+import {Redirect, Route, Switch} from "react-router-dom";
+
+import ProjectsListContent from "./ProjectsListContent";
+import ProjectEditContent from "./ProjectEditContent";
+import PageContainer from "../PageContainer";
+
+const ProjectsPage = () => <PageContainer>
```
jleblond, line 9: There is missing the route for individual projects (such as /projects/:id which would have the ProjectsDetailContent component for the detailed information about a specific project)

frontend/src/components/projects/ProjectsListContent.js
```diff
@@ -0,0 +1,175 @@
...
+    {
+      title: "Status",
+      dataIndex: "status",
+      sorter: true,
+      align: "right",
+      className: "table-column-numbers",
+      width: 115,
+    },
+    {
+      title: "Comments",
```
jleblond, line 83: We never (or almost never) add the 'Comments' fields in the columns of a list of items. In the event where we would have a very lengthy comment, the table would become ugly :P also the list should not display all the attributes of a Model, but the most significant ones

frontend/src/components/filters/descriptions.js
```diff
@@ -244,5 +244,31 @@ export const USER_FILTERS = {
     type: FILTER_TYPE.INPUT,
     key: "email",
     label: "Email"
-  }, 
+  },
+}
+
+export const PROJECT_FILTERS = {
```
jleblond, line 250: Could be great to have a filter on the request_date of a project (see Experiment start_date or Sample creation_date as examples of date range filter)

jleblond:
I am getting this error when I try to export the projects as an excel file
image

jleblond:
Also, it might be great to have the 'Add One project' shortcut link in the dashboard (to keep it uniform with the other entities dashboards)
image

jleblond:
Visually, I think it would look even better to have the different project status at the right of the box, such as in the Protocols box (which lists the different type of protocols at the right, and the total at the left)
image

backend/fms_core/models/project.py
```diff
@@ -0,0 +1,53 @@
...
+
+    principal_investigator = models.CharField(blank=True, max_length=200, help_text="The principal investigator of the project.")
+
+    requestor_name = models.CharField(blank=True, max_length=200, help_text="The name of the requestor of the project.")
+
+    requestor_email = models.CharField(blank=True, max_length=200,
+                                       help_text="The email of the requestor of the project.",
+                                       validators=[email_validator])
+
+    targeted_end_date = models.DateTimeField(blank=True, help_text="Targeted date to conclude the project.")
```
UlysseFG, line 32: This should be a date field.

backend/fms_core/models/sample.py
```diff
@@ -22,6 +22,7 @@
 from .container import Container
 from .individual import Individual
 from .sample_kind import SampleKind
+from .project import Project
```
UlysseFG, line 25: Is this needed ?

backend/fms_core/serializers.py
```diff
@@ -268,3 +271,20 @@ class Meta:
...
         fields = ("id", "name", "permissions")
         depth = 1
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = "__all__"
+
+class ProjectExportSerializer(serializers.ModelSerializer):
+    created_by = serializers.CharField(read_only=True, source="created_by.username")
```
UlysseFG, line 281: You can remove the tracking fields from the export serializer (created_by, updated_by, created_at, updated_at). I don't think this is information that someone would want exported.

backend/fms_core/migrations/0022_v3_4_0.py
```diff
@@ -0,0 +1,39 @@
...
+                ('deleted', models.BooleanField(default=False, help_text='Whether this instance has been deleted.')),
+                ('name', models.CharField(help_text='The name of the project.', max_length=200, unique=True, validators=[django.core.validators.RegexValidator(re.compile('^[a-zA-Z0-9.\\-_]{1,200}$'))])),
+                ('principal_investigator', models.CharField(blank=True, help_text='The principal investigator of the project.', max_length=200)),
+                ('requestor_name', models.CharField(blank=True, help_text='The name of the requestor of the project.', max_length=200)),
+                ('requestor_email', models.CharField(blank=True, help_text='The email of the requestor of the project.', max_length=200, validators=[django.core.validators.EmailValidator()])),
+                ('targeted_end_date', models.DateTimeField(blank=True, help_text='Targeted date to conclude the project.')),
+                ('status', models.CharField(choices=[('Ongoing', 'Ongoing'), ('Completed', 'Completed')], help_text='The status of the project.', max_length=20)),
```
UlysseFG, line 30: If we do not get a clear list of status from the users, i would add personally an "On-Hold" status.

sebastianballesteros: So we would have Ongoing, Completed and On-Hold?

backend/fms_core/serializers.py
sebastianballesteros: Perfect!

backend/fms_core/migrations/0022_v3_4_0.py
UlysseFG: Maybe this is something we should discuss once we have the users feedback ... i just noted this here to share and see what you and Jasmine think about this.

frontend/src/components/projects/ProjectsListContent.js
```diff
@@ -0,0 +1,167 @@
...
+      title: "Principal Investigator",
+      dataIndex: "principal_investigator",
+      sorter: true,
+      width: 70,
+    },
+    {
+      title: "Requestor Name",
+      dataIndex: "requestor_name",
+      sorter: true,
+      align: "right",
```
UlysseFG, line 54: Text is usually aligned left (default) and the className do not seem appropriate.

backend/fms_core/models/project.py
sebastianballesteros: @jleblond good that I changed it 👍

frontend/src/components/projects/ProjectEditContent.js
```diff
@@ -0,0 +1,178 @@
...
+  );
+}
+
+function deserialize(values) {
+  if (!values)
+    return undefined
+  const newValues = {...values}
+
+  if (newValues.creation_date)
+    newValues.creation_date = moment(newValues.creation_date, 'YYYY-MM-DD')
```
UlysseFG, line 165: You are unlikely to ever get creation_date here ... Maybe this can apply to targeted_end_date instead ...

frontend/src/components/projects/ProjectEditContent.js
```diff
@@ -0,0 +1,178 @@
...
+  return newValues
+}
+
+function serialize(values) {
+  const newValues = {...values}
+
+  if (newValues.creation_date)
```
UlysseFG, line 172: Same as for deserialize ...

backend/fms_core/models/sample.py
jleblond: Right, @sebastianballesteros this can be removed from the sample model because we don't deal with projects there yet

frontend/src/utils/api.js
```diff
@@ -78,7 +78,15 @@ const api = {
...
   },
-
+  projects: {
+    get: projectId => get(`/projects/${projectId}/`),
+    add: project => post("/projects/", project),
+    update: project => patch(`/projects/${project.id}/`, project),
+    list: (options, abort) => get("/projects", options, { abort }),
+    listExport: options => get("/project/list_export/", {format: "csv", ...options}),
+    summary: () => get("/projects/summary"),
+    search: q => get("/projets/search/", { q }),
```
jleblond, line 88: Where is this used? I could be mistaken, but I think that was part of the search viewset.. which is not implemented for projects? (from what I remember that was used for the global searchbar at the top left menu in the interface)

jleblond:
I noticed there aren't unit tests/integration tests for this one. So just adding unit tests for the project model, and integration tests (testing that the form for adding projects, and the lists of projects is fine) would be great! (we usually add the model unit tests when creating a new model)

frontend/src/utils/api.js
UlysseFG: Would we want the projects to be searched by the Quick search ? I think if someone type Potatosalad it should point them to the Project if it exists ...

sebastianballesteros: I think that's done directly in the table with the table filters.

jleblond: We are talking two different things here. The filters are done automaticallyon the table columns.

We have another functionality which is the jump bar search (general search across the entities of the website). 
image


What Ulysse meant is that it would be useful to add project search (by name) to that general jump bar. This is done with the search action in the viewset/search endpoint, and some hooking (would have to check exactly how it is done)

jleblond:
Just to make sure we use the same terminology, what you have just committed is our front-end 'integration' tests with Cypress.

What we mean by unit tests is what's in backend/fms_core/tests. As you can see, there are 2 main type of tests there, tests for models (in the folder test_models) and resource tests (test_resources.py; which will eventually go away and will be replaced by some other tests with the new import tool)

jleblond (approved):
Looks good, you can merge it to master :)

PR #123, opened 2021-08-05
Import tool prototype

jleblond:
WIP

* Demo so you can follow my progress and give me feedback*


backend/fms_core/viewsets/_utils.py
```diff
@@ -104,30 +111,45 @@ def template_check(self, request):
...
-                "validation_error": r.validation_error,
-                "warnings": r.warnings,
-                "diff": r.diff,
-                "import_type": r.import_type,
-            } for r in result.rows if r.import_type != RowResult.IMPORT_TYPE_SKIP],
-        })
+
+        # result = resource_instance.import_data(dataset, dry_run=True)
+
+        eri = ExperimentRunImporter(file=file, format='xlsx')
```
UlysseFG, line 120: The importer should be called generically (RunImporter) the correct instance should be created using the input.

backend/fms_core/import_tool/_utils.py
```diff
@@ -0,0 +1,8 @@
+from pandas import pandas as pd
+
+def data_row_ids_range(starting_row, df):
```
UlysseFG, line 3: This is too low level to my taste. You should be able to fetch the needed column using the structure and methods provided by panda.

jleblond: it is not fetching the columns.. its just saying that this section is from Row X to Row Y and generating a Range() for it, so we can iterate through it. It's an helper method i create so we can calculate  the rows from the initial header to the end of the dataframe according to its length.

backend/fms_core/import_tool/handlers/experiment_run.py
```diff
@@ -0,0 +1,67 @@
...
+                self.experiment_run['instrument'] = Instrument.objects.get(name=instrument_name)
+            except Exception as e:
+                self.errors["instrument"] = f"No instrument named {instrument_name} could be found."
+
+        # Get or Create Container
+        barcode = container['barcode']
+        kind = container['kind']
+        if barcode and kind:
+            try:
+                self.experiment_run['container'], _ = Container.objects.get_or_create(
```
UlysseFG, line 31: Experiment_run should always create its container ... You don't really want a recycled container...

jleblond: Same thing is done in the resource ExperimentRun...

backend/fms_core/import_tool/handlers/experiment_run.py
```diff
@@ -0,0 +1,67 @@
...
+from ...services import create_experiment_run
+
+
+
+
+class ExperimentRunHandler(GenericHandler):
+    def __init__(self, experiment_type_obj, instrument, container, start_date,
+                 sample_rows, properties, protocols_dict, properties_by_name_dict):
+
+        self.experiment_run = {'experiment_type': experiment_type_obj, 'instrument': None, 'container': None,
```
UlysseFG, line 15: Does this have to be a class attribute ? as far as i see this is local storage. Not sure why you even copy all of this locally...

backend/fms_core/import_tool/handlers/experiment_run.py
```diff
@@ -0,0 +1,67 @@
...
+
+from fms_core.models import ExperimentType, PropertyType, Instrument, Container, Sample
+from ._generic import GenericHandler
+from ...services import create_experiment_run
+
+
+
+
+class ExperimentRunHandler(GenericHandler):
+    def __init__(self, experiment_type_obj, instrument, container, start_date,
```
UlysseFG, line 12: Is it normal that all of this happen during the init ? The class is a big init (more or less)?

backend/fms_core/import_tool/importers/experiment_run.py
```diff
@@ -0,0 +1,138 @@
...
+
+class ExperimentRunImporter(GenericImporter):
+    base_errors = []
+    global_data = {'experiment_type': None, 'protocols_dict': {}, 'property_types_by_name': {}}
+
+    def __init__(self):
+        self.sheet_names = ['Experiments', 'Samples']
+        super().__init__()
+
+    # has to be declared ?? or can directly inherit from its parent?
```
UlysseFG, line 18: if you omit the function it will call the parent one ...

backend/fms_core/import_tool/handlers/experiment_run.py
jleblond: yeah, I dont mind creating a __init__ with no args, and passing this all to a method instead (in a similar way to what im doing for the Importer) if you prefer

backend/fms_core/import_tool/handlers/experiment_run.py
jleblond: Yeah, for now it doesnt have to be a class attribute. I thought of generating the attributes automatically in the Handler objects at some point (and the idea is not totally dropped; I wanted to try with other templates to see how it would turn out), but for now it is not necessary.

backend/fms_core/import_tool/importers/experiment_run.py
```diff
@@ -0,0 +1,138 @@
...
+                    name=property_column)
+            except Exception as e:
+                self.base_errors.append(f"Property Type {property_column} could not be found")
+
+
+    def import_template_inner(self):
+        """
+            SAMPLES SHEET
+        """
+        samples_sheet = self.sheets['Samples']
```
UlysseFG, line 47: I would move most variables to local variables and pass them as method params ... The reason in import export we were relying on the class attribute was that we could not modify class signature without potential consequences ... We do not have this restriction here. Local variables are always more desirable than global ones. All template will have sheets you can pass sheets to import_template_inner...

backend/fms_core/import_tool/importers/experiment_run.py
```diff
@@ -0,0 +1,138 @@
...
+from ._generic import GenericImporter
+from fms_core.import_tool.handlers import ExperimentRunHandler
+from .._utils import (data_row_ids_range,
+                      convert_property_value_to_str,
+                      blank_and_nan_to_none)
+
+
+class ExperimentRunImporter(GenericImporter):
+    base_errors = []
+    global_data = {'experiment_type': None, 'protocols_dict': {}, 'property_types_by_name': {}}
```
UlysseFG, line 12: global_data the name itself give me shivers ... This is unlikely to be the right way to deal with anything :) From what i see most of this can be made local and passed around as params.

jleblond: I can change the name (such as preloaded_data or so) but I want it to be a class variable to illustre the notion of it being related to the template, not to a specific row.

backend/fms_core/import_tool/importers/experiment_run.py
```diff
@@ -0,0 +1,138 @@
...
+                      'experiment_id': blank_and_nan_to_none(row_data['Experiment ID']),
+                      'container_barcode': blank_and_nan_to_none(row_data['Source Container Barcode']),
+                      'container_coordinates': blank_and_nan_to_none(row_data['Source Container Position']),
+                      'volume_used': blank_and_nan_to_none(row_data['Source Sample Volume Used']),
+                      'experiment_container_coordinates': blank_and_nan_to_none(row_data['Experiment Container Position'])}
+            sample_rows_data.append(sample)
+
+
+        """
+            EXPERIMENTS SHEET
```
UlysseFG, line 65: Don't fear to break it down in functions / methods Experiment and sample sheets could have their own little function.

backend/fms_core/import_tool/importers/experiment_run.py
jleblond: Sheets and sheet_names have to be class variables in order for them to be handled automatically by the GenericImporter (which automatically slice the sheets, read them and store them), rather than repeating this section of code for each Importer.

backend/fms_core/import_tool/importers/experiment_run.py
jleblond: It is not very modular; which is why i didnt divide it into functions/methods. The only reason why it would be divided would be for better readability. I think those multiple line comments (with more information to be added) will be enough.

backend/fms_core/import_tool/importers/experiment_run.py
```diff
@@ -0,0 +1,138 @@
...
+        samples_sheet = self.sheets['Samples']
+        samples_row_header = 2
+        samples_sheet.columns = samples_sheet.values[samples_row_header]
+
+        sample_rows_data = []
+
+        for row_id in data_row_ids_range(samples_row_header + 1, samples_sheet):
+            row_data = samples_sheet.iloc[row_id]
+            sample = {'row_id': row_id,
+                      'experiment_id': blank_and_nan_to_none(row_data['Experiment ID']),
```
UlysseFG, line 56: blank_and_nan_to_none ... this really has a nice ring to it... I would do this on each single data cell before you get started with anything ... you extract each sheet payload region and you apply this to every single cell. then you can be safe that later if you add a field it will get the same treatment instead of having to remember each time to check for blank and nan ...

jleblond: good idea, Ill have to see how I could do it while keeping good performance

backend/fms_core/models/experiment_type.py
```diff
@@ -34,4 +49,19 @@ def add_error(field: str, error: str):
     def save(self, *args, **kwargs):
         # Normalize and validate before saving, always!
         self.full_clean()
-        super().save(*args, **kwargs)  # Save the object
\ No newline at end of file
+        super().save(*args, **kwargs)  # Save the object
+
+    @property
+    def get_protocols_dict(self):
```
UlysseFG, line 55: This will likely be a method of workflows at some point ... I would comment it so we do not forget to relocate it someday ...

backend/fms_core/import_tool/handlers/experiment_run.py
UlysseFG: This just seem a bit unusual to have most of the actions performed by a class done in its init.

backend/fms_core/import_tool/importers/experiment_run.py
UlysseFG: I see a reason for sheet_names but you can just pass sheets to import_template_inner. I would replace individual Sheet_name with a const to make sure all reference to them are correct.

backend/fms_core/services/experiment_run.py
```diff
@@ -0,0 +1,170 @@
...
+            PropertyValue.objects.create(value=str(value),
+                                         property_type=property_type,
+                                         content_object=process)
+        except Exception as e:
+            errors[property] = e.error_dict['value']
+
+    return {'errors': errors}
+
+
+def handle_samples(experiment_run, samples):
```
UlysseFG, line 87: Handle is often a term used to described an ID that point to a an object ... Handle is also a very general term to describe pretty much anything we do on something. Is there a more precise way to describe what is happening here ?

backend/fms_core/services/experiment_run.py
```diff
@@ -0,0 +1,170 @@
...
+    processes_by_protocol_id = {}
+    for protocol in protocols_objs_dict.keys():
+        parent_process = Process.objects.create(protocol=protocol, comment="")
+        for subprotocol in protocols_objs_dict[protocol]:
+            sp = Process.objects.create(protocol=subprotocol,
+                                        parent_process=parent_process,
+                                        comment="Experiment")
+            processes_by_protocol_id[subprotocol.id] = sp
+
+    return (parent_process, processes_by_protocol_id)
```
UlysseFG, line 63: What happens if there is 2 parent processes ?

jleblond: could be renamed to top_process... an experiment run has only 1 process, so it must be the top_process

UlysseFG:
I would say the thing that does worry me the most is the structure and dedicated responsibilities of each part.
I was expecting all the creation and DB interaction to be relegated to the Service layer but there is DB stuff that are dealt with at the handler level too. There is the importer than import the template data but also iterate on the resulting data to produce results. There is no clear concept behind each object.

Since we need to report back to the user using the structure submitted we cannot totally ignore the submitted structure. 

This is an idea but i would segregate more the job of each part. Do not tell me i told you so ... i realize my position is wobbly. I would propose the idea of a model like the one below instead of the more linear Importer => handler => Service..

template file -- *Importer* -- Confirmation Table + errors            < == iterative layer between sheets(if needed). holds global vision.
                         |        |                                                                         Deals with global or non-sheet specific questions.
    *DataExtracter*     *DataInserter*      < == iterative layer within a sheet. Returns by sheet with row annotation.
                                                                      extract_experiment_runs(), extract_run_samples()
                                                                      create_experiment_runs(), create_run_samples()

This is not that clear to me what need to be done that is my general feeling though ...

jleblond:
Importer is about ingesting the data from the dataframe and putting it in a more organized structure.
The Handler will be renamed RowHandler, it handles only one row, and it creates the foreign objects, or does any manipulation needed so the data can be passed to the Service layer.

The Service layer deals with a more complex flow of interrelated object creations. If we only want to create a single container without more complexity, this can be done directly in the (Row)Handler. The Service layer is totally optional, and is to be used only to deal with complexity. That's how I see it.

backend/fms_core/import_tool/importers/_generic.py
```diff
@@ -0,0 +1,86 @@
...
+            self.base_errors.append(e)
+            print('Importers/Generic create_sheet_data exception ', e)
+            return None
+
+
+    def preload_data_from_template(self, **kwargs):
+        pass
+
+    @property
+    def is_valid(self):
```
UlysseFG, line 78: Not perfectly clear what is valid ... The values in the template ? the importer itself ? Maybe define the name a bit more.

backend/fms_core/services/process.py
```diff
@@ -0,0 +1,22 @@
+from django.core.exceptions import ValidationError
+from fms_core.models import Process
+
+def create_processes_for_experiment_from_protocols_dict(protocols_objs_dict,
```
UlysseFG, line 4: Could this simply be create_process_hierarchy_from_protocol ? This may be more inviting to reuse it. I'm not sure about the top process naming ...

jleblond: You didnt like the parent process, I suggest top process unless you have a better idea.

Dict is more clear that were not dealing with straight up objects

backend/fms_core/import_tool/sheet_data.py
```diff
@@ -0,0 +1,75 @@
...
+    def prepare_rows(self):
+        self.rows = []
+        self.rows_results = []
+        for row_id in data_row_ids_range(self.header_row_nb + 1, self.dataframe):
+            row_data = self.dataframe.iloc[row_id]
+
+            print('row_data', row_data)
+
+            required_values = [row_data[key] for key in self.minimally_required_columns]
+            if any(list(map(lambda x: x is None, required_values))):
```
UlysseFG, line 39: All the lambda/map stacking are ugly and very poorly readable ... i would make functions out of them to at least convey the general idea of what is performed. Python list comprehensions are much more readable and can usually be used instead of lambda.

jleblond: Any/all are more pythonic and more efficient. They will break out the first time they encounter a value that does not respect the condition (as opposed to the list comprehension which will go through all the items)

backend/fms_core/viewsets/_utils.py
```diff
@@ -141,14 +126,16 @@ def template_submit(self, request):
...
-        resource_instance = action_def["resource"]()
-        result = resource_instance.import_data(dataset)
+        importer_instance = action_def["importer"]()
 
-        if result.has_errors() or result.has_validation_errors():
-            # TODO: Better message
-            return HttpResponseBadRequest(json.dumps({"detail": "Template errors encountered in submission"}),
-                                          content_type="application/json")
+        try:
+            importer_instance.import_template(file=file, format='xlsx', dry_run=False)
```
UlysseFG, line 134: What is the plan for csv handling ?

jleblond: Will be added with tests (If even tests are still in csv)

backend/fms_core/import_tool/sheet_data.py
UlysseFG: Any/all is the not the problem... the lambda is ... It's been introduced to many languages but it's basically a way to obfuscate your code. This one is not so bad i have seen worse though ... the idea being try the comprehensions to build your list and still use any if you want ... the efficiency gain is negligible here compared to the pain it is on the eye ... ;)

jleblond: Pain in the eye and in the variable names is always subjective ;) but ok I get it, you want me to keep any/all but change the lambda for a list comprehension

jleblond:
Continued at https://github.com/c3g/freezeman/pull/136 with fixes

PR #126, opened 2021-08-11
Added tube box 7x7

sebastianballesteros:
Should be straightforward. However, I might be missing something.


UlysseFG:
You also need to add the new container kind to the xlsx Templates that require container kind in one of their drop down...
Container Creation, Sample Submission, Sample Extraction, Sample Transfer, off the top of my head. To be perfectly clean you should also mirror these template changes to the corresponding test templates. Update the changed template to reflect the release version (3.4 atm).

UlysseFG:
You should update the cypress test template too to match the version name and add the new container kind to the options. This won't change anything but doing this ensure we keep the test templates matching to the real ones even for changes that are unlikely to cause any trouble.

frontend/cypress/integration/containers.spec.js
```diff
@@ -2,7 +2,7 @@
 
 // Fixtures:
 //  - credentials.json
-//  - Container_creation_v3_2_0_F_A_2.xlsx
+//  - Container_creation_v3_4_0_F_A_2.xlsx
```
UlysseFG, line 5: For the naming, the "2" was intended as an intra-version counter to differentiate between test template versino within a version of the template... It should reset to 1 if you change the template version (vx_y_z part).

PR #127, opened 2021-08-11
Displayed ID in list, detailed page and export

sebastianballesteros:
Clarifications: 

- Should ID be a sorter?


jleblond:
I think this should apply to any entity page (container, sample, process, experiment etc)

Also, sorter and filter could be a +

sebastianballesteros:
I agree that there should be consistency. Process and experiments already have an ID so its only container and sample (and potentially users) that need in the interface list. 

I will add the filter.

UlysseFG:
Yeah ID could / should be everywhere but i am not sure if multiplying available unique identifier for containers will be helpful, but we can go ahead and extend it to all entities i think ...

UlysseFG:
Ok looks like this is pretty nice.
my only reservations : 
1) Filtering on the ID do not seem to be working,
2) The list header is very crowded. I would strip the "Sample" from "Sample ID" and "Sample Kind" to leave some breathing space there, but only in the list.

sebastianballesteros:
I am trying to make everything consistent. For the tables I will leave just ID without the entity name i.e. "Sample ID" or "Individual ID". Should I get rid of the "Sample Kind" entry?
I also noticed the ID filtering does not work. The filtering on "Process ID" works perfectly, I am trying to find any differences that may cause the problem.

jleblond:
Are there viewsets filters for IDs? 

I agree with using ID uniformly across the tables. I would keep the sample kind in the list its relevant. But if its really too crowded maybe use ‘kind’ only?

UlysseFG:
> Should I get rid of the "Sample Kind" entry?

Don't get rid of the column, i would just display "Kind" for the header ... Since we are on a Sample page this may be sufficient though we have also have container info on the page ... I would think "Kind" is enough but its not that important anyway ...

UlysseFG:
My impression is this likely broke a few cypress tests that were relying on list column order ... This might be worth a check on that side ...

UlysseFG:
Are you sure you caught all broken tests ? The protocol.spec seem to have a test on column 3 which should no longer target the right column.

sebastianballesteros:
You are right,  the tests were not failing because source sample and generated sample were the same. But that may not be always the case, right? I will modify that one and one that I just spotted in samples.spec

UlysseFG (approved):
This looks complete to me ... just need to fix the merge ...

PR #128, opened 2021-08-11
Removed the requirement for a sample to have an individual tied to it

sebastianballesteros:
Clarifications: 

- In order to create an individual you need Taxon, Sex, or Name. If any of these is missing then the individual is not created. However, let's say that the user inputs the Taxon in the template (user knows is Homo sapiens but knows nothing about the individual) the taxon won't be included/recorded in the sample cause the individual was not created. Is this the behaviour we want?
- In addition, I had to get rid of template error messages concerning the individual field
- I added the related migration. However, if we want this to be in v3.4 then we would need to add this specific AlterField to the migration in the add-projects branch.


backend/fms_core/resources/sample.py
```diff
@@ -175,7 +175,6 @@ def import_obj(self, obj, data, dry_run):
         except Exception as e:
             individual_created = False
             individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
```
jleblond, line 178: If it's ok to not have an individual associated to a sample, an exception should not be thrown at all for that. So I would put a condition before attempting to get/create an individual

jleblond:
It would be good practise to add unit tests that confirm that creating a sample without an individual is ok.

UlysseFG:
>     * In order to create an individual you need Taxon, Sex, or Name. If any of these is missing then the individual is not created. However, let's say that the user inputs the Taxon in the template (user knows is Homo sapiens but knows nothing about the individual) the taxon won't be included/recorded in the sample cause the individual was not created. Is this the behaviour we want?

If user input Taxon they expect it to be saved. So individual becomes required ... We need to raise an error that explains that the user need all require all individual fields to complete the submission.

>     * In addition, I had to get rid of template error messages concerning the individual field

There is still cases where individuals will be expected to be present (see my above comment).

>     * I added the related migration. However, if we want this to be in v3.4 then we would need to add this specific AlterField to the migration in the add-projects branch.

Create your migration with the same name we'll merge the migrations together when we merge the Project PR...

backend/fms_core/resources/sample.py
UlysseFG: I would test if there is an individual id i would only try to create or get an individual if there is one (i dislike using exception to deal with known rules). If there is no individual ID but there is other individual column populated(cohort, pedigree, mother, father, taxon, sex), i would raise an error saying the user need to provide an individual ID if said field(cohort, pedigree, mother, father, taxon, sex) is provided to be saved.

backend/fms_core/resources/sample.py
```diff
@@ -161,21 +161,29 @@ def import_obj(self, obj, data, dry_run):
...
+                    **({"cohort": cohort} if cohort else {}),
+                    **({"mother": mother} if mother else {}),
+                    **({"father": father} if father else {}),
+                )
+                obj.individual = individual
+            except Exception as e:
+                individual_created = False
+                errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+        elif data["Individual ID"] or data["Sex"] or taxon and not individual_is_complete:
+            errors["individual"] = ValidationError("In order to save the individual's information all the fields are required (Taxon, Sex and ID)", code="invalid")
```
UlysseFG, line 184: i would replace "... all the fields ..." by " ... these fields ...". In the context of the template, all the fields are not required. Also ID is not clear since on the template this info is in a "individual name" column.

backend/fms_core/resources/sample.py
```diff
@@ -161,21 +161,29 @@ def import_obj(self, obj, data, dry_run):
...
+                    **({"pedigree": pedigree} if pedigree else {}),
+                    **({"cohort": cohort} if cohort else {}),
+                    **({"mother": mother} if mother else {}),
+                    **({"father": father} if father else {}),
+                )
+                obj.individual = individual
+            except Exception as e:
+                individual_created = False
+                errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+        elif data["Individual ID"] or data["Sex"] or taxon and not individual_is_complete:
```
UlysseFG, line 183: You should include the other individual fields for this one test (cohort, pedigree, mother, father) if for some reason those fields are set, this condition applies even if those fields are not required (not part of individual_is_complete).

backend/fms_core/resources/sample.py
```diff
@@ -161,21 +161,31 @@ def import_obj(self, obj, data, dry_run):
...
-            individual_created = False
-            individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+
+        # If there's the individual's complete information then get/create
+        individual_is_complete = True if data["Individual ID"] and data["Sex"] and taxon else False
+        partial_individual = True if data["Individual ID"] or data["Sex"] or mother or father or taxon or pedigree or cohort else False
+        individual = None
+
+        if individual_is_complete:
```
UlysseFG, line 170: There is the case also where the individual already exists in which case the individual ID would be sufficient i think...
Something like
individual_already_exists = ID and not (all other individual fields)
if individual_already_exists or individual_is_complete :

sebastianballesteros: In that case it would be necessary to make sure the individual actually exists by checking in the DB and not just ensuring the ID is provided, right?

UlysseFG: Yeah if you want to provide a clear error message you would need to break down the get_or_create into a get and a create otherwise the error returned will be more generic (as it is atm).

backend/fms_core/resources/sample.py
```diff
@@ -161,21 +161,31 @@ def import_obj(self, obj, data, dry_run):
...
-            )
-            obj.individual = individual
-        except Exception as e:
-            individual_created = False
-            individual = None
-            errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
+
+        # If there's the individual's complete information then get/create
+        individual_is_complete = True if data["Individual ID"] and data["Sex"] and taxon else False
+        partial_individual = True if data["Individual ID"] or data["Sex"] or mother or father or taxon or pedigree or cohort else False
```
jleblond, line 167: Not very Pythonic :P I'm pretty sure you can do this with 'any' or 'all'

PR #130, opened 2021-08-19
Implemented batch filtering for samples, containers and individuals

sebastianballesteros:
Please leave the comments in this new PR.

### DIFF
--- a/backend/fms_core/filters.py
+++ b/backend/fms_core/filters.py
@@ -42,6 +42,14 @@ class ContainerFilter(GenericFilter):
 
 class FullSampleFilter(GenericFilter):
     name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+    container__barcode = django_filters.CharFilter(field_name="container__barcode", method="batch_container_barcode_filter")
+
+    def batch_container_barcode_filter(self, queryset, name, value):
+        query = Q()
+        for v in value.split(" "):
+            query |= Q(container__barcode=v)
+        query_set = queryset.filter(query)
+        return query_set
 
     class Meta:
         model = FullSample
--- a/backend/fms_core/filters.py
+++ b/backend/fms_core/filters.py
@@ -31,13 +31,13 @@ class ContainerFilter(GenericFilter):
     barcode = django_filters.CharFilter(field_name="barcode", method="batch_barcode_filter")
     name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
 
-class Meta:
-    model = Container
-    fields = {
-        **_container_filterset_fields,
-        **_prefix_keys("location__", _container_filterset_fields),
-        **_prefix_keys("samples__", _sample_minimal_filterset_fields),
-    }
+    class Meta:
+        model = Container
+        fields = {
+            **_container_filterset_fields,
+            **_prefix_keys("location__", _container_filterset_fields),
+            **_prefix_keys("samples__", _sample_minimal_filterset_fields),
+        }
 
 class SampleFilter(GenericFilter):
     name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
@@ -49,6 +49,6 @@ class SampleFilter(GenericFilter):
 class IndividualFilter(django_filters.FilterSet):
     name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
 
-class Meta:
-    model = Individual
-    fields = _individual_filterset_fields
+    class Meta:
+        model = Individual
+        fields = _individual_filterset_fields
--- a/backend/fms_core/filters.py
+++ b/backend/fms_core/filters.py
@@ -1,0 +1,58 @@
+from django.db.models import Q
+
+from .models import Container, Sample, Individual
+
+import django_filters
+
+from .viewsets._constants import (
+ _container_filterset_fields,
+ _sample_filterset_fields,
+ _individual_filterset_fields,
+)
+
+class ContainerBatchFilter(django_filters.FilterSet):
+ barcode = django_filters.CharFilter(field_name="barcode", method="batch_barcode_filter")
+ name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+ class Meta:
+     model = Container
+     fields = _container_filterset_fields
+
+ def batch_barcode_filter(self, queryset, name, value):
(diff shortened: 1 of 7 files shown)


frontend/src/utils/serializeFilterParams.js
```diff
@@ -7,6 +7,10 @@ import {FILTER_TYPE} from "../constants"
 export default function serializeFilterParams(filters, descriptions) {
   const params = {}
 
+  function hasWhiteSpace(s) {
+    return /\s/g.test(s);
```
UlysseFG, line 11: This may not be the most targeted this can be... it would be better to check specifically for space instead of generic white space (including tab and newline). If your filter just split by " " you should test for that ... or change the filter to accept the other white space characters as separators...

frontend/src/utils/serializeFilterParams.js
```diff
@@ -38,23 +42,25 @@ export default function serializeFilterParams(filters, descriptions) {
...
+        if(!isBatch){
+          if (options) {
+            if (options.recursiveMatch)
+                key += "__recursive"
+            else if (options.exactMatch)
+                key += "__startswith"
+            else
+                key += "__icontains"
+        } else{
+           key += "__icontains"
```
UlysseFG, line 56: I don't see icontains or containts as a good batch comparison. Most likely this will be used to extensively list all desired barcodes or names not an approximation of each (this could be ... i'm just not sure this would be the expected result)

backend/fms_core/filters.py
```diff
@@ -0,0 +1,58 @@
...
+ _individual_filterset_fields,
+)
+
+class ContainerBatchFilter(django_filters.FilterSet):
+ barcode = django_filters.CharFilter(field_name="barcode", method="batch_barcode_filter")
+ name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+ class Meta:
+     model = Container
+     fields = _container_filterset_fields
```
UlysseFG, line 19: Are the fields set by the filterset_fields shortcut in the viewset ?

backend/fms_core/filters.py
```diff
@@ -0,0 +1,58 @@
...
+
+import django_filters
+
+from .viewsets._constants import (
+ _container_filterset_fields,
+ _sample_filterset_fields,
+ _individual_filterset_fields,
+)
+
+class ContainerBatchFilter(django_filters.FilterSet):
```
UlysseFG, line 13: I would not define the whole filterset by the case of the 2 fields that are batched ... I would name it ContainerFilter ... unless you plan to have more than one filterset for each model.

backend/fms_core/filters.py
```diff
@@ -0,0 +1,65 @@
...
+
+ def batch_barcode_filter(self, queryset, name, value):
+     query = Q()
+     for barcode in value.split(" "):
+         query |= Q(barcode__contains=barcode)
+     return queryset.filter(query)
+
+ def batch_name_filter(self, queryset, name, value):
+     query = Q()
+     for name in value.split(" "):
```
jleblond, line 36: Maybe give it a different name than 'name' in 'for name in value.split(" ")' ? 

Very quickly I thought the expression Q(name__icontains=name) contained the 'name' arg from the batch_name_filter args (self, queryset, name, value), but those are different 'name' values.

backend/fms_core/filters.py
```diff
@@ -0,0 +1,65 @@
...
+         query |= Q(name__contains=name)
+     return queryset.filter(query)
+
+class SampleFilter(django_filters.FilterSet):
+ barcode = django_filters.CharFilter(field_name="barcode", method="batch_name_filter")
```
jleblond, line 41: Just realized by checking out your branch in my editor that it only has 1 space indentation? same goes for all your function definitions within filter classes

jleblond:
Regarding my suggestion of having re-usable methods. Here is an example for one possible solution: 

image

frontend/src/utils/serializeFilterParams.js
jleblond: I agree with Ulysse, if we do a batch search, we expect an exact match. Also, ideally, it should be reflected in the interface that the search is 'exact' when we do a batch search.

frontend/src/utils/serializeFilterParams.js
```diff
@@ -38,23 +42,25 @@ export default function serializeFilterParams(filters, descriptions) {
...
-          if (options.recursiveMatch)
-            key += "__recursive"
-          else if (options.exactMatch)
-            key += "__startswith"
-          else
+        const isBatch = description.batch && hasSpaces(value)
+
+        if(!isBatch){
+          if (options) {
+            if (options.recursiveMatch)
```
jleblond, line 49: Pure stylistic comment, would there be a way to re-organize a bit the logic to have less nested 'if' statements?

backend/fms_core/filters.py
```diff
@@ -0,0 +1,54 @@
...
+        for v in value.split(" "):
+            query |= Q(name=v)
+        return queryset.filter(query)
+
+
+class ContainerFilter(GenericFilter):
+    barcode = django_filters.CharFilter(field_name="barcode", method="batch_barcode_filter")
+    name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+class Meta:
```
jleblond, line 34: Indentation?

backend/fms_core/filters.py
```diff
@@ -0,0 +1,54 @@
...
+    name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+    class Meta:
+        model = Sample
+        fields = _sample_filterset_fields
+
+class IndividualFilter(django_filters.FilterSet):
+    name = django_filters.CharFilter(field_name="name", method="batch_name_filter")
+
+class Meta:
```
jleblond, line 52: Same as above. Shouldnt this be indented?

UlysseFG:
I tried to run this locally and the containers do not seem to be loading (both on the container list and on the sample list). This is not working at the moment. Do we want to get this to work for 3.4 ?

jleblond:
@UlysseFG what error are you getting?

@sebastianballesteros don't forget to have the 'class Meta' under their respective Filter classes. For now some of them are not indented

PR #132, opened 2021-08-19
Addition of model and resource to link sample and project.

UlysseFG:
TODO 

- [x] Create model sample_by_project.
- [x] Create resource and template for updating sample_by_project.
- [x] Add template submission to frontend and backend.
- [x] Add unit tests for sample_by_project and integration tests for template submission.
- [x] Change sample submission template and resource to accept project name as optional field.
- [x] Add display of samples on the Project page.
- [x] Add Projects column (with filter) to samples in sampleList


backend/fms_core/models/sample.py
```diff
@@ -167,6 +168,8 @@ class Sample(TrackedModel):
     child_of = models.ManyToManyField("self", blank=True, through="SampleLineage",
                                       symmetrical=False, related_name="parent_of")
 
+    projects = models.ManyToManyField("project", blank=True, through="SampleByProject", related_name="samples")
```
UlysseFG, line 171: I don't think you need to create ManyToMany relationship on both end of the relationship. One side should be enough to provide the same result (either on the sample or the project)... The related_name will point to the other side of the relationship...

backend/fms_core/serializers.py
```diff
@@ -273,12 +274,16 @@ class Meta:
         depth = 1
 
 class ProjectSerializer(serializers.ModelSerializer):
+    samples = SampleSerializer(read_only=True, many=True)
```
UlysseFG, line 277: I think simply having the sample pk would be more appropriate than the whole samples ...

backend/fms_core/serializers.py
```diff
@@ -273,12 +274,16 @@ class Meta:
...
         model = Project
         fields = "__all__"
+        extra_fields = ('samples')
 
 class ProjectExportSerializer(serializers.ModelSerializer):
+    samples = SampleSerializer(read_only=True, many=True)
     class Meta:
         model = Project
-        fields = ("id", "name", "principal_investigator", "requestor_name", "requestor_email", "status", "targeted_end_date",  "comments" )
+        fields = ("id", "name", "principal_investigator", "requestor_name", "requestor_email", "status", "targeted_end_date",  "comments", "samples")
```
UlysseFG, line 287: Same as above (sample pk only) and i would place the samples before the comments.

backend/fms_core/viewsets/project.py
```diff
@@ -24,6 +26,15 @@ class ProjectViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
         **_project_filterset_fields,
     }
 
+    template_action_list = [
+        {
+            "name": "Link Projects with Samples",
+            "description": "Upload the provided template with up to 384 links between projects and samples.",
```
UlysseFG, line 32: Not sure we have a limit here ... i would remove the "up to 384" from the description.

frontend/src/components/projects/ProjectsDetailedContent.js
```diff
@@ -46,6 +46,11 @@ const ProjectsDetailedContent = ({projects, projectsByID, isFetching, get}) => {
                 <Descriptions.Item label="Targeted End Date" span={2}>{project.targeted_end_date}</Descriptions.Item>
                 <Descriptions.Item label="Comments" span={4}>{project.comments}</Descriptions.Item>
             </Descriptions>
+            <Descriptions bordered={true} size="small" title="Associated Samples" style={{marginTop: "24px"}}>
+                {project.samples && project.samples.map(sample => {
+                  return (<div> <Link to={`/samples/${sample.id}`}> • {sample.name} </Link> </div>);
```
UlysseFG, line 51: We'll need to have something a bit more elaborated than that. Maybe this is just a stub, but i will comment to make sure we do not forget. We will need to include in this section a table with the sample name and sample kind minimally with a way to order and filter them by name and kind. The business plan lists cohort and volume, but i am not sure if this was updated to the latest agreement with the stakeholders... I would add those 2 fields too (cohort and volume) since there may be projects where cohorts are important to organize the samples within and volume maybe be a relevant information for PMs.

frontend/src/components/projects/ProjectsListContent.js
```diff
@@ -121,6 +122,7 @@ const ProjectsListContent = ({
   return <>
     <AppPageHeader title="Projects" extra={[
       <AddButton key='add' url="/projects/add" />,
+      <LinkButton key="link" url="/projects/actions/0" />,
```
UlysseFG, line 125: Maybe use a constant for the URL since this is hard to identify what the action 0 is looking at this. There might also be other actions added down the line and it would help make sure we point to the right one.

jleblond: There is definitely some method to do that automatically somewhere. The back-end viewsets send a list of actions, and there is a front-end 'helper' that translates those actions into buttons/urls

backend/fms_core/serializers.py
jleblond: Agree. If you need to access the Samples information from the Project page, you can fetch the Samples from the redux store with their IDs with a method such as 'list()' or worse to worse with 'withItem'

frontend/src/components/projects/ProjectsListContent.js
jleblond: Lookup actionsToButtonList in the front-end, this is what is used in the front-end for template actions

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+            obj.project = Project.objects.get(name=data["Project Name"])
+        except Project.DoesNotExist:
+            obj.project = None
+            errors["project"] = ValidationError([f"No project with name [{data['Project Name']}]"], code="invalid")
+
+        # Set object for deletion if REMOVE_ACTION
+        try:
+            link_exists = SampleByProject.objects.filter(sample=obj.sample, project=obj.project).exists()
+        except ObjectDoesNotExist:
+            link_exists = False
```
jleblond, line 57: Why not having a simpler logic such as if  queryset.exists(): blabla else: blablabla and re-arranging a bit the logic? I don't think there is a need to define an intermediate variable for this purpose

backend/fms_core/models/sample_by_project.py
```diff
@@ -0,0 +1,37 @@
...
+__all__ = ["SampleByProject"]
+
+@reversion.register()
+class SampleByProject(TrackedModel):
+    sample = models.ForeignKey("Sample", help_text="Sample assigned to a project.",
+                              on_delete=models.CASCADE, related_name="project_association")
+    project = models.ForeignKey("Project", help_text="Project to which the sample is associated.",
+                               on_delete=models.CASCADE, related_name="sample_association")
+    class Meta:
+        constraints = [
```
jleblond, line 21: Interesting solution, I wasn't aware of that in Django :)

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+    def import_obj(self, obj, data, dry_run, **kwargs):
+        errors = {}
+        # Get field info
+        try:
+            super().import_obj(obj, data, dry_run, **kwargs)
+        except ValidationError as e:
+            errors.update(e.update_error_dict(errors).copy())
+        # Get sample object
+        try:
+            obj.sample = Sample.objects.get(name=data["Sample Name"],
```
jleblond, line 40: Are both fields (sample name and container information) necessary in the template? Is it still ok if we just get the Container info?

backend/fms_core/resources/project_link_sample.py
sebastianballesteros: @UlysseFG What do you think?

backend/fms_core/models/sample.py
sebastianballesteros: So on one of them it would have to be just foreign key?

frontend/src/components/projects/ProjectsListContent.js
UlysseFG: I think Sebastian wanted a custom button with a link icon...

backend/fms_core/resources/project_link_sample.py
jleblond: Also, from the documentation I dont think exists() ever throws a ObjectDoesNotExist exception

backend/fms_core/resources/project_link_sample.py
UlysseFG: I would either keep it required or just remove it altogether ... we do not really need it, but making it optional will just make it so no one fills it (making it useless). This is a small layer of validation, if it is too demanding for the lab we could remove it in a future iteration.

frontend/src/components/projects/ProjectsListContent.js
jleblond: Aren't we in ProjectListContent with the same list of buttons as in the other sections?

backend/fms_core/resources/project_link_sample.py
UlysseFG: eheh i think i actually wrote this ... I am not sure this would be simpler ... it could be different but i don't see anything simpler (this is pretty simple). The intermediate variable could be removed but you would need to test the action at 2 different place in your code ... I do not know... up to you Sebastian really.

jleblond: But at least, if this is never going to throw an except, please remove the exception catching :)

backend/fms_core/resources/project_link_sample.py
jleblond: Yeah, I wasnt sure what we decided on required fields. If the two are mandatory that's fine.

backend/fms_core/resources/project_link_sample.py
UlysseFG: oh yeah you can most likely remove the exception handling for the exists()

jleblond: I suggest something along those lines:
image


IMPORTANT: Also, to verify: is there any 'deleted' object handling when making queries? For example, if you query SampleByProject.objects.all() do the objects marked as 'deleted' still appear in the query? if not it should be managed in some of the model Manager or at a general place

frontend/src/components/samples/SamplesAssociatedProjects.js
```diff
@@ -80,56 +88,58 @@ const SamplesAssociatedProjects = ({
   setSortBy,
 }) => {
 
-  const filters = {}
+  useEffect(() => {
+    // returned function will be called on component unmount
+    return () => {
+      clearFilters()
```
jleblond, line 94: Is this behaviour consistent with the other entities list pages?

sebastianballesteros: It is not. However, it is necessary because of the filtering behaviour I implemented in order to re-use some components. I can explain it with details if you want.

jleblond:
I am not so sure about the approach of having all the samples, then automatically filtering them by project name in the project page here: 
image
The 'one filter applied' by default with the project name lets the user think they have control over that filter; while it's a filter by default based on the project page they are on.

Would there be a way to mention that these are the samples associated to Project X, but not count the project name as '1 filter applied'? 

Also, I recommend having that associated samples section in a different tab because for some projects the list could be quite long (i.e. see Experiment detail page for an example)

sebastianballesteros:
I agree that creating an endpoint would be a more complete solution than just filtering the list. If @UlysseFG agrees I can go in that way.

UlysseFG:
> I agree that creating an endpoint would be a more complete solution than just filtering the list. If @UlysseFG agrees I can go in that way.

not sure about the endpoint i need to think it over ... i fear we are opening the door to the multiplication of endpoints (since you could filter on other fields eventually). 

I would prefer a "frontend endpoint" that is redirected by api.js to the correct backend endpoint (regular sample endpoint) with the inserted filter. this would only bloat the frontend since it does not really provide a new functionality to the backend...

backend/fms_core/filters.py
```diff
@@ -0,0 +1,23 @@
+import django_filters
```
UlysseFG, line 1: Not sure i would start implementing filters for each viewset ... If we decide to go that way we should do all of them or at least create only the ones that are required because of the batch filtering (samples, container). From what i see here it is also a departure from the pattern we were using before where filters and lookup_expr are liste in the fields. We should go one way or the other not both unless there is a reason.

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+            obj.sample = None
+            errors["sample"] = ValidationError([f"No sample with name [{data['Sample Name']}] at those container coordinates."], code="invalid")
+        # Get project object
+        try:
+            obj.project = Project.objects.get(name=data["Project Name"])
+        except Project.DoesNotExist:
+            obj.project = None
+            errors["project"] = ValidationError([f"No project with name [{data['Project Name']}]"], code="invalid")
+
+        # Set object for deletion if REMOVE_ACTION
```
UlysseFG, line 53: Not sure this comment belongs here anymore ...

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+        if SampleByProject.objects.filter(sample=obj.sample, project=obj.project).exists():
+            # Perform the desired action
+            if data["Action"] == REMOVE_ACTION:
+                    obj.deleted = True
```
UlysseFG, line 57: too much indentation

backend/fms_core/resources/sample.py
```diff
@@ -118,6 +120,7 @@ def before_import(self, dataset, using_transactions, dry_run, **kwargs):
 
     def import_obj(self, obj, data, dry_run):
         errors = {}
+        self.project = {}
```
UlysseFG, line 123: This is a dangerous initialization ... you should use None. Setting self.project with {} make a dictionary out of it ... self.project is a Field() normally... You are changing the type which may have unintended consequences.

backend/fms_core/resources/sample.py
```diff
@@ -59,6 +59,8 @@ class SampleResource(GenericResource):
 
     volume = Field(attribute='volume', column_name='Volume (uL)', widget=DecimalWidget())
 
+    project = Field(attribute='projects', column_name='Project', widget=ManyToManyWidget(Project))
```
UlysseFG, line 62: If you are manually linking in the project, do not set the attribute and widjet. It will not help you since those are there to automatically create many-to-many entries when possible (not the case here). You may have to manually add project to the confirmation columns (by adding it in the export order).

backend/fms_core/resources/sample.py
```diff
@@ -177,6 +180,12 @@ def import_obj(self, obj, data, dry_run):
             individual = None
             errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
 
+        if data["Project"]:
+            try:
+                self.project = Project.objects.get(name=get_normalized_str(data, "Project"))
+            except Project.DoesNotExist:
+                errors["project "] = ValidationError("Project does not exist.", code="invalid")
```
UlysseFG, line 187: Extra space in errors["project "]

backend/fms_core/resources/sample.py
```diff
@@ -177,6 +180,12 @@ def import_obj(self, obj, data, dry_run):
             individual = None
             errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
 
+        if data["Project"]:
+            try:
+                self.project = Project.objects.get(name=get_normalized_str(data, "Project"))
```
UlysseFG, line 185: self.Project went from a Field() to a Dict() to a Project() ... this is not good... You already create the relationship by creating the samplebyproject entry in the m2m section. You should not set self.project here and instead simply test for the project with exists() and raise and error if it does not exist.

backend/fms_core/resources/sample.py
```diff
@@ -278,6 +287,11 @@ def after_save_instance(self, instance, using_transactions, dry_run):
         super().after_save_instance(instance, using_transactions, dry_run)
         reversion.set_comment("Imported samples from template.")
 
+    def save_m2m(self, obj, data, using_transactions, dry_run):
+        sampleByProject = SampleByProject.objects.create(project=self.project, sample=obj)
```
UlysseFG, line 291: I would encourage you to use another attribute than self.project (self.project_obj maybe ? or change self.project to self.project_name to free up self.project) to store the reference to the project for this ... You don't need to assign it if you do not need sampleByProject.

backend/fms_core/resources/sample.py
```diff
@@ -278,6 +287,11 @@ def after_save_instance(self, instance, using_transactions, dry_run):
...
+        data["Project"] = self.project.id
```
UlysseFG, line 292: This is not useful unless you want import_export to try to create a second copy of your sampleByProject when the super() method is called.

backend/fms_core/serializers.py
```diff
@@ -209,7 +211,7 @@ class Meta:
                   'individual_id', 'sex', 'pedigree', 'mother_name', 'father_name',
                   'current_volume', 'concentration', 'collection_site', 'tissue_source', 'creation_date', 'phenotype',
                   'depleted', 'coordinates',
-                  'comment')
+                  'comment', 'projects')
```
UlysseFG, line 214: i would insert projects before the comment for readability (comments tend to be long).

backend/fms_core/template_paths.py
```diff
@@ -24,3 +25,5 @@
 SAMPLE_SUBMISSION_TEMPLATE = static("submission_templates/Sample_submission_v3_2_0.xlsx")
```
UlysseFG, line 25: This link was not updated to the latest version. Also in the template, i would move the column between Alias and Cohort. This is an important column and it should be toward the start of the line. One more little detail, when you save you xlsx it saves the cursor position. Ideally you'll want to reset it at a more interesting position (A:1 possibly or the first cell of the header) and make sure you are on the template sheet.

backend/fms_core/tests/invalid_templates/Project_link_samples_invalid_project_and_sample.csv
```diff
@@ -0,0 +1,108 @@
...
+,,,,,
+#,Action,Project Name,Sample Name,Sample Container Barcode,Sample Container Coord
+1,Add,TestProject1,InvalidSample,tube001,
+2,Add,InvalidProject,sample2,tube002,
+3,,,,,
+4,,,,,
+5,,,,,
+6,,,,,
+7,,,,,
+8,,,,,
```
UlysseFG, line 16: You can cut all those extra lines. They are not really useful. You can leave one so it tests the case where there is empty lines in the template but the rest are not of any use.

backend/fms_core/tests/test_resources.py
```diff
@@ -595,3 +599,57 @@ def test_invalid_container_rename(self):
...
+        sample2 = Sample.objects.get(name="sample2", container__barcode="tube002")
+
+        # Test basic import success
+        self.assertEqual(len(SampleByProject.objects.all()), 2)
+
+        # Test parent record auto-generation
+        self.assertTrue(SampleByProject.objects.filter(sample=sample1, project=self.project1).exists())
+        self.assertTrue(SampleByProject.objects.filter(sample=sample2, project=self.project2).exists())
+
+    def unlink_projects_to_samples(self):
```
UlysseFG, line 629: Tests methods should start with "test_" .  The real test seem to have been commented out below ... Any reason ?

backend/fms_core/tests/valid_templates/Project_link_samples_v3_4_0_B_A_1.csv
```diff
@@ -0,0 +1,1001 @@
...
+Instructions:,,,,,
+,Each line define the addition or removal of a sample instance from a Project. Each sample instance will ,,,,
+,inherit from its parent’s association when extracted or transfered. A sample instance can be associated,,,,
+,to more than one project.,,,,
+,,,,,
+#,Action,Project Name,Sample Name,Sample Container Barcode,Sample Container Coord
+1,Add,TestProject1,sample1,tube001,
+2,Add,TestProject2,sample2,tube002,
+3,,,,,
+4,,,,,
```
UlysseFG, line 12: 1000 lines of csv ... is this a stress test? :)

backend/fms_core/tests/valid_templates/Project_link_samples_v3_4_0_B_B_1.csv
```diff
@@ -0,0 +1,108 @@
...
+Instructions:,,,,,
+,Each line define the addition or removal of a sample instance from a Project. Each sample instance will ,,,,
+,inherit from its parent’s association when extracted or transfered. A sample instance can be associated,,,,
+,to more than one project.,,,,
+,,,,,
+#,Action,Project Name,Sample Name,Sample Container Barcode,Sample Container Coord
+1,Remove,TestProject1,sample1,tube001,
+2,,,,,
+3,,,,,
+4,,,,,
```
UlysseFG, line 12: Too long again ... cut some lines

frontend/src/components/LinkButton.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
```
UlysseFG, line 1: Is this used anywhere ? You can remove it if it is not used anymore ...

backend/fms_core/filters.py
jleblond: Right, if an entity filter doesnt require a custom filter, I also would not implement a custom filter for that one.

backend/fms_core/resources/sample.py
jleblond: Agreed. In your Sample resource, the field defined as being 'project = Field(attribute='projects', column_name='Project', widget=ManyToManyWidget(Project))' can actually be referenced with self.project since it is a class attribute.

Re-initiating self.project = {} in import_obj can be dangerous here. You could simply rename to variable to something like self.project_obj for example to make a clear distinction

backend/fms_core/resources/sample.py
jleblond: It could also help the user to have a more precise error such as "Project named {name} does not exist."

backend/fms_core/resources/sample.py
jleblond: Agreed. Also, btw if you want to use it for an object, you can initiate it by default with variable_name = None. if it is meant to represent a dictionary you can initiate it with variable_name = {}

backend/fms_core/tests/test_resources.py
jleblond: Not sure that this method 'unlink_projects_to_samples' is ever called (it is in your commented block).

If it is a test case by itself, indeed it should start with "test_"
If it is a helper method (such as in load_samples, load_containers, etc) it should indeed not start with "test_"

So from what I understand from your code, you just forgot to uncomment your test_unlink_proejcts_to_sample function :)

backend/fms_core/tests/valid_templates/Project_link_samples_v3_4_0_B_A_1.csv
jleblond: Right, I assume you probably generated your csv automatically from an xlsx file; so you can remove the extra lines at the end of your csv

backend/fms_core/resources/project_link_sample.py
jleblond: From my understanding here, you want to set the deleted attribute of that ProjectLinkSample to True. I wonder if there exists any sorts of filters which would prevent attributes with 'deleted == True' to show up in search and in lists?

jleblond: EDIT: While reading the code down there, now I understand that it is a marker that you put on your resource object to know if it is to be deleted or not later in the code. For it to be more explicit, I suggest you rename 'deleted' by 'to_be_deleted' to indicate it's a marker.

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+        elif field.column_name == "Sample Container Coord":
+            data["Sample Container Coord"] = get_normalized_str(data, "Sample Container Coord").upper()
+        super().import_field(field, obj, data, is_m2m)
+
+    def after_save_instance(self, instance, using_transactions, dry_run):
+        super().after_save_instance(instance, using_transactions, dry_run)
+        # If the instance is deleted (saved version), delete the instance
+        instance.refresh_from_db()
+        if instance.deleted:
+            super().instance.delete() # Delete without creating a new deleted version
```
jleblond, line 90: This will be calling our custom delete() method in tracked_model if your entity model inherits from TrackedModel.
I wonder if there is any special handling of the 'deleted' objects; this is a custom attribute, so Django for now does not what to do with those objects.

i.e: you call: Container.objects..get(id=1).delete() it will call the delete() custom method in our tracked_model.py file. The attribute 'deleted' in our Container row will be set to 'true'
However, if you lookup Container.objects.all(), unless the 'deleted' objects are handled, the Container id 1 will still show up in your list of Container objects. 
Did we think/implemented a mechanism for that?

backend/fms_core/resources/project_link_sample.py
jleblond: I would rename the obj."deleted" to something that indicates that it is only a marker for future handling (down in the code). Maybe something like "to_be_deleted" would emphasis that it is not the resource own deletion status, but a marker?

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+class ProjectLinkSampleResource(GenericResource):
+    action = Field(column_name='Action')
+    project_name = Field(column_name='Project Name') 
+    sample_name = Field(column_name='Sample Name')
+    barcode = Field(column_name='Sample Container Barcode',
+                     widget=ForeignKeyWidget(Container, 'barcode'))
+    coordinates = Field(column_name='Sample Container Coord')
+
+    class Meta:
+        model = SampleByProject
```
jleblond, line 25: Don't forget the import_id_fields and fields such as in: 
        import_id_fields = ()
        fields = (
        )

Otherwise you get a bunch of internal fields that are not relevant to the user:
image

jleblond:
I suggest that the 'associated projects' section goes to a new tab, so the Version information (which is relevant to the lab) is more visible, and also for consistency 

image

frontend/src/components/projects/ProjectsAssociatedSamples.js
```diff
@@ -0,0 +1,155 @@
...
+        <Link to={`/samples/${sample.id}`}>
+          <div>{name}</div>
+          {sample.alias &&
+            <div><small>alias: {sample.alias}</small></div>
+          }
+        </Link>,
+    },
+    {
+      title: "Cohort",
+      dataIndex: "cohort",
```
jleblond, line 45: 'Cohort' attribute is on an Individual, not a Sample. Therefore, it will be listed as empty even if your sample is associated to an individual which does have a cohort.

i.e:
image

image

image

frontend/src/components/projects/ProjectsDetailedContent.js
```diff
@@ -33,6 +44,8 @@ const ProjectsDetailedContent = ({projects, projectsByID, isFetching, get}) => {
     const title =
         `Project ${project.name}`;
 
+    let  isFetchingSamples = samples.every(sample =>  sample ? sample.isFetching : true)
```
jleblond, line 47: Unused variable?

Also lint (1 space after 'let', not 2)

frontend/src/components/projects/ProjectsDetailedContent.js
```diff
@@ -2,29 +2,40 @@ import React from "react";
...
 const mapDispatchToProps = dispatch =>
     bindActionCreators({ get }, dispatch);
 
-const ProjectsDetailedContent = ({projects, projectsByID, isFetching, get}) => {
+const ProjectsDetailedContent = ({projects, projectsByID, samplesByID, isFetching, get}) => {
     const history = useHistory();
     const {id} = useParams();
     const isLoaded = id in projectsByID;
     const project = projectsByID[id] || {};
+    const isSamplesEmpty = project.samples && project.samples.length === 0;
```
jleblond, line 30: Unused variable? 

Trick: I think you are using Pycharm or any JetBrains software right? If there are unused variables they will be grey instead of coloured 

Ex: 
image

(same applies for unused packages) ex:
image

frontend/src/utils/serializeFilterParams.js
```diff
@@ -44,10 +43,13 @@ export default function serializeFilterParams(filters, descriptions) {
             key += "__recursive"
           else if (options.exactMatch)
             key += "__startswith"
+          else if (options.many_to_many)
```
UlysseFG, line 46: It is not really clear to me what is that many_to_many option. Does this represent a real option (we can actually set it to false ?) or is it simply a descriptor for the filter. Here both are used but i doubt from what i have seen that you actually have a choice (could possibly only use descriptor.many_to_many). I am also uncertain what it differentiate between. Is seem to be a way to say exact_match when there is optional filtering but nothing when there is no options ... Lastly it seem very specific to the project and sample case ... can't we use the preexisting options and descriptors to cover that case ? or maybe rename it so it's clear what it means ?

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,104 @@
...
+            data["Sample Container Barcode"] = get_normalized_str(data, "Sample Container Barcode")
+        elif field.column_name == "Sample Container Coord":
+            data["Sample Container Coord"] = get_normalized_str(data, "Sample Container Coord").upper()
+        super().import_field(field, obj, data, is_m2m)
+
+    def after_save_instance(self, instance, using_transactions, dry_run):
+        super().after_save_instance(instance, using_transactions, dry_run)
+        # If the instance is deleted (saved version), delete the instance
+        instance.refresh_from_db()
+        if instance.deleted:
```
UlysseFG, line 89: We could explain in a comment here why we actually skip model.delete() and that we manually set the deleted flag instead.  This has to do with preventing the creation of 2 saved versions before we delete the object (having import_export delete the object instead of saving it more complicated than this).

frontend/src/utils/serializeFilterParams.js
jleblond: Same. I dont understand the 'many_to_many' as an option AND as a description 
(it can be both? )
Ref: 
image

frontend/src/components/projects/ProjectsDetailedContent.js
```diff
@@ -47,6 +60,8 @@ const ProjectsDetailedContent = ({projects, projectsByID, isFetching, get}) => {
                 <Descriptions.Item label="Comments" span={4}>{project.comments}</Descriptions.Item>
             </Descriptions>
             <TrackingFieldsContent entity={project}/>
+            <Title level={4} style={{marginTop: '2rem'}}> Associated Samples </Title>
+            <ProjectsAssociatedSamples projectName={project.name} />
```
UlysseFG, line 64: Might be better to pass IDs around since all the entities are stored according to their ids.

frontend/src/components/projects/ProjectsDetailedContent.js
```diff
@@ -2,29 +2,40 @@ import React from "react";
...
-    projectsByID: state.projects.itemsByID
+    projectsByID: state.projects.itemsByID,
+    samplesByID: state.samples.itemsByID,
 });
 
 const mapDispatchToProps = dispatch =>
     bindActionCreators({ get }, dispatch);
 
-const ProjectsDetailedContent = ({projects, projectsByID, isFetching, get}) => {
+const ProjectsDetailedContent = ({projects, projectsByID, samplesByID, isFetching, get}) => {
```
UlysseFG, line 25: Not sure if projects is used and useful here.

frontend/src/utils/filterProjectsBySample.js
```diff
@@ -0,0 +1,13 @@
+export default function filterProjectsBySample(setFilter, filters, sampleName, clearFilters) {
```
UlysseFG, line 1: I really don't like these two functions they are really redundant and do next to nothing ... At the very least if you cannot find a more generic solution to the issue try to make this patch more generic ... make one function that is more generic to cover present and future needs. I would suggest  setDefaultFilter(filterName, filterValue, setFilter, filters, clearFilters). This will prevent at least the multiplication of those copy-pasted functions.

frontend/src/components/projects/ProjectsAssociatedSamples.js
UlysseFG: dataIndex would be individual__cohort ?

UlysseFG:
You seem to be missing this in the current migration
migrations.AddField(
            model_name='project',
            name='samples',
            field=models.ManyToManyField(blank=True, related_name='projects', through='fms_core.SampleByProject', to='fms_core.Sample'),
        ),

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,127 @@
...
+            return True
+        else:
+            return False
+
+    def get_or_init_instance(self, instance_loader, row):
+        action = row["Action"].upper()
+        if action == REMOVE_ACTION:
+            try:
+                sample = Sample.objects.get(name=row["Sample Name"],
+                                                container__barcode=row["Sample Container Barcode"],
```
UlysseFG, line 102: indentation here

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,127 @@
...
+        try:
+            obj.project = Project.objects.get(name=data["Project Name"])
+        except Project.DoesNotExist:
+            obj.project = None
+            errors["project"] = ValidationError([f"No project with name [{data['Project Name']}]"], code="invalid")
+
+        # Ensure link exists
+        if SampleByProject.objects.filter(sample=obj.sample, project=obj.project).exists():
+            # Remove action is valid if the link already exists
+            if data["Action"] == REMOVE_ACTION:
```
UlysseFG, line 59: Not sure this flag setting is useful anymore ... This should now be done more "transparently" through the delete().

backend/fms_core/serializers.py
UlysseFG: samples are still after comments

backend/fms_core/tests/test_resources.py
```diff
@@ -48,9 +49,11 @@ def get_ds():
 CONTAINER_RENAME_CSV = APP_DATA_ROOT / "Container_rename_v3_2_0_B_A_1.csv"
 SAMPLE_EXTRACTION_CSV = APP_DATA_ROOT / "Sample_extraction_v3_3_0_B_A_1.csv"
 SAMPLE_TRANSFER_CSV = APP_DATA_ROOT / "Sample_transfer_v3_2_0_B_A_1.csv"
-SAMPLE_SUBMISSION_CSV = APP_DATA_ROOT / "Sample_submission_v3_2_0_B_A_2.csv"
+SAMPLE_SUBMISSION_CSV = APP_DATA_ROOT / "Sample_submission_v3_4_0_B_A_2.csv"
```
UlysseFG, line 52: This is unlikely to be the second iteration of the sample submission csv for this version since its not out yet ;). Sample_submission_v3_4_0_B_A_2 => Sample_submission_v3_4_0_B_A_1
We increment the counter within each version when we change a test template and the version do not change. We reset the counter when we change the version even if the test template change.

frontend/src/utils/serializeFilterParams.js
```diff
@@ -47,7 +48,7 @@ export default function serializeFilterParams(filters, descriptions) {
           else
             key += "__icontains"
         } else {
-          key += "__icontains"
+            key += "__icontains"
```
UlysseFG, line 51: I think the indentation was actually correct before. Front end has 2 space indentation, back end 4 spaces.

sebastianballesteros: Shouldn't line 51 should be indented since it is inside an else statement?

backend/fms_core/resources/project_link_sample.py
sebastianballesteros: are yo referring to the obj.deleted flag?

UlysseFG: yeah

UlysseFG:
Found a small issue with sample listing (at the sample and project level). Samples with 2 projects get listed twice ... (possibly once by project association ...

backend/fms_core/resources/project_link_sample.py
```diff
@@ -0,0 +1,131 @@
...
+            if SampleByProject.objects.filter(sample=obj.sample, project=obj.project).exists() and data["Action"] == ADD_ACTION:
+                errors["link"] = ValidationError(
+                    [f"Sample [{data['Sample Name']}] is already associated to project [{data['Project Name']}]."],
+                    code="invalid")
+
+            # If the link doesn't exists we can't perform a remove action
+            else:
+                if data["Action"] == REMOVE_ACTION:
+                    errors["link"] = ValidationError([f"Sample [{data['Sample Name']}] is not currently associated to project [{data['Project Name']}]."],
+                                                     code="invalid")
```
jleblond, line 67: I found this conditional statement rather difficult to read imo. 

In the case where the association already exists, but the action to be performed is REMOVE_ACTION, the if should evaluate to False (True and False => False), we then fall in the else: and there data['Action'] is not REMOVE, so nothing happens but we have to read carefully each conditional branch.

backend/fms_core/resources/sample.py
```diff
@@ -177,6 +181,13 @@ def import_obj(self, obj, data, dry_run):
             individual = None
             errors["individual"] = ValidationError(e.messages.pop(), code="invalid")
 
+        if data["Project"]:
+            project_name = get_normalized_str(data, "Project")
+            if not Project.objects.filter(name=project_name).exists():
```
jleblond, line 186: Another comment about readability :P Do we need to do a 'not' ? why dont you do 
"if Project.objects.filter(name=project_name).exists():
     self.project_object ... 
 else:
    errors["project'] = ValidationError .... "

jleblond: EDIT: I re-read the lines, and I rather suggest:

try:
   self.project = Project.objects.get(..... )
except ObjectDoesNotExist:
   errors["project'] = ...  
   
More straightforward to me (+ possibly save 1 db hit :P )

sebastianballesteros: Good catch!!

backend/fms_core/resources/project_link_sample.py
sebastianballesteros: I suggest maybe changing the else to an elif data["Action"] == REMOVE_ACTION?

frontend/src/components/samples/SamplesListContent.js
```diff
@@ -22,7 +22,15 @@ import FiltersWarning from "../filters/FiltersWarning";
...
 import mergedListQueryParams from "../../utils/mergedListQueryParams";
 
-const getTableColumns = (containersByID, individualsByID, sampleKinds) => [
+const getProjectLinks = (ids, projectsByID) => {
+  return (
+    ids.map(id => {
+      return (<div> <Link to={`/projects/${projectsByID[id].id}`}> {projectsByID[id].name} </Link> </div>);
+    })
+  )
+}
```
jleblond, line 31: No need for a distinct function for that. This is used only to render the content of a cell in the list. The other cells are rendered directly in the render function of the appropriate getTableColumns column

backend/fms_core/resources/project_link_sample.py
jleblond: Yes, something like: 
if data action == ADD and SampleByProject filter blablab exists   
     do .... 
elif data action == REMOVE 
   do ...

frontend/src/utils/serializeFilterParams.js
UlysseFG: yeah but the else should be aligned where the "}" is.

backend/fms_core/tests/test_resources.py
```diff
@@ -595,3 +599,57 @@ def test_invalid_container_rename(self):
...
+        self.assertFalse(SampleByProject.objects.filter(sample=sample1, project=self.project1).exists())
+        self.assertTrue(SampleByProject.objects.filter(sample=sample2, project=self.project2).exists())
+
+    def test_invalid_link_project_sample(self):
+        self.load_samples()
+        self.create_projects()
+
+        with self.assertRaises(ValidationError), open(TEST_DATA_ROOT / "Project_link_samples_invalid_project_and_sample.csv") as sf:
+            s = Dataset().load(sf.read())
+            self.pr.import_data(s, raise_errors=True)
```
jleblond, line 655: Does this test work for you? On my end it fails with error: django.core.exceptions.ValidationError: {'sample': ['No sample with name [InvalidSample] at those container coordinates.']}

UlysseFG:
> Found a small issue with sample listing (at the sample and project level). Samples with 2 projects get listed twice ... (possibly once by project association ...

Looks like the distinct() you added did do the job i cannot reproduce the multiple sample issue anymore...

backend/fms_core/tests/test_resources.py
sebastianballesteros: pull ;)

frontend/src/components/projects/ProjectsAssociatedSamples.js
```diff
@@ -0,0 +1,168 @@
...
+    filters,
+    setFilter,
+    setFilterOption,
+  )))
+
+  const nFilters = getNFilters(filters)
+  const nFiltersForWarning = nFilters - 1
+
+  //To ensure user doesn't see the full list of samples
+  samplesByID = samples.length === totalCount ? samplesByID : {}
```
UlysseFG, line 133: As far as i can see this flushes the whole local cache of samples ... this may be overkill ... When i browse the site if i jump from samples to container and other tabs the sample show up instantly when i come back ... when i load a project list of samples, if i go back to the samples list i get a spinner and wait for the samples to load. I think clearing the samples (items) would be less damaging to the navigation if it makes any sense.

frontend/src/utils/serializeFilterParams.js
```diff
@@ -8,6 +8,8 @@ export default function serializeFilterParams(filters, descriptions) {
   const params = {}
 
   Object.keys(filters).forEach(field => {
+    console.log(descriptions)
+    console.log(filters)
```
UlysseFG, line 12: Couple of lines that are likely forgotten testing bits...

frontend/src/utils/serializeFilterParams.js
UlysseFG: Still too much indentation there...

frontend/src/components/projects/ProjectEditContent.js
```diff
@@ -9,7 +9,6 @@ import {
   Form,
   Input,
   Select,
-  Switch,
```
UlysseFG, line 12: This is needed

PR #133, opened 2021-08-25
Renamed the options for status of a project in frontend/backend 

sebastianballesteros:
Modify them to Open/Closed


frontend/src/components/projects/ProjectEditContent.js
```diff
@@ -169,11 +166,14 @@ function deserialize(values) {
 function serialize(values) {
   const newValues = {...values}
 
+  if (newValues.status === false) newValues.status = "Closed"
```
UlysseFG, line 169: I think we should stick to putting code on a new line after an if or an else ...

frontend/src/components/projects/ProjectEditContent.js
```diff
@@ -116,11 +116,8 @@ const ProjectEditContent = ({token, projectsByID, add, update, listTable, summar
           <Form.Item label="Requestor Email" {...props("requestor_email")} rules={emailRules} >
             <Input />
           </Form.Item>
-          <Form.Item label="Status" {...props("status")} rules={requiredRules}>
-            <Select>
-              <Option key="Ongoing" value="Ongoing" > Ongoing </Option>
-              <Option key="Completed" value="Completed" > Completed </Option>
-            </Select>
+          <Form.Item label="Status" {...props("status")} >
+            <Switch style={{width: 80}} checkedChildren="Open" unCheckedChildren="Closed" defaultChecked/>
```
jleblond, line 120: Was the toggle button part of the design requirements? If not, I would use a different button (select) because the status of a project is a state, and not a boolean. But that's a personal preference... Also if we're sure to have only open/closed choices for the status of a project, we could even use a boolean instead. 

What do you guys think? @UlysseFG @sebastianballesteros

UlysseFG: Not sure about the boolean ... it might paint us in a corner (a little bit) when/if people will decide to add more status. The same applies for the toggle option while very nice for 2 options will need to be change if we move to a many stat. I would not change it now since it is easier for the user, but it's up to you Sebastian it's not worth changing back that and the tests to me.

jleblond: Right, that's a status; so it should stay as a string choice in the Project model. 

I think the 'select' option would have been better to represent a status (and that is what it was before this PR) with default value selected 'Open' (I doubt they will often have to change it to 'Closed' when creating a new project from this form), but that's ok to me if we dont want to change it for the moment.

sebastianballesteros: I agree that it should have stayed as I originally implemented it. However, Andres asked it as a requirement, maybe we can discuss it in today's meeting?

jleblond: That's ok if it is a requirement agreed on :)

jleblond (approved):
Looks good to me

UlysseFG (approved):
You can merge when you are ready.

backend/fms_core/models/project.py
```diff
@@ -12,7 +12,7 @@
 
 __all__ = ["Project"]
 
-STATUS_CHOICES = ["Ongoing", "Completed"]
+STATUS_CHOICES = ["Open", "Closed"]
```
UlysseFG, line 15: Little note here while i think about it. Maybe we should move this to _constants.py and rename it PROJECT_STATUS_CHOICES.

PR #134, opened 2021-09-02
Correction to the infinium beadchip format.

UlysseFG:


UlysseFG:
The cypress tests still fail. I will investigate further.

jleblond:
2 Container creation fixtures (for front-end testing) template files were deleted?

Otherwise looks good to me. Do your cypress tests still fail?

UlysseFG:
> 2 Container creation fixtures (for front-end testing) template files were deleted?
Nope only one the number 2 ... the other is just modified...

jleblond:
I think it's a very great start for the front-end tests, I dont see those 'ResizableObserver' errors anymore. From the tests I ran, there are still some tests that are invalid though, maybe that could be done in a separate PR, but I think some lines of tests are  no longer valid. Let me know if it's the same for you, or if all the tests do pass.

PR #136, opened 2021-09-16
Template importer

jleblond:
Continuation of https://github.com/c3g/freezeman/pull/123 with fixed back-end tests/packages and new tool name 'Template importer'

TODO:
- Way to ensure that the row number ingested does correspond to the header of the template


backend/fms_core/services/container.py
```diff
@@ -0,0 +1,50 @@
...
+from django.core.exceptions import ValidationError
+from fms_core.models import Container
+
+def get_container(barcode=None):
+    container = None
+    errors = []
+    warnings = []
+
+    try:
+        container = Container.objects.get(barcode=barcode)
```
UlysseFG, line 11: We should differentiate the error message when a barcode is missing (None) than when there is no DB entry for a barcode.

backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -0,0 +1,108 @@
...
+                         'result_previews': self.previews_info,
+                         }
+        return import_result
+
+
+    def create_sheet_data(self, sheet_name, header_row_nb):
+        try:
+            pd_sheet = pd.read_excel(self.file, sheet_name=sheet_name)
+            # Convert blank and NaN cells to None and Store it in self.sheets
+            dataframe = pd_sheet.applymap(lambda x: blank_and_nan_to_none(x))
```
UlysseFG, line 75: pretty confident that this is the same as dataframe = pd_sheet.applymap(blank_and_nan_to_none). Not sure i have ever found a good reason to use lambda in python ... (or anywhere other than lisp ;) this is like wrapping your function in a function ;)

backend/fms_core/services/container.py
UlysseFG: This is not just true here ... It could go a long way in identifying the issues. In some case, barcode may be optional but in those case this function should not be called at all. If it is called, we want to know.

jleblond: I suggest renaming this function get_container_from_barcode so it makes it more obvious that the Barcode value is required for this function

backend/fms_core/services/container.py
```diff
@@ -0,0 +1,50 @@
...
+
+def get_or_create_container(barcode=None, kind=None, name=None, coordinates=None,
+                            container_parent=None,
+                            creation_comment=None):
+    container = None
+    errors = []
+    warnings = []
+
+    container_data = dict(
+        **(dict(location=container_parent) if container_parent else dict(location__isnull=True)),
```
UlysseFG, line 26: Do we really need to build a dict in such a convoluted way ?
you could do 
container_data = {} and then add each key using an if (eg : if barcode : container_data["barcode"]=barcode)
or 
keys = ["barcode", "name", "coordinates"]
values = [None, "Patate", "C01"]
dict(itertools.compress(zip(key, values), values))

jleblond: More easily readable imo (you have a direct correspondance object attribute = argument)

backend/fms_core/services/container.py
jleblond: I really wanted to make it a convention in Services to have defaults for all the params passed to those functions. Or we could have no default for all of them, but then there is some refactor to do here and there.

I don't really understand what you dont like here? If you try to find a Container without barcode, then it still doesnt exist

backend/fms_core/template_importer/importers/sample_submission.py
```diff
@@ -0,0 +1,78 @@
...
+                'alias': row_data['Alias'],
+                'experimental_group': row_data['Experimental Group'],
+                'concentration': row_data['Conc. (ng/uL)'],
+                'volume': row_data['Volume (uL)'],
+                'collection_site': row_data['Collection Site'],
+                'tissue_source': row_data['Tissue Source'],
+                'creation_date': row_data['Reception Date'],
+                'phenotype': row_data['Phenotype'],
+                'comment': row_data['Comment'],
+                'coordinates': row_data['Location Coord'],
```
UlysseFG, line 56: Location Coord can be used as the location of the sample on a plate or as the location of the tube in a rack ... Here we use it just for the position of the sample. It is a bad design in the current template and we could change that in the new template (adding independant columns for each : sample coordinates, container coordinates (and remove location that is vague as can be)).

backend/fms_core/services/container.py
UlysseFG: well ok if you like it :)

backend/fms_core/services/container.py
jleblond: Right I'll add make it a conditional statement that appends an error if the barcode value is None.

I understand your point regarding having a default value that is not a valid case, but at the same time I don't want to start to manage all the params default values in the code, so I'll leave it as it is for now, with the suggestion we agreed on.

backend/fms_core/template_importer/sheet_data.py
```diff
@@ -0,0 +1,66 @@
...
+            row_data = self.dataframe.iloc[row_id]
+            self.rows.append(row_data)
+
+            print('row_data', row_data)
+
+            row_repr = f"#{row_id + 2}"
+
+            result = {
+                'row_repr': row_repr,
+                'diff': [row_repr] + panda_values_to_str_list(row_data),
```
UlysseFG, line 43: Maybe we could return the info as value before and after and do the diff and the repr stuff on the front-end side ?

jleblond: There is no 'before' and 'after' anymore. It is simply a copy of what was entered by the user in the Excel spreadsheet. I could rename 'diff' to a more appropriate name to the current context, such as 'cell_values' if preferred

UlysseFG: Then we need to add one ... simply printing the new value is a move in the wrong direction we are removing information users have at the moment. Users want to be able to tell from the confirmation table that the change they are submitting is correct. Simply posting back the template content will not help them do that.

UlysseFG: Maybe we can use the information stored in the front-end by using the ID of the object and compare them while rendering the report to prevent hitting the database anymore than needed. 
Some objects may not need before image also since they are always created and never updated.

jleblond: We have to verify at what extent the users use that. From what I recall this feature was half working with import-export and wasn't working at all for cases where the objects didnt follow closely the resource object.

This would require substantial work, and was not part of the technical requirements I initially listed.

jleblond: > Maybe we can use the information stored in the front-end by using the ID of the object and compare them while rendering the report to prevent hitting the database anymore than needed. Some objects may not need before image also since they are always created and never updated.

I don't agree with that solution. The diff, if any, would have to be done entirely in the back-end.

PS: The document 'Workflow Import Tool Freezeman' in 'Technical Analysis' in the FreezeMan shared drive is what I based my design on. If we are to state that this is an essential feature/a blocker, more work will have to be done.

UlysseFG: Without a comparison with the values that were present before the table is a useless copy of the template content. This would be sufficient for object creations (container, sample, individual, ...) but would be lacking for action like update and transfer. I agree that it is significant work ... but we spent significant time doing it once i think it warrant the need to do it again. Maybe we can reuse what is there now until we find the time ...

jleblond: To be discussed in team: at what point this is used by the lab. 
from my understanding it is a 'nice to have' but definitely not a blocker.

UlysseFG: Yeah lets discuss this in team...

UlysseFG:
A few things that emerged from testing :
- Missing Pandas in the requirements.txt file
- Dates in the confirmation tables need some formatting
- Error message need work still (might just be adding more detail in some places but we essentially have the same issues as before  ( i create a sample with an individual with the same ID and different sex and i get a error raised with a message that complain about an individual with an existing name (which should just issue a warning if it was really the issue). I try to create a sample and the parent container cannot be created (coordinates not processed correctly) i get an error that the container name and kind exist and do not match ... The unrelated error is just a cascading error related to the real issue, we will need at some point to better address it. It does not need to be addressed right now since it is an existing issue, but there is no real progress on this side.

backend/fms_core/template_importer/importers/sample_update.py
```diff
@@ -0,0 +1,57 @@
+from fms_core.models import Process, Protocol
+
+from ._generic import GenericImporter
+from fms_core.template_importer.row_handlers.sample_update import SampleRowHandler
+
+class SampleSubmissionImporter(GenericImporter):
+    SHEETS_INFO = [
+        {'name': 'SampleUpdate', 'header_row_nb': 4},
```
UlysseFG, line 8: This will cause troubles eventually people sometime remove lines or add lines. Ideally we would Check for a marker in the file instead. I realized for exemple that if i have one less line in a one line template it will tell me that my operation is successful (because there is supposedly no entry ... This is false and could lead to missing data in the system.
The most obvious marker is the list of headers. If it does match it is the header ... We could also use another marker but it risk being removed edited.

backend/fms_core/template_importer/importers/extraction.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self):
+        super().__init__()
+        # Preload objects accessible to the whole template (not only by row)
+        self.preload_data_from_template()
+
+
+    def preload_data_from_template(self):
+        self.preloaded_data = {'process': None, 'sample_kinds': {}}
+
+        self.preloaded_data['process'] = Process.objects.create(protocol=Protocol.objects.get(name="Extraction"),
```
UlysseFG, line 21: This would not qualify as loading ... I do not oppose bunching up preloading and global creation together but we might want to rename the function... initialize_data_for_template() ?

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
```diff
@@ -0,0 +1,76 @@
...
+        super().__init__()
+
+
+    def process_row_inner(self, source_sample, resulting_sample, process_measurement):
+        original_sample, self.errors['sample'], self.warnings['sample'] = get_sample_from_container(
+            barcode=source_sample['container']['barcode'],
+            coordinates=source_sample['coordinates'])
+
+        if original_sample:
+            original_sample.depleted = True
```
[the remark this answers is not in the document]
jleblond: this was corrected in the tests

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+        if original_sample and destination_container:
+            new_sample, self.errors['extracted_sample'], self.warnings['extracted_sample'] = create_sample(
+                coordinates=resulting_sample['coordinates'],
+                volume=resulting_sample['volume'],
+                concentration=resulting_sample['concentration'],
+                creation_date=resulting_sample['creation_date'],
+                sample_kind=resulting_sample['kind'],
+
```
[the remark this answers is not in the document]
jleblond: i dont understand the comment?

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
```diff
@@ -0,0 +1,76 @@
...
+                name=original_sample.name,
+                collection_site=original_sample.collection_site,
+                alias=original_sample.alias,
+                phenotype=original_sample.phenotype,
+                experimental_group=original_sample.experimental_group,
+                individual=original_sample.individual,
+            )
+
+            if new_sample:
+                process_measurement, self.errors['process_measurement'], self.warnings['process_measurement'] = \
```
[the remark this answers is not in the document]
jleblond: right didnt notice it

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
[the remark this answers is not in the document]
jleblond: for readability, but i cant delete them if preferred

backend/fms_core/template_importer/importers/sample_update.py
```diff
@@ -0,0 +1,57 @@
...
+            }
+            sample_updated = {
+                'volume': row_data['New Volume (uL)'],
+                'concentration': row_data['New Conc. (ng/uL)'],
+                'depleted': row_data['Depleted'],
+            }
+
+            process_measurement = {
+                'process': self.preloaded_data['process'],
+                'volume_used': row_data['Delta Volume (uL)'],
```
UlysseFG, line 39: Delta Volume is an alternative to the new Volume update. The volume used is either the  `- delta volume` or `old_volume - new_volume`.

backend/fms_core/services/container.py
```diff
@@ -0,0 +1,157 @@
...
+        container_to_update.name = name
+    if update_comment:
+        container_to_update.update_comment = update_comment
+
+    try:
+        container_to_update.save()
+    except Exception as e:
+        errors.append(str(e))
+
+    return (container_to_update, errors, warnings)
```
UlysseFG, line 125: Do we need to return the container ... this appear a useless return value. Unless you create a copy of the object you are basically giving back the parameter reference... Same applies to move

PR #138, opened 2021-09-21
Implemented automatic project heritage when performing transfer, extr…

sebastianballesteros:
Please note that it is based on the assign-sample-to-project branch.

### DIFF
--- a/backend/fms_core/resources/extraction.py
+++ b/backend/fms_core/resources/extraction.py
@@ -213,7 +213,6 @@ class ExtractionResource(GenericResource):
 
         # automatic project inheritance
         for project in self.extracted_from.projects.all():
-            print(project)
             SampleByProject.objects.create(project=project, sample=obj)
 
         super().save_m2m(obj, data, using_transactions, dry_run)
--- a/backend/fms_core/resources/extraction.py
+++ b/backend/fms_core/resources/extraction.py
@@ -10,7 +10,7 @@ from ..containers import (
     CONTAINER_SPEC_TUBE,
     CONTAINER_SPEC_TUBE_RACK_8X12,
 )
-from ..models import Container, Process, ProcessMeasurement, Protocol, Sample, SampleKind, SampleLineage
+from ..models import Container, Process, ProcessMeasurement, Protocol, Sample, SampleKind, SampleLineage, SampleByProject
 from ..utils import (
     blank_str_to_none,
     check_truth_like,
@@ -210,6 +210,12 @@ class ExtractionResource(GenericResource):
 
     def save_m2m(self, obj, data, using_transactions, dry_run):
         lineage = SampleLineage.objects.create(parent=self.extracted_from, child=obj, process_measurement=self.process_measurement)
+
+        # automatic project inheritance
+        for project in self.extracted_from.projects.all():
+            print(project)
+            SampleByProject.objects.create(project=project, sample=obj)
+
         super().save_m2m(obj, data, using_transactions, dry_run)
 
     def import_data(self, dataset, dry_run=False, raise_errors=False, use_transactions=None, collect_failed_rows=False, **kwargs):
--- a/backend/fms_core/resources/transfer.py
+++ b/backend/fms_core/resources/transfer.py
@@ -6,7 +6,7 @@ from import_export.fields import Field
 from import_export.widgets import DateWidget, DecimalWidget, JSONWidget, ForeignKeyWidget, ManyToManyWidget
 from ._generic import GenericResource
 from ._utils import skip_rows, add_columns_to_preview
-from ..models import Container, Process, ProcessMeasurement, Protocol, Sample, SampleKind, SampleLineage
+from ..models import Container, Process, ProcessMeasurement, Protocol, Sample, SampleKind, SampleLineage, SampleByProject
 from ..utils import (
     blank_str_to_none,
     check_truth_like,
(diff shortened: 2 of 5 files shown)


backend/fms_core/resources/extraction.py
```diff
@@ -210,6 +210,12 @@ def after_save_instance(self, instance, using_transactions, dry_run):
 
     def save_m2m(self, obj, data, using_transactions, dry_run):
         lineage = SampleLineage.objects.create(parent=self.extracted_from, child=obj, process_measurement=self.process_measurement)
+
+        # automatic project inheritance
+        for project in self.extracted_from.projects.all():
```
UlysseFG, line 215: self.extracted_from.projects.all() just raised a question in my mind ... does projects get the projects relation or the projects property of sample ? I think we should rename the relation (and the related_name) to make sure there is no ambiguity ... Maybe rename the manytomany relation on project to "samples_linked" and related_name to "projects_linked" that way we can be sure which one is used. In the case here, you point to the property and the all() would be redundant.

backend/fms_core/resources/extraction.py
```diff
@@ -210,6 +210,12 @@ def after_save_instance(self, instance, using_transactions, dry_run):
...
+            print(project)
```
UlysseFG, line 216: Remove this print please once you are done.

backend/fms_core/resources/transfer.py
```diff
@@ -183,6 +183,11 @@ def after_save_instance(self, instance, using_transactions, dry_run):
     def save_m2m(self, obj, data, using_transactions, dry_run):
         if self.transferred_from and self.process_measurement:
             SampleLineage.objects.create(parent=self.transferred_from, child=obj, process_measurement=self.process_measurement)
+
+            # automatic project inheritance
+            for project in self.transferred_from.projects.all():
```
UlysseFG, line 188: same as above ...

backend/fms_core/tests/test_resources.py
```diff
@@ -70,6 +70,10 @@ def setUp(self) -> None:
         self.rr = ContainerRenameResource()
         self.pr = ProjectLinkSampleResource()
 
+    def create_projects(self):
+        self.project1, _ = Project.objects.get_or_create(name="TestProject1", status="Ongoing")
```
UlysseFG, line 74: I think we should have a default status for projects in order to prevent having to initialize them everywhere to a hard coded identical string.

backend/fms_core/resources/extraction.py
sebastianballesteros: I think it retrieves the projects property of sample, i.e. if that extracted sample is linked to PROJECT1 and PROJECT2 then projects.all() will be those 2 project objects. If the change is necessary do you want me to do it on this PR or create another one?

backend/fms_core/tests/test_resources.py
sebastianballesteros: You mean modifying the attribute of "status" in Project model by adding a default value of "Ongoing(Open)"

backend/fms_core/resources/extraction.py
UlysseFG: well it is not directly linked to this PR so if you do not need to do it here...  If this uses the property the property already returns projects.all(). more .all() do not really hurt it is just unnecessary i think.

backend/fms_core/tests/test_resources.py
UlysseFG: yeah i think most of the time we'll create Open Projects (is Ongoing or Open the right status?)...

sebastianballesteros: Yes, I agree that most of the times the user will create a Project it will have an Open status

UlysseFG:
Maybe i was a bit early in my review but there does not seem to be any inheritance (sampleByProject creation) in experiment_run.py ressource...

backend/fms_core/tests/test_resources.py
UlysseFG: If Open is the correct status don't forget to fix that on the receiving end PR if you want to use ongoing here ...

sebastianballesteros:
Inheritance in experiment_run resource is done automatically since in the code you are getting the source_sample and then setting the pk to null so it inherits all attributes. In any case I test the inheritance in the test_resource.py

UlysseFG:
trouble is that project is not a fk on sample ... since you do create a new sample you need to link the new sample IDs to the project in the samplebyproject table... i was wondering about the tests but "ss" stands for sample source you should be testing "se" for sample experiment to adequatly testing the inheritance...

backend/fms_core/tests/test_resources.py
jleblond: Agree. Default should be Ongoing.

backend/fms_core/resources/extraction.py
jleblond: Yup I wasnt too sure about this one too. Reimplementing a relationship with a property in Sample. First, in what case do we need to have that relationship reimplemented through a property instead of having the relationship directly?

UlysseFG: It gives a different tool : the property was made to provide a simplified queryset to the records, the relation provides an object list if i recall correctly ... This is not that useful but so are most properties ... I just wanted to make sure there was no name collision between them.

backend/fms_core/tests/test_resources.py
sebastianballesteros: So, Open/Close are the correct options. However, the active PR's are not updated to these options (since they were merged a few days ago), we would have to rebase.

backend/fms_core/resources/extraction.py
jleblond: Yes at least have different names for them so that you dont get one when you expect the other one and vice versa

UlysseFG (approved):

PR #142, opened 2021-09-24
Added projects to the jumpbar search

UlysseFG:


jleblond (approved):
Works like a charm! :) 

I was also curious about why you removed the all() from the queries in viewsets/projects.py? Does it have an impact on performance?

UlysseFG:
> Works like a charm! :)
> 
> I was also curious about why you removed the all() from the queries in viewsets/projects.py? Does it have an impact on performance?

I would not think so. it is just to keep it simple. It is more a cleanup than a performance thing ... I found out that the all() was not required while checking on the count() for performance reasons ;).

PR #144, opened 2021-09-28
Rename Project attribute 'comment' for naming convention

jleblond:


sebastianballesteros:
I don't think this is needed (It does not break the front-end tests because "comment" is a substring of "comments"),  but you may want to change the integration tests from #comments to #comment in the projects.spec.js file to have consistency

jleblond:
> I don't think this is needed (It does not break the front-end tests because "comment" is a substring of "comments"), but you may want to change the integration tests from #comments to #comment in the projects.spec.js file to have consistency

Right you mean this line here? cy.get('#comments').type("This is a test comment.")   
To change it to comment? Yup I think it would be great to change it!

UlysseFG (approved):

PR #145, opened 2021-09-28
Fix samples by project lists

sebastianballesteros:


jleblond:
I would suggest not duplicating itemsByID. From what I understand, you now have itemsByID and itemsByProjectByID which would both contain Samples objects serialized, by ID. I don't think we should duplicate this source of information. All the SamplesByID  should go in the same itemsByID location.

Also, I am not too sure either about storing all of the items ID in the store too.

Eventually, it would be great to have the local setFilter, local variables and all, in a separate component, untied from the  Sample Project association page, and as a component that is reusable (will be helpful for the display of the list of projects by Sample, and other possible relations).

To be discussed with @UlysseFG at what point this is needed for this 3.4 release, again, but I'd prefer to soonish have a better separation of concerns/components in the front-end files and in the store.

sebastianballesteros:
I agree with the reusable and untied local variables. For the concerns about the store, are you suggesting just storing the ID's (not the objects) and just filtering itemsByID manually in the component?

UlysseFG:
> I agree with the reusable and untied local variables. For the concerns about the store, are you suggesting just storing the ID's (not the objects) and just filtering itemsByID manually in the component?

I think we have a misunderstanding about itemsByID ... This is the repository for entities, it is not filterer or manipulated, it's just updated with the latest or missing entities needed by the frontend. We are speaking of another parameter that is used to filter the item list that is currently in the store by entity. this is the one the provide the items to show on the page and that can change according to filters. That item list should be made local for any query that has default filtering ... we don't want to erase the default list used in the sample, container, individual sections ... we want a temporary (or local) one that is used in the project sample list. the itemsByID should not be affected it should still be the entity repository everywhere.

jleblond:
> > I agree with the reusable and untied local variables. For the concerns about the store, are you suggesting just storing the ID's (not the objects) and just filtering itemsByID manually in the component?
> 
> I think we have a misunderstanding about itemsByID ... This is the repository for entities, it is not filterer or manipulated, it's just updated with the latest or missing entities needed by the frontend. We are speaking of another parameter that is used to filter the item list that is currently in the store by entity. this is the one the provide the items to show on the page and that can change according to filters. That item list should be made local for any query that has default filtering ... we don't want to erase the default list used in the sample, container, individual sections ... we want a temporary (or local) one that is used in the project sample list. the itemsByID should not be affected it should still be the entity repository everywhere.

Exactly. If you need to fetch the list of Samples in a ProjectDetailPage, that list of Samples should be stored (and intact, not filtered) in the samples.itemsByID part of the store.  So you have access to those objects whenever needed.

What is being filtered is the items  (itemsIDS) list [], which as Ulysse mentioned should be somewhere else (different than sample.items) or ideally stored as a local variable.

Let me know if it is not clear, we can probably work on it tomorrow.

jleblond:
I am getting the following error when changing from a Project detail where there are aswsociated samples, to one with no associated samples:

image


Also I'm not quite sure we should re-use the list() action/reducer already existing to do the new changes. I would create a new action/reducer for that specific purpose of storing temporaryItems. Maybe we could call it listTemporaryItems()

(Review is in progress, will continue testing, looking and commenting)

frontend/src/components/FilteredList.js
```diff
@@ -0,0 +1,120 @@
...
+          disabled={nFiltersForWarning === 0}
+          onClick={clearFilters}
+        >
+          Clear Filters
+        </Button>
+      </div>
+      <PaginatedTable
+        columns={columns}
+        items={items}
+        itemsByID={itemsFiltered}
```
UlysseFG, line 106: You are still "filtering" the wrong parameter ... items is the list of items to filter, itemsByID is the same all across the application and contain the dictionary of all entity items...

jleblond: I have to dig more into the code and play around it, but I think it is just the name of the variable that leads to confusion. A few lines upper, you see the line 
` const itemsFiltered = isFetching ? {} : itemsByID;`

so it appears to be the itemsByID but with a name that can be confusing?

sebastianballesteros: You are right, my bad. I meant to filter items.

sebastianballesteros: it was meant to be `const itemsFiltered = isFetching ? [] : items;`

frontend/src/modules/projects/actions.js
```diff
@@ -42,7 +42,21 @@ export const update = (id, project) => async (dispatch, getState) => {
 };
 
 export const list = (options) => async (dispatch, getState) => {
-    const params = { limit: 100000, ...options }
+    //Prevents the default fetch without any filters
+    if (options.filters && !options.filters[PROJECT_FILTERS.samples__id.key]["value"])
+        return;
+
+    const limit = getState().pagination.pageSize;
+    const filters = options.filters ? serializeFilterParams(options.filters, PROJECT_FILTERS) : {}
```
UlysseFG, line 50: Can options.filters be falsy ? you just tested it to be true ...

frontend/src/components/projects/ProjectsAssociatedSamples.js
```diff
@@ -75,93 +72,51 @@ const mapStateToProps = state => ({
...
-      />
-    </PageContent>
+    <FilteredList
+      description = {SAMPLE_FILTERS}
+      columns={columns}
+      list={list}
+      items={samples}
+      itemsByID={samplesByID}
+      filterID={projectID}
+      filterKey={filterKey}s
```
UlysseFG, line 111: is this trailing 's' a Typo ?

frontend/src/components/projects/ProjectsAssociatedSamples.js
```diff
@@ -75,93 +72,51 @@ const mapStateToProps = state => ({
...
-        totalCount={totalCount}
-        page={page}
-        filters={filters}
-        sortBy={sortBy}
-        onLoad={listTable}
-        onChangeSort={setSortBy}
-      />
-    </PageContent>
+    <FilteredList
+      description = {SAMPLE_FILTERS}
```
jleblond, line 105: space before and after = (lint)

jleblond:
By examing the store, I am wondering why the filteredItems array [] has that many items IDS? shouldnt it have only the first 20 items IDS displayed ?

image

frontend/src/modules/samples/actions.js
```diff
@@ -51,6 +52,22 @@ export const list = (options) => async (dispatch, getState) => {
     ));
 };
 
+export const listFilter = ({ offset = 0, limit = DEFAULT_PAGINATION_LIMIT, filters = {}, sortBy, filterKey }) => async (dispatch, getState) => {
```
UlysseFG, line 55: Not sure why you need filterKey in here ... You can test empty filters otherwise without having a key.
Also why is there no abort param for this list action ? i don't think its used a lot but i think it is cleaner to have it for every list action.

sebastianballesteros: When the component is mounted, for a few seconds, the filtering ID is undefined. Thus, checking if the ID is not undefined is necessary to filter properly.

UlysseFG: This is an element related to the component and it's loading state ... Shouldn't this be part of isFetching or isLoading ? I do not see this as normal to include in a function a value that is not used in it just for the sake of testing it. if its undefined when you call listFilter ... you can avoid calling listFilter since you'll just exit anyway ...

sebastianballesteros: Yes, that could work better actually

backend/fms_core/serializers.py
```diff
@@ -275,11 +275,9 @@ class Meta:
         depth = 1
 
 class ProjectSerializer(serializers.ModelSerializer):
-    samples = serializers.PrimaryKeyRelatedField(read_only=True, many=True)
     class Meta:
         model = Project
-        fields = "__all__"
-        extra_fields = ('samples')
+        fields = ("id", "name", "principal_investigator", "requestor_name", "requestor_email", "targeted_end_date", "status")
```
jleblond, line 280: I'm curious as to why you changed the fields = "__all__" to explicitely listing all the fields from the Project model?

frontend/src/utils/api.js
```diff
@@ -78,13 +78,13 @@ const api = {
...
   },
-  
+
   projects: {
     get: projectId => get(`/projects/${projectId}/`),
     add: project => post("/projects/", project),
     update: project => patch(`/projects/${project.id}/`, project),
     list: (options, abort) => get("/projects", options, { abort }),
-    listExport: options => get("/project/list_export/", {format: "csv", ...options}),
+    listExport: options => get("/projects/list_export/", {format: "csv", ...options}),
```
jleblond, line 87: Good catch :)

backend/fms_core/serializers.py
sebastianballesteros: Because I think actually the extra fields were superfluous (cause "samples" is included in __alll__)

sebastianballesteros: so I need to explicitly list them to avoid __samples__ to get serialized

UlysseFG: It is a Model Serializer ... samples are not part of the Project model so it should not appear in _all_ i think ...

backend/fms_core/serializers.py
```diff
@@ -275,11 +275,9 @@ class Meta:
...
 
 class ProjectExportSerializer(serializers.ModelSerializer):
     samples = serializers.PrimaryKeyRelatedField(read_only=True, many=True)
```
UlysseFG, line 283: We should remove or fix the samples in the export. It is not working at the moment... Not sure we need the list of samples here.

UlysseFG:
I have a problem with the "Item created by" and "Last modification by" tracking information not loading for projects when i test. Is it only on my side ?

UlysseFG (approved):

PR #147, opened 2021-10-01
Importer container creation

UlysseFG:


backend/fms_core/services/container.py
```diff
@@ -13,41 +15,87 @@ def get_container(barcode=None):
...
+                try:
+                    container = Container.objects.create(**container_data, comment=comment)
+                    created_entity = True
+                # Pile up all validation error raised during the creation of the container
+                except ValidationError as e:
+                    errors.append(';'.join(e.messages))
+    else:
+        errors.append(f"Barcode is required to get or create a container.")
+
+    return (container, created_entity, errors, warnings)
```
jleblond, line 69: Good idea to have such an indicator. What would you think about generalizing this kind of indicator so that all the service functions return it? Could be useful

backend/fms_core/template_paths.py
```diff
@@ -17,7 +17,7 @@
     "PROJECT_LINK_SAMPLES_TEMPLATE",
 ]
 
-CONTAINER_CREATION_TEMPLATE = static("submission_templates/Container_creation_v3_4_0.xlsx")
+CONTAINER_CREATION_TEMPLATE = static("submission_templates/Container_creation_v3_5_0.xlsx")
```
jleblond, line 20: Minor thing here: for the T.I. branch, I renamed all the templates to _vtest.xlsx so we have a quick way of identifying which one will go with the T.I changes.

backend/fms_core/services/container.py
jleblond: I mean the created_entity indicator. So all the services would return if the item was created, updated, or retrieved for example with a string

jleblond:
Is it expected that the Excel template does not have some container kinds such as plates?

jleblond:
I am not sure about some of the logic or data cleaning in the importer or row_handler for the container creation. I get this error with this template, the container parent barcode is empty in the template, it should not return an error

image

backend/fms_core/services/container.py
UlysseFG: This is specifically for the get_or_create so the business end can know if this is a new entity or a fetched one. We should indeed add this indicator to all get_or create.

As for it being a string: no. This is only to differentiate between those two options. It already complicates the error handling anymore than this and it will start to be more an hindrance than anything.

backend/fms_core/template_paths.py
UlysseFG: Ok i'll rename the file to in line with the rest.

UlysseFG:
> Is it expected that the Excel template does not have some container kinds such as plates?

Yeah the sample holding containers are created when handling samples (sample submission, transfer, extractions ...)

UlysseFG:
> I am not sure about some of the logic or data cleaning for the container creation. I get this error with this template, the container parent barcode is empty in the template, it should not return an error
> 
> image

Yeah there must be a case i missed ... oh yeah i see. This was meant for the get_or_create ... Not sure it was working on the main branch though...

backend/fms_core/services/container.py
jleblond: Ok that works for me!

PR #149, opened 2021-10-05
Addition of  the creation_date column in the sample list of the project detail page.

UlysseFG:


jleblond:
Aren't there filters and sorters too for that column?

jleblond (approved):

PR #151, opened 2021-10-14
Add project name case index

jleblond:


backend/fms_core/models/project.py
```diff
@@ -45,6 +45,10 @@ def clean(self):
         def add_error(field: str, error: str):
             _add_error(errors, field, ValidationError(error))
 
+        projects_names_similar = list(Project.objects.filter(name__iexact=self.name).values_list('name', flat=True))
+        if projects_names_similar:
+            add_error("name", f"Project(s) named {projects_names_similar} exist and are too simlar to {self.name}")
```
UlysseFG, line 50: I don't see how more than one project could exist in that list ... Also you should point out what too similar implies specifically. Something like "Two project names cannot be distinguished only by letter case."

PR #156, opened 2021-10-18
Template importer sample transfer

UlysseFG:
Missing tests.


UlysseFG:
> Missing tests.

Tests added. Did a small modification to ti extraction to prevent test from failing...

backend/fms_core/template_importer/importers/transfer.py
```diff
@@ -0,0 +1,67 @@
...
+        self.preloaded_data = {'process': None}
+
+        self.preloaded_data['process'] = Process.objects.create(protocol=Protocol.objects.get(name="Transfer"),
+                                                                comment="Sample Transfer (imported from template)")
+
+    def import_template_inner(self):
+        sheet = self.sheets['SampleTransfer']
+
+        for row_id, row_data in enumerate(sheet.rows):
+            volume_used_decimal = float_to_decimal(row_data['Volume Used (uL)']) if row_data['Volume Used (uL)'] else None
```
jleblond, line 27: Please use the new helper function float_to_decimal_and_none

backend/fms_core/services/process_measurement.py
```diff
@@ -9,15 +9,25 @@ def create_process_measurement(process=None, source_sample=None, execution_date=
...
-    except ValidationError as e:
-        errors.append(f"Could not create ProcessMeasurement with Process ID {process.id} and Source sample {source_sample}")
+    # Validate parameters
+    if not process:
+        errors.append(f"Process is required for process measurement creation.")
+    if not source_sample:
+        errors.append(f"Source sample is required for process measurement creation.")
+    if not execution_date:
+        errors.append(f"Execution date is required for process measurement creation.")
+
```
jleblond, line 19: This might be useful for the devs to know if there is any problem with the code, but I doubt this information would be any helpful to the user if displayed in the front-end in the preview table

backend/fms_core/services/sample_lineage.py
```diff
@@ -7,12 +7,21 @@ def create_sample_lineage(parent_sample, child_sample, process_measurement):
...
-                                                      )
-    except ValidationError as e:
-        errors.append(str(e))
+     # Validate parameters
+    if not parent_sample:
+        errors.append(f"Parent sample is required for sample lineage creation.")
+    if not child_sample:
+        errors.append(f"Child sample is required for sample lineage creation.")
+    if not process_measurement:
+        errors.append(f"Process measurement is required for sample lineage creation.")
```
jleblond, line 16: Same comment as the one I made previously, I wonder if this is helpful for the user having it displayed in the error messages

backend/fms_core/template_importer/importers/transfer.py
```diff
@@ -0,0 +1,67 @@
...
+        {'name': 'SampleTransfer', 'header_row_nb': 5},
+    ]
+
+    def __init__(self):
+        super().__init__()
+        # Preload objects accessible to the whole template (not only by row)
+        self.preload_data_from_template()
+
+
+    def preload_data_from_template(self):
```
jleblond, line 17: Rename the preload_data_from_template function to whats been changed in the root template-importer branch

backend/fms_core/template_importer/row_handlers/transfer/transfer.py
```diff
@@ -0,0 +1,73 @@
...
+
+class TransferRowHandler(GenericRowHandler):
+
+    def process_row_inner(self, source_sample_info, resulting_sample_info, process_measurement_info):
+        original_sample, self.errors['sample'], self.warnings['sample'] = get_sample_from_container(
+            barcode=source_sample_info['container']['barcode'],
+            coordinates=source_sample_info['coordinates'])
+
+        if original_sample:
+            if source_sample_info['depleted']:
```
jleblond, line 19: Is it possible that there is no Depleted information entered?

backend/fms_core/services/sample_lineage.py
UlysseFG: Services do not pick and choose error messages. If there is an error we identify it and return it. If there is a choice to be made about what is to be reported to the user it is done elsewhere ... The caller can decide to combine all of these with another simplified message (sample lineage will not return an error unless there is something else that is going wrong)... We could just set a generic message at the importer level if there is any sample lineage error ... It is likely accompanied by more useful error messages from parent sample, child sample or sample_measurement.

backend/fms_core/template_importer/row_handlers/transfer/transfer.py
```diff
@@ -17,7 +18,7 @@ def process_row_inner(self, source_sample_info, resulting_sample_info, process_m
 
         if original_sample:
             if source_sample_info['depleted']:
-                original_sample.depleted = True if source_sample_info['depleted'] == "YES" else False
+                original_sample.depleted = original_sample.depleted or check_truth_like(source_sample_info['depleted'])
```
jleblond, line 21: This reminds me that we should add a way of checking whether the sample is depleted or not before attempting a transfer or extraction. Also, there is the case where this cell is not filled in the spreadsheet; it should have no impact (no change) to the original_sample; we should be careful not to have it set to False (not depleted) if the field is None (not filled)

PR #168, opened 2021-11-01
Converting experiment type to run type

UlysseFG:


backend/fms_core/experiments.py
```diff
@@ -1,4 +1,4 @@
-PROTOCOLS_BY_EXPERIMENT_TYPE_NAME = {
+PROTOCOLS_BY_RUN_TYPE_NAME = {
```
jleblond, line 1: Do we still need this dictionary if it is to be stored in a migration?

backend/fms_core/models/run_type.py
```diff
@@ -3,16 +3,16 @@
...
 
 from ._constants import STANDARD_NAME_FIELD_LENGTH
 from ._validators import name_validator
 from ._utils import add_error as _add_error
 
-__all__ = ["ExperimentType"]
+__all__ = ["RunType"]
 
-PROTOCOLS_BY_EXPERIMENT_TYPE_NAME = {
+PROTOCOLS_BY_RUN_TYPE_NAME = {
```
jleblond, line 15: Same comment as previously, didnt you get rid of dictionaries with Protocol list to have that stored in the db?

backend/fms_core/models/run_type.py
```diff
@@ -50,13 +53,12 @@ def save(self, *args, **kwargs):
         self.full_clean()
         super().save(*args, **kwargs)  # Save the object
 
-
-    # This method should be moved out of ExperimentType when Workflows are implemented in FreezeMan
+    # REPLACE and move to procotol
     @property
     def get_protocols_dict(self):
```
jleblond, line 58: Same comment as the 2 previous ones.. shouldnt we fetch this from the db from now on ?

backend/fms_core/tests/valid_templates/Experiment_Infinium_24_v3_4_0_B_A_1.csv
```diff
@@ -1,6 +1,6 @@
 Experiment Submission Template,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```
jleblond, line 1: I think this file won't be used anymore. The tests are not made from the CSV files for the importers

backend/fms_core/experiments.py
UlysseFG: I decided not to do real changes just to convert the model... This will go away in another PR... but yeah this is part of the things that will go away...

backend/fms_core/tests/valid_templates/Experiment_Infinium_24_v3_4_0_B_A_1.csv
UlysseFG: yeah i know but i changed it in case there was still stuff tied to it ... We need to do some cleanup ...

PR #170, opened 2021-11-02
Implemented sample db view

sebastianballesteros:


frontend/src/utils/serializeFilterParams.js
```diff
@@ -10,6 +10,7 @@ export default function serializeFilterParams(filters, descriptions) {
   Object.keys(filters).forEach(field => {
     const value = filters[field]?.value
     const description = descriptions[field]
+    console.log(descriptions)
```
jleblond, line 13: Dont forget to remove it :)

sebastianballesteros: oops thanks :)

backend/fms_core/migrations/0027_v3_5_0.py
```diff
@@ -0,0 +1,60 @@
...
+        migrations.AlterField(
+            model_name='derivedsample',
+            name='sample',
+            field=models.ForeignKey(help_text='Sample associated to this DerivedSample', on_delete=django.db.models.deletion.PROTECT, related_name='derived_samples', to='fms_core.sample'),
+        ),
+        migrations.RunSQL(
+            """
+                DROP VIEW IF EXISTS fms_core_fullsample;
+                CREATE OR REPLACE VIEW fms_core_fullsample AS
+                SELECT sample.id, sample.id AS sample_id, sample.name, sample.container_id, sample.coordinates, sample.volume, sample.concentration, sample.depleted, array_remove(array_agg(sbyp.project_id), NULL) AS projects, array_remove(array_agg(pm.id), NULL) AS process_measurements_id, array_remove(array_agg(sl.parent_id), NULL) AS child_of_id, sample.creation_date, sample.created_at, sample.updated_at, sample.created_by_id, sample.updated_by_id, sample.deleted, sample.update_comment, sample.alias, derived.sample_kind_id, derived.tissue_source, derived.experimental_group, derived.id AS derived_sample_id, derived.biosample_id,  biosample.individual_id, biosample.collection_site
```
jleblond, line 43: Are both SELECT sample.id, sample.id AS sample_id necessary?

Also I suggest escaping the long line for readability (have it in 2 or more lines)

backend/fms_core/migrations/0027_v3_5_0.py
```diff
@@ -0,0 +1,60 @@
...
+                JOIN fms_core_sample AS sample 
+                ON dbys.sample_id =  sample.id 
+                JOIN fms_core_derivedsample AS derived
+                ON dbys.derived_sample_id  = derived.id 
+                LEFT JOIN fms_core_samplebyproject AS sbyp
+                ON sbyp.sample_id = sample.id
+                LEFT JOIN fms_core_processmeasurement AS pm 
+                ON pm.source_sample_id = sample.id
+                LEFT JOIN fms_core_samplelineage AS sl 
+                ON sl.child_id = sample.id
```
jleblond, line 54: Was the # of sample still ok?

Does the number of entries in that full_sample table equivalent to the number of derived_by_sample?

backend/fms_core/models/full_sample.py
```diff
@@ -0,0 +1,80 @@
...
+from .process_measurement import ProcessMeasurement
+from django.contrib.auth.models import User
+from django.contrib.postgres.fields import ArrayField
+
+__all__ = ["FullSample"]
+
+
+class FullSample(models.Model):
+    """ Class to provide information about a sample as a view. """
+    sample_kind = models.ForeignKey(SampleKind, on_delete=models.DO_NOTHING, db_column='sample_kind_id',
```
jleblond, line 17: I dont think those on_delete attributes are necessary on FullSample attributes. This is a read-only model, so theoretically there should be not deletion operation performed

https://stackoverflow.com/a/35780859

backend/fms_core/migrations/0027_v3_5_0.py
sebastianballesteros: I think so, at the beginning I just had one sample.id. However, I wanted to store a "Sample" object reference in the Full Sample model to be able to filter the sample fields, i.e. "sample__projects__name"

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,26 @@
...
+from rest_framework.response import Response
+
+from fms_core.models import FullSample
+from fms_core.serializers import FullSampleSerializer
+
+from ._utils import _list_keys
+from ._constants import _full_sample_filterset_fields
+
+class FullSampleViewSet(viewsets.ModelViewSet):
+    queryset = FullSample.objects.select_related("individual", "container", "sample_kind", "biosample", "derived_sample").all().distinct()
```
jleblond, line 12: What FullSample attribute is the distinct performed on ? The FullSample id?

jleblond (commented):
Seems to be working well in general, just noticed a few things:

Projects seem to appear many times in some cases:
image


Don't forget the detached advanced filters for samples:
image


Don't forget projects in the export (or simply remove the column if it is not to be used)
image

backend/fms_core/migrations/0027_v3_5_0.py
sebastianballesteros: Yes!

backend/fms_core/models/full_sample.py
sebastianballesteros: Screen Shot 2021-11-10 at 2 31 26 PM

Django does complain if the attributes are not there

backend/fms_core/viewsets/full_sample.py
sebastianballesteros: I think distinct here is superfluous, I used it when I was trying to do prefetch_related() but now we don't need it anymore.

backend/fms_core/models/full_sample.py
UlysseFG, line 17: Missing alias... make sure all old sample fields are present at least.

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,44 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
         else:
             return obj.container.location.barcode
 
     def get_current_volume(self, obj):
-        return obj.volume
+        return obj.volume if obj.volume else None
 
     def get_father_name(self, obj):
-        father = '' if obj.individual.father is None else obj.individual.father.name
+        father = '' if obj.individual and obj.individual.father is None else obj.individual.father.name
```
UlysseFG, line 209: This will very likely crash if obj.individual is Falsy.

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,44 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
         return father
 
     def get_mother_name(self, obj):
-        mother = '' if obj.individual.mother is None else obj.individual.mother.name
+        mother = '' if obj.individual and obj.individual.mother is None else obj.individual.mother.name
```
UlysseFG, line 213: Same as above ,,, you likely want or instead of and...

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,44 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
+    process_measurements = serializers.PrimaryKeyRelatedField(source='process_measurement', many=True, read_only=True)
+    projects = serializers.PrimaryKeyRelatedField(read_only=True, many=True)
+
+    class Meta:
+        model = Sample
+        fields = "__all__"
+        extra_fields = ('extracted_from', 'projects')
+
+    def get_extracted_from(self, obj):
+        if obj.extracted_from is None:
```
UlysseFG, line 227: maybe simply `return obj.extracted_from and obj.extracted_from.id`

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,216 @@
...
+        }
+    ]
+
+    def create(self, request, *args, **kwargs):
+        error = {}
+        full_sample = request.data
+
+        #Container related information
+        try:
+            container_obj = Container.objects.get(pk=full_sample['container'])
```
UlysseFG, line 58: I would not manually test each error ... you are duplicating the "clean" behaviour. i would create the biosample, derived sample and sample. The clean used to return the correct error mapping you can keep using it. It also creates some overhead (not much but if it is not necessary...)

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,216 @@
...
+                ])
+
+            derived_sample = DerivedSample.objects.create(**derived_sample_data)
+
+            sample_data = dict(
+                name=full_sample['name'],
+                volume=full_sample['volume'],
+                creation_date=full_sample['creation_date'],
+                container=container_obj,
+                comment=(full_sample['comment'] or (f"Automatically generated on {datetime.utcnow().isoformat()}Z")),
```
UlysseFG, line 104: why change the input of the user ? All field should be exactly as submitted (other than the possible small formatting)... It was not there before i do not see why it need to be there now ... Users expect to have an empty comment inserted ...

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,216 @@
...
+        serializer = SampleSerializer(sample)
+        data = serializer.data
+        return Response(data)
+
+    def update(self, request, *args, **kwargs):
+        full_sample = request.data
+
+        #Retrive the sample to update
+        try:
+            sample_to_update = Sample.objects.get(pk=full_sample['id'])
```
UlysseFG, line 127: sample_to_update = Sample(**request.data) ?

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,216 @@
...
+            },
+            "tissue_source_counts": {
+                c["tissue_source"]: c["tissue_source__count"]
+                for c in DerivedSample.objects.values("tissue_source").annotate(Count("tissue_source"))
+            },
+            "collection_site_counts": {
+                c["collection_site"]: c["collection_site__count"]
+                for c in Biosample.objects.values("collection_site").annotate(Count("collection_site"))
+            },
+            "experimental_group_counts": dict(experimental_groups),
```
UlysseFG, line 192: Are these counts used somewhere ? we may want to get rid of some of them if they are not useful. summary is called very often and having those extra counts (the total is displayed in the dashboard) may be wasted.

sebastianballesteros: I don't think so, imo we can get rid of them indeed

backend/fms_core/viewsets/full_sample.py
sebastianballesteros: What do you mean by each error? Container, Individual and Sample Kind are given directly by the frontend (id's) so I need to fetch the objects anyways to be able to create biosample/derived/sample. Or you mean not to wrap them in a try/except block?

UlysseFG: you receive the primary key from the frontend for each of container individual and sample kind ... you do not need to get them. you just need to set new_sample.individual_id = full_sample["individual"]. If the given ID is not correct for some reason you'll get an "individual" error back from the model that you can catch and forward to the front end.

backend/fms_core/viewsets/full_sample.py
sebastianballesteros: What?

UlysseFG: you already have the date and id ... you can just load the object using the data and save it as far as i know, maybe you need to add the tracking info ... but i am not sure you need to go one attribute at a time fetching intermediate objects.

backend/fms_core/migrations/0027_v3_5_0.py
```diff
@@ -0,0 +1,68 @@
...
+                ('container', models.ForeignKey("Container", on_delete=models.DO_NOTHING, help_text="Container associated to the sample.")),
+                ('individual', models.ForeignKey("Individual", blank=True, null=True, on_delete=models.DO_NOTHING, help_text="Individual associated with to sample.")),
+                ('sample_kind', models.ForeignKey("SampleKind", on_delete=models.DO_NOTHING, help_text="Sample Kind associated to the sample.")),
+                ('derived_sample', models.ForeignKey("DerivedSample", on_delete=models.DO_NOTHING, help_text="Sample from which the current sample was extracted from.")),
+                ('biosample', models.ForeignKey("Biosample", on_delete=models.DO_NOTHING, help_text="Root Sample")),
+                ('child_of', models.ForeignKey("Sample", on_delete=models.DO_NOTHING, help_text="Parent Sample")),
+                ('process_measurements', models.ForeignKey("ProcessMeasurement", on_delete=models.DO_NOTHING, help_text="Process measurements associated to the sample")),
```
UlysseFG, line 36: This is unlikely to work ... each sample can have many process measurement related to it ...

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,41 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
         else:
             return obj.container.location.barcode
 
     def get_current_volume(self, obj):
-        return obj.volume
+        return obj.volume if obj.volume else None
 
     def get_father_name(self, obj):
-        father = '' if obj.individual.father is None else obj.individual.father.name
+        father = '' if obj.individual or obj.individual.father is None else obj.individual.father.name
```
UlysseFG, line 209: Still not working ... you should have ! obj.individual unless you want the father to always be "".

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,41 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
-                  'individual_id', 'sex', 'pedigree', 'mother_name', 'father_name',
-                  'current_volume', 'concentration', 'collection_site', 'tissue_source', 'creation_date',
-                  'depleted', 'coordinates', 'projects', 'comment' )
+        model = FullSample
+        fields = "__all__"
+        extra_fields = ("taxon", "sex", "pedigree", "cohort", "container_name", "container_kind", "container_barcode")
 
     def get_location_barcode(self, obj):
-        if obj.container.location is None:
+        if obj.container and obj.container.location is None:
```
UlysseFG, line 200: only works because obj.container is always true but otherwise it would be the same as for the mother and father case ...

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,184 @@
...
+from fms_core.template_importer.importers import SampleSubmissionImporter, SampleUpdateImporter
+
+from fms_core.template_paths import SAMPLE_SUBMISSION_TEMPLATE, SAMPLE_UPDATE_TEMPLATE
+
+from ._constants import _full_sample_filterset_fields
+from ._utils import TemplateActionsMixin, _list_keys, versions_detail
+from fms_core.services.sample import create_full_sample
+
+class FullSampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
+    queryset = FullSample.objects.select_related("individual", "container", "sample_kind", "biosample", "derived_sample").all()
```
UlysseFG, line 26: biosample and derived_sample should not be part of the select_related. You should have all their content already in full sample. The only use of those FK are the id.

backend/fms_core/models/full_sample.py
```diff
@@ -0,0 +1,83 @@
...
+    biosample = models.ForeignKey(Biosample, on_delete=models.DO_NOTHING, related_name="samples",
+                                          help_text="Designated location of the sample.")
+
+    projects = ArrayField(models.IntegerField(blank=True))
+
+    projects_names = ArrayField(models.CharField(blank=True, max_length=20))
+
+    child_of = models.ForeignKey(Sample, on_delete=models.DO_NOTHING)
+
+    process_measurements = models.ForeignKey(ProcessMeasurement, on_delete=models.DO_NOTHING)
```
UlysseFG, line 72: This should be an ArrayField as projects. Your view aggregate the process_measurements ids.

sebastianballesteros: In that case child_of, process_measurements, projects and project_names would all have to be ArrayFields in the model? Since they are practically returned as arrays in the query

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,184 @@
...
+        {
+            "name": "Update Samples",
+            "description": "Upload the provided template with up to 384 samples to update.",
+            "template": SAMPLE_UPDATE_TEMPLATE,
+            "importer": SampleUpdateImporter,
+        }
+    ]
+
+    def get_queryset(self):
+        project_name_filter = self.request.query_params.get('sample__projects__name__icontains')
```
UlysseFG, line 53: This is a strange place to hide a filter ... Can't you make an annotation for this to list it as a regular filter ?

backend/fms_core/models/full_sample.py
sebastianballesteros: i.e. ` projects = ArrayField(models.IntegerField(blank=True))

    projects_names = ArrayField(models.CharField(blank=True, max_length=20))

    child_of = ArrayField(models.CharField(blank=True, max_length=20))

    process_measurements = ArrayField(models.CharField(blank=True, max_length=20))`

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,184 @@
...
+        },
+        {
+            "name": "Update Samples",
+            "description": "Upload the provided template with up to 384 samples to update.",
+            "template": SAMPLE_UPDATE_TEMPLATE,
+            "importer": SampleUpdateImporter,
+        }
+    ]
+
+    def get_queryset(self):
```
UlysseFG, line 52: Where did the whole recursive search part go ?

backend/fms_core/models/full_sample.py
sebastianballesteros: ` projects = ArrayField(models.IntegerField(blank=True))

    projects_names = ArrayField(models.CharField(blank=True, max_length=20))

    child_of = ArrayField(models.CharField(blank=True, max_length=20))

    process_measurements = ArrayField(models.CharField(blank=True, max_length=20))`

sebastianballesteros: `
 projects = ArrayField(models.IntegerField(blank=True))

    projects_names = ArrayField(models.CharField(blank=True, max_length=20))

    child_of = ArrayField(models.CharField(blank=True, max_length=20))

    process_measurements = ArrayField(models.CharField(blank=True, max_length=20))
`

sebastianballesteros: `
    projects = ArrayField(models.IntegerField(blank=True))

    projects_names = ArrayField(models.CharField(blank=True, max_length=20))

    child_of = ArrayField(models.CharField(blank=True, max_length=20))

    process_measurements = ArrayField(models.CharField(blank=True, max_length=20))
`

frontend/src/components/filters/descriptions.js
```diff
@@ -35,14 +35,14 @@ export const SAMPLE_FILTERS = {
     label: "Container Barcode",
     recursive: true,
   },
-  projects__name: {
+  sample__projects__name: {
     type: FILTER_TYPE.INPUT,
-    key: "projects__name",
+    key: "sample__projects__name",
```
UlysseFG, line 40: why not use the correct field name and add it to the backend full_sample filter_fields ?

backend/fms_core/serializers.py
```diff
@@ -202,33 +190,41 @@ class SampleExportSerializer(serializers.ModelSerializer):
...
+
+    class Meta:
+        model = Sample
+        fields = "__all__"
+        extra_fields = ('extracted_from', 'projects')
+
+    def get_extracted_from(self, obj):
+        return obj.extracted_from and obj.extracted_from.id
 
 class NestedSampleSerializer(serializers.ModelSerializer):
```
UlysseFG, line 229: This should be replaced by a FullSampleSerializer version.

backend/fms_core/viewsets/sample.py
```diff
@@ -90,68 +75,13 @@ def get_serializer_class(self):
     def get_renderer_context(self):
         context = super().get_renderer_context()
         if self.action == 'list_export':
-            fields = SampleExportSerializer.Meta.fields
+            fields = SampleSerializer.Meta.fields
```
UlysseFG, line 78: This is something that is directly tied to the Export Serializer it should not be moved to a normal serializer.

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,184 @@
...
+        database.
+        """
+        return Response({"total_count": Sample.objects.all().count()})
+
+    @action(detail=False, methods=["get"])
+    def list_collection_sites(self, _request):
+        samples_data = Biosample.objects.filter().distinct("collection_site")
+        collection_sites = [s.collection_site for s in samples_data]
+        return Response(collection_sites)
+
```
UlysseFG, line 168: Missing get serializer class to provide the FullSampleNestedSerializer if requested... (it's not used inside Freezeman but it is part of the "reporting" endpoint.") People currently use this endpoint from outside the application.

backend/fms_core/migrations/0027_v3_5_0.py
```diff
@@ -0,0 +1,70 @@
...
+                ('coordinates', models.CharField(blank=True, help_text='Coordinates of the sample in a parent container. Only applicable for containers that directly store samples with coordinates, e.g. plates.', max_length=10)),
+                ('container', models.ForeignKey("Container", on_delete=models.DO_NOTHING, help_text="Container associated to the sample.")),
+                ('individual', models.ForeignKey("Individual", blank=True, null=True, on_delete=models.DO_NOTHING, help_text="Individual associated with to sample.")),
+                ('sample_kind', models.ForeignKey("SampleKind", on_delete=models.DO_NOTHING, help_text="Sample Kind associated to the sample.")),
+                ('derived_sample', models.ForeignKey("DerivedSample", on_delete=models.DO_NOTHING, help_text="Sample from which the current sample was extracted from.")),
+                ('biosample', models.ForeignKey("Biosample", on_delete=models.DO_NOTHING, help_text="Root Sample")),
+                ('child_of', ArrayField(models.CharField(blank=True, max_length=20), help_text="Parent Sample")),
```
UlysseFG, line 36: Is there a reason to store IDs as char arrays ?

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,235 @@
...
+            raise ValidationError(err)
+
+        #Save the new sample
+        try:
+            sample_to_update.save()
+        except Exception as err:
+            raise ValidationError(err)
+
+        #Return new sample
+        serializer = SampleSerializer(sample_to_update)
```
UlysseFG, line 187: return fullsample using sample_id and derived_id to identify uniquely your full sample.

sebastianballesteros: The sample_id is returned simply as id and derived_id is returned as an array of ids derived_ids. Do we need to return them differently?

backend/fms_core/migrations/0027_v3_5_0.py
sebastianballesteros: Not necessarily, we could store them as an array of integers

backend/fms_core/viewsets/full_sample.py
UlysseFG: I mean the object you return. You are right that there might be multiple derived sample id for a single full_sample update. I just would not want a sample stored in the frontend following an update. The fullsample need to be returned. When you submit your request you have the required info to get the fullsample out.

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,235 @@
...
+
+            DerivedBySample.objects.create(derived_sample_id=derived_sample.id,
+                                           sample_id=sample.id,
+                                           volume_ratio=1)
+
+        except ValidationError as err:
+            raise ValidationError(err)
+
+        #Serialize created sample
+        serializer = SampleSerializer(sample)
```
UlysseFG, line 142: you should get the full sample and serialize it instead of patching in the missing fields. It will break as soon as we add a field or remove one.

backend/fms_core/viewsets/full_sample.py
```diff
@@ -0,0 +1,231 @@
...
+        """
+        Searches for samples that match the given query
+        """
+        search_input = _request.GET.get("q")
+
+        query = Q(name__icontains=search_input)
+        query.add(Q(alias__icontains=search_input), Q.OR)
+        query.add(Q(id__icontains=search_input), Q.OR)
+
+        samples_data = Sample.objects.filter(query)
```
UlysseFG, line 228: Should return a full_sample too...

PR #172, opened 2021-11-08
Changes to models related to the split of sample.

UlysseFG:
Test on imports still fail but should be fixed in parent branch.


backend/fms_core/models/container.py
```diff
@@ -17,7 +17,6 @@
 from ._utils import add_error as _add_error
```
jleblond, line 17: Are those changes necessary? I think it would be easier to have it done later if it's simple string formatting

backend/fms_core/models/derived_sample.py
```diff
@@ -96,21 +111,56 @@ class DerivedSample(TrackedModel):
...
                                                     "It is the way to designate a subgroup within a study.")
-
     tissue_source = models.CharField(max_length=200, blank=True, choices=TISSUE_SOURCE_CHOICES,
                                      help_text="Can only be specified if the biospecimen type is DNA or RNA.")
-
     library = models.CharField(max_length=200, blank=True, null=True)
-
     index = models.CharField(max_length=200, blank=True, null=True)
 
+    def extracted_from(self): # returns a tuple of samples (extracted, extracted_from)
```
jleblond, line 124: Could you turn this into a '@property'? 

Also Im not sure about the logic. You are using an iterable collection? So if I want to get the extracted_from from a sample, I would get the first and only sample.extracted_from, if it's a pool I get the first item of an iterable collection?

backend/fms_core/models/project.py
```diff
@@ -14,28 +14,16 @@
```
jleblond, line 14: Same. Are those necessary in this PR (different attributes on different lines)? I'd recommend keeping the changes to minimal because we are going to end up with different PRs to merge into the split-sample PR, which will then be merged into master

backend/fms_core/models/sample.py
```diff
@@ -159,35 +54,9 @@ class Meta:
...
-
     @property
-    def individual_mother(self) -> Optional["Individual"]:
-        return self.individual.mother if self.individual else None
-
-    @property
-    def individual_father(self) -> Optional["Individual"]:
-        return self.individual.father if self.individual else None
+    def is_pool(self) -> bool:
+        return self.derived_samples.objects.count() > 1
```
jleblond, line 59: Would the self.derived_samples association work directly?
You could simply check if there are many DerivedBySample with this self (sample).id

backend/fms_core/models/sample.py
```diff
@@ -264,65 +127,27 @@ def add_error(field: str, error: str):
...
-                add_error("concentration", "Concentration must be specified if the sample_kind is DNA")
-
-            # Check tissue source given extracted_from
-
-            if self.tissue_source and self.sample_kind.name not in Sample.BIOSPECIMEN_TYPES_NA:
-                add_error("tissue_source", "Tissue source can only be specified for a nucleic acid sample.")
-
-            if self.transferred_from and self.sample_kind != self.transferred_from.sample_kind:
-                    add_error("sample_kind", "Sample kind need to remain the same during transfer")
+        DerivedSample = apps.get_model("fms_core", "DerivedSample")
```
jleblond, line 130: Why use the apps.get_model when you're outside of a migration here?
For now we used to import the other models in an import statement (from fms_core.models import DerivedSample).

Might not cause such a problem, but the apps.get_model looks up directly what is in the db, it does not rely on Django models

backend/fms_core/models/sample.py
```diff
@@ -264,65 +127,27 @@ def add_error(field: str, error: str):
...
+        # Check concentration fields given sample_kind
+        if self.concentration is None and (self.is_pool or self.derived_samples.objects.first().sample_kind.name in DerivedSample.BIOSPECIMEN_TYPES_CONC_REQUIRED):
```
jleblond, line 132: self.derived_samples.objects.first() ? 

You want to get the only one DerivedSample of a sample? Rather than calling this manually there should be a property that gets you the wanted DerivedSample. We should avoid those ... objects.first() directly in the code

backend/fms_core/models/sample.py
```diff
@@ -264,65 +127,27 @@ def add_error(field: str, error: str):
...
-
-            original_sample_kind = self.extracted_from.sample_kind.name
-            if self.tissue_source != Sample.BIOSPECIMEN_TYPE_TO_TISSUE_SOURCE[original_sample_kind]:
-                add_error(
-                    "tissue_source",
-                    (f"Mismatch between sample tissue source {self.tissue_source} and original sample kind "
-                     f"{original_sample_kind}")
-                )
+        if self.transferred_from:
+            if list(self.derived_samples.objects.all()) != list(self.transferred_from.derived_samples.objects.all()):
```
jleblond, line 136: I'd suggest comparing an unordered (or sorted) list of ids. Not sure that you can compare a django queryset transformed to a python list that easily. Would the order matter? You would be comparing a list of python objects I think

backend/fms_core/models/sample_lineage.py
```diff
@@ -27,23 +25,27 @@ def add_error(field: str, error: str):
 
         protocol_name = self.process_measurement.process.protocol.name
         if protocol_name == 'Extraction':
-            Sample = apps.get_model("fms_core", "Sample")
-            if self.child.sample_kind.name not in Sample.BIOSPECIMEN_TYPES_NA:
+            DerivedSample = apps.get_model("fms_core", "DerivedSample")
```
jleblond, line 28: same comment, try to import the models with import and not with the apps.get_model

backend/fms_core/models/sample_lineage.py
```diff
@@ -27,23 +25,27 @@ def add_error(field: str, error: str):
...
+            # There is only a single expected derived sample for the child and parent of an extraction
+            try:
+                child_derived = self.child.derived_samples.objects.get()
+                parent_derived = self.parent.derived_samples.objects.get()
```
jleblond, line 32: Doesnt get() take attributes-args?

jleblond (commented):
A few comments here and there. 

Also, in general, I think we should avoid making unnecessary changes (such as those in the models) in those types of large PRs which will get merged many times

backend/fms_core/models/derived_sample.py
UlysseFG: you would get (None, None) since pools are not a byproduct of an extraction...

backend/fms_core/models/sample.py
UlysseFG: This is less simple but maybe a bit more performant ... i'll change it and comment...

backend/fms_core/models/sample.py
UlysseFG: You mean it may affect performance ... Ok i'll import it ;).

backend/fms_core/models/sample.py
UlysseFG: We'll have to test it but i assume the list will be ordered by id since i think by default queryset orders using the primary key.

backend/fms_core/models/sample_lineage.py
UlysseFG: ok

backend/fms_core/models/sample_lineage.py
UlysseFG: Here again this is context specific ... Extraction should never be from a pool or to a pool ... The derived samples in this case are expected to have a single value ... I test the alternate case in the except which should raise an error ... something is not right in this situation.

backend/fms_core/models/container.py
UlysseFG: They are not necessary but they are done ... i'll deal with the pain of the merge ... I don't see me gaining anything by reverting those changes ... I kind of thought about it once it was done and i thought as you did but in the end it will clean the code a bit and i was going through the code line by line which does not happen often... Anyway ... unless you think these are terrible edits i'll let them stand ... You can point out if some of them you don't agree with but i do not think i want to roll all of them back.

backend/fms_core/models/sample.py
jleblond: It might work, but in all cases converting a queryset to a python list, and comparing the two lists that are filled with objects... i am not sure about the soundness of that solution. at least there is probably a function or an operator for list comparison in python

backend/fms_core/models/sample_lineage.py
jleblond: Ok, assuming that the extraction from pool would raise there an error, that is fine

backend/fms_core/models/sample.py
jleblond: hehe I disagree, this is better abstracted into a single function, you could simply raise an exception if there are more than 1 derived_samples for a sample.

What about we create a custom exception type for operations attempted on pools, that should not be done on a pool (such as extraction for example). That'd help us with the coding

backend/fms_core/models/derived_sample.py
jleblond: referring to my other comments I just posted (below), we could possibly create a type of custom exception for operations wrongly attempted

The return value (None, None) does not point to the problem

backend/fms_core/models/container.py
jleblond: that's ok for now since its done, I mean, it's just that you will probably merge your branch into the base branch before me and Sebastian does it, so we will have to deal with the merges from our branches then

UlysseFG: yeah i'll merge after you guys ...

jleblond: I need your part to have it merged before mine so the models can work :P thats ok, I will solve the conflicts on my branch after

backend/fms_core/models/derived_sample.py
UlysseFG: There is no problem here as far as i can see (at least not in this case) since it is a function to return the extracted_from which it does (and a bit more for practical reasons). This operation is a case where we do not have to worry ... i want the extracted_from a sample its either something or nothing (if the sample is not produced through extraction).

backend/fms_core/models/sample.py
UlysseFG: I would like it in a function property but i would be useless as far as i can see. The idea of an exception (validationError) for wrong operation is not too bad but the issue here is that mostly these are not bad operation for Pools they are just workaround that are specific to some cases. It is ok to get the first derivedsample in a pool ... it is wrong to think it is the only one though. The logic i did here assume it is not a pool but i would have to cover the case of the pool too if it was not an extraction (for a transfer for exemple). Give me an example maybe i do not get your idea.

backend/fms_core/models/sample.py
UlysseFG: yeah i am uncertain too but i would like to try it ... a list of objects should be easily enough comparable to another list of object. Each object should be equal if their primary key is identical ... it is a much clearer simpler way to to this than to build 2 list of ids in manually then comparing them ... i think in the end it's likely doing the same...

backend/fms_core/models/sample.py
jleblond: yes if I understand it correctly you mention them being 'workaround that are specific to some cases'. the workaround is using 'first()' to denote fetching the only one derived_sample attached to a sample if the sample is not a pool. I prefer the idea you suggested of having the derived_sample_when_not_pool function that tests that it is indeed not a pool, but of course we can have a shorter/more concise function (should be a property imo) name

UlysseFG: ok i'll make a shortcut property for that case ... we'll see if it's used more than once ;).

PR #173, opened 2021-11-08
Split importer service

jleblond:


backend/fms_core/services/derived_sample.py
```diff
@@ -0,0 +1,43 @@
+from django.core.exceptions import ValidationError
+from fms_core.models import Biosample, DerivedSample, DerivedBySample, Sample, Container, Process
+
+def get_single_derived_sample_from_sample(sample):
```
UlysseFG, line 4: ugh une fonction de 10 ligne pour remplacer ".first()" ?

jleblond: could be the function you create in the model. for now, I had to repeat that logic more than 5 times in the code..

backend/fms_core/services/derived_sample.py
```diff
@@ -0,0 +1,43 @@
...
+    ds = DerivedSample.objects.filter(derived_by_sample__sample_id=sample.id)
+    if ds.count() > 1:
+        errors.append(f"")
+    else:
+        derived_sample = ds.first()
+
+    return (derived_sample, errors, warnings)
+
+
+def derive_sample_from_sample(sample, new_sample_data, new_derived_sample_data):
```
UlysseFG, line 20: This would need some work to be compatible with pools ... This will only be useful for extraction as far as i see. instead of getting a single derived_sample from the sample (which is not true in most cases) we could make it copy of all derived_sample and connect them to a new sample that is returned (we could have an array of derived_samples returned too). Also the name is confusing i think we should not use derive in it...

backend/fms_core/services/derived_sample.py
UlysseFG: you still have to copy the logic outside unless you want this to crash all around the place ; ). the only original code i see in here is the ".first()"

jleblond: that is if you have a direct link from sample to derived_sample, which is not the case on my branch

UlysseFG: mmm yeah you are right about that ... well we'll see for another review down the line if it is still required but yeah you are correct about the immediate use ...

backend/fms_core/services/derived_sample.py
jleblond: Indeed this is not intended to handle the case of pools. We could handle this later on in the second part of the release since nowhere in the code the pools are handled for now.

Also, what name do you suggest for this function? Basically, the function creates a new derived_sample, a new sample, and the proper derived_by_sample from a sample that is neither a pool

UlysseFG: yeah its pretty much where derived sample name come from ... maybe just Derive_sample() you take a sample and you derive it into a new sample (and substructures)...

PR #175, opened 2021-11-15
MGI T7 Experiment

jleblond:


backend/fms_core/migrations/0027_v3_5_0.py
```diff
@@ -0,0 +1,127 @@
...
+        # Create PropertyType and Protocols
+        PROPERTY_TYPES_BY_PROTOCOL = {
+            "DNBSEQ Preparation": [("Experiment Name", "str"),
+                                   ("Flowcell Lot", "str"),
+                                   ("Loading Method", "str"),
+                                   ("Sequencer Side", "str"),
+                                   ("Sequencer Kit Used", "str"),
+                                   ("Sequencer Kit Lot", "str"),
+                                   ("Load DNB Cartridge Lot", "str"),
+                                   ("Primer Kit", "str"),
```
sebastianballesteros, line 53: Not sure if you are missing "Primer Kit Lot" from the Template or there's a reason why we don't include it.

PR #181, opened 2021-11-26
WIP: Implementing sample QC. 

sebastianballesteros:


backend/fms_core/migrations/0028_v3_5_0.py
```diff
@@ -0,0 +1,96 @@
...
+                                               type=type,
+                                               created_by_id=admin_user_id,
+                                               updated_by_id=admin_user_id)
+
+            reversion.add_to_revision(it)
+
+        # Create PropertyType and Protocols
+        # TODO: Maybe the dictionary is not necessary since there's only 1 protocol
+        PROPERTY_TYPES_BY_PROTOCOL = {
+            "Sample Quality Control": [("RIN (only for RNA)", "str"),
```
UlysseFG, line 50: We don't need to match the column name here unless the name is good enough. The "(only for RNA)" is not really helpful here.

sebastianballesteros: Okidoki

backend/fms_core/models/derived_sample.py
```diff
@@ -120,6 +120,9 @@ class DerivedSample(TrackedModel):
     library = models.CharField(max_length=200, blank=True, null=True)
     index = models.CharField(max_length=200, blank=True, null=True)
 
+    quality_flag = models.CharField(choices=[('Passed', 'Passed'), ('Failed', 'Failed')], help_text='Quality flag of the sample.', max_length=20)
```
UlysseFG, line 123: This would be better represented by a boolean i think (True, "Passed"), (False, "Failed"). Make sure it's Nullable to be able to tell the difference from an unset flag and a false flag.

backend/fms_core/models/derived_sample.py
```diff
@@ -120,6 +120,9 @@ class DerivedSample(TrackedModel):
...
+    quantity_flag = models.CharField(choices=[('Passed', 'Passed'), ('Failed', 'Failed')], help_text='Quantity flag of the sample.', max_length=20)
```
UlysseFG, line 124: Same as the other flag.

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def initialize_data_for_template(self):
+        self.preloaded_data = {'process': None, 'process_properties': {}}
+
+        self.preloaded_data['process'] = Process.objects.create(protocol=Protocol.objects.get(name="Sample Quality Control"),
+                                                                comment="Samples' Quality Control (imported from template)")
+
+        # Preload PropertyType objects for the sample qc in a dictionary for faster access
+        try:
+            protocol = Protocol.objects.get(name="Sample Quality Control")
```
UlysseFG, line 34: Would make more sense to get the protocol first then use it to set the preloaded_data['process']...

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -0,0 +1,36 @@
...
+                    process=process_measurement['process'],
+                    source_sample=sample,
+                    execution_date=process_measurement['execution_date'],
+                    volume_used=process_measurement['volume_used'],
+                    comment=process_measurement['comment'],
+                )
+
+            #TODO: Validate instruments according to platform
+
+            _, self.errors['properties'], properties_warnings = create_process_properties(process_properties,
```
UlysseFG, line 35: Do not forget that properties will be tied to the process_measurement instead of the process as in the experiment_run importer...

backend/fms_core/services/derived_sample.py
```diff
@@ -14,4 +14,23 @@ def inherit_derived_sample(derived_sample_source, new_derived_sample_data):
...
-    return (new_derived_sample, errors, warnings)
\ No newline at end of file
+    return (new_derived_sample, errors, warnings)
+
+def update_qc_flags(sample, quantity_flag, quality_flag):
+    errors = []
+    warnings = []
+
+    try:
+        derived_sample = sample.derived_sample_not_pool
```
UlysseFG, line 24: This is true so far but will it still be true for library QC ... If you have a service for updating the QC flags you must expect to reuse it ... Maybe you should move the service to sample.py since you pass a sample along ... or to full_sample.py...
I see that function as a facilitator that will set all derived samples attached to a sample to the values given...

backend/fms_core/services/derived_sample.py
```diff
@@ -14,4 +14,23 @@ def inherit_derived_sample(derived_sample_source, new_derived_sample_data):
...
+        if derived_sample:
+            if quantity_flag and quality_flag:
+                derived_sample.quantity_flag = True if quantity_flag == 'Passed' else False
```
UlysseFG, line 27: you could also used the simplified : derived_sample.quantity_flag = (quantity_flag == 'Passed').

backend/fms_core/services/derived_sample.py
```diff
@@ -14,4 +14,23 @@ def inherit_derived_sample(derived_sample_source, new_derived_sample_data):
...
+                derived_sample.quality_flag = True if quality_flag == 'Passed' else False
+                derived_sample.save()
+            else:
+                errors['flags'] = 'Quantity and Quality flags are required'
```
UlysseFG, line 31: Let's establish a standard that error messages and help text will end with a dot (.).

backend/fms_core/services/property_value.py
```diff
@@ -23,4 +23,30 @@ def create_process_properties(properties, processes_by_protocol_id):
         except ValidationError as e:
             errors.append(';'.join(e.messages))
 
+    return (property_values, errors, warnings)
+
+def create_process_measurement_properties(properties, process_measurement):
```
UlysseFG, line 28: I would add a layer of security for us that make sure that each property_type is related to the protocol to which is linked the process_measurement.

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -0,0 +1,85 @@
...
+
+        #Preload data
+        self.preloaded_data = {'process': None, 'process_properties': {}}
+
+        self.preloaded_data['process'] = Process.objects.create(protocol=protocol,
+                                                                comment='Sample Quality Control (imported from template)')
+
+        # Preload PropertyType objects for the sample qc in a dictionary for faster access
+        try:
+            self.preloaded_data['process_properties'] = { property.name: {'property_type_obj': property } for property in
```
UlysseFG, line 35: i would change process_properties to process_measurement_properties to make it clear those are not tied to a process.

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -0,0 +1,85 @@
...
+                'initial_volume': float_to_decimal_and_none(row_data['Initial Volume (uL)']),
+                'measured_volume': float_to_decimal_and_none(row_data['Measured Volume (uL)']),
+                'concentration': float_to_decimal_and_none(row_data['Concentration (ng/uL)']),
+                'quantity_flag': row_data['Quantity Flag'],
+                'quality_flag': row_data['Quality Flag']
+            }
+
+            process_measurement = {
+                'process': self.preloaded_data['process'],
+                'execution_date': row_data['QC Date'],
```
UlysseFG, line 68: Use the _util function to make sure the Date format is correct.

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def import_template_inner(self):
+        sample_qc_sheet = self.sheets['SampleQC']
+
+        for row_id, row_data in enumerate(sample_qc_sheet.rows):
+            process_properties = self.preloaded_data['process_properties']
+
+            #Populate process properties
+            for i, (key, val) in enumerate(row_data.items()):
+                #Since property names in db don't necessarily match those in the headers
```
UlysseFG, line 48: This is not a generic approach here. we have specific template fields mapped to each property. You can hardcode the mapping. Its safer that substring matching ... you are just hiding the harcoding this way (which is worse).

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -0,0 +1,78 @@
...
+            coordinates=sample['coordinates'],
+        )
+
+        if sample_obj:
+            # Update sample with sample_information
+            new_volume = None
+            # TODO: Maybe validate initial volume?
+            if all([sample_information['initial_volume'], sample_information['measured_volume'],
+                    process_measurement['volume_used']]):
+                delta_volume = sample_information['initial_volume'] - sample_information['measured_volume']
```
UlysseFG, line 36: I would reverse the 2 values... i would expect if the measured_volume is superior to the initial _volume to have a positive Delta_volume ...

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -0,0 +1,78 @@
...
+                new_volume = sample_obj.volume - delta_volume - process_measurement['volume_used']
```
UlysseFG, line 37: Add instead of Subtract the delta_volume.

backend/fms_core/migrations/0028_v3_5_0.py
```diff
@@ -0,0 +1,96 @@
...
+            platform = Platform.objects.get(name=INSTRUMENT_TYPES[type])
+            it = InstrumentType.objects.create(platform=platform,
+                                               type=type,
+                                               created_by_id=admin_user_id,
+                                               updated_by_id=admin_user_id)
+
+            reversion.add_to_revision(it)
+
+        # Create PropertyType and Protocols
+        # TODO: Maybe the dictionary is not necessary since there's only 1 protocol
```
UlysseFG, line 48: You can remove this comment. A dict is not necessary, but it is not wrong either.

backend/fms_core/migrations/0028_v3_5_0.py
```diff
@@ -0,0 +1,96 @@
...
+        }
+        protocol_content_type = ContentType.objects.get_for_model(Protocol)
+
+        for protocol_name in PROPERTY_TYPES_BY_PROTOCOL.keys():
+            protocol = Protocol.objects.create(name=protocol_name,
+                                               created_by_id=admin_user_id, updated_by_id=admin_user_id)
+            reversion.add_to_revision(protocol)
+
+            for (property, value_type) in PROPERTY_TYPES_BY_PROTOCOL[protocol_name]:
+                is_optional = False if 'instrument' in property.lower() else True
```
UlysseFG, line 64: do not set stuff using string match ... it is much safer and clearer to set them directly. Add an if with the specific name of both instrument types and set them to false directly.

backend/fms_core/services/property_value.py
```diff
@@ -23,4 +23,36 @@ def create_process_properties(properties, processes_by_protocol_id):
...
+    for value_dict in properties.values():
+        property_type = value_dict['property_type_obj']
+        value = value_dict['value']
+
+        # Validate relation between the property type and the correct protocol
+        if property_type.object_id != process_measurement.process.protocol.id:
+            errors.append(f'{property_type} is not linked to protocol {process_measurement.process.protocol.id}.')
+
+        # Validate non-optional properties
+        if not property_type.is_optional and not value:
```
UlysseFG, line 43: This should not be validated at this level (service), but possibly at the row_handler level. We may for some reason not set all the properties at once ... Also we could simply not pass a property along to bypass this validation altogether. Maybe create another service that is called after the properties are created and that will validate if they are have values when not optional.

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -7,29 +7,46 @@ const {Title} = Typography;
...
+    const propertiesAreLoaded = processMeasurement?.properties?.every(property => property in propertyValuesByID)
 
     if (!isLoaded)
         get(id);
 
+    if (isLoaded && !propertiesAreLoaded) {
+      listPropertyValues({ object_id__in: processMeasurement.id, content_type__model: "processmeasurement" });
+    }
+
+    console.log(propertyValuesByID)
```
UlysseFG, line 48: remove the log when you are done testing.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -0,0 +1,80 @@
...
+                        it = InstrumentType.objects.get(type=type)
+                        # Validate platform and type
+                        if it.platform.name != TYPES_BY_PLATFORM[it.type]:
+                            self.errors['instrument type'] = f'Invalid type: {it.platform} for instrument: {it.type}.'
+                    except Exception as e:
+                        self.errors['instrument'] = f'Invalid instrument {type}.'
+
+                # Validate required RIN for RNA
+                if sample_obj.derived_sample_not_pool.sample_kind.name == 'RNA' and process_measurement_properties['RIN']['value'] is None:
+                    self.errors['RIN'] = 'It has to be specified for RNA.'
```
UlysseFG, line 76: The message will be very mysterious ... "It" should be more clearly stated as RIN.

backend/fms_core/viewsets/process_measurement.py
```diff
@@ -42,6 +42,12 @@ class ProcessMeasurementViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
             "template": SAMPLE_TRANSFER_TEMPLATE,
             "importer": TransferImporter,
         },
+        {
+            "name": "Sample Quality Control",
+            "description": "Upload the provided template with samples that underwent a quality control.",
+            "template": SAMPLE_QC_TEMPLATE,
+            "importer": SampleQCImporter,
```
UlysseFG, line 49: This format will crash with the new way to forward the template to the frontend. You need to have the template in a list of objects. You can compare to the other templates once you rebase.

backend/fms_core/viewsets/process_measurement.py
```diff
@@ -42,6 +42,12 @@ class ProcessMeasurementViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
             "template": SAMPLE_TRANSFER_TEMPLATE,
             "importer": TransferImporter,
         },
+        {
+            "name": "Sample Quality Control",
+            "description": "Upload the provided template with samples that underwent a quality control.",
+            "template": SAMPLE_QC_TEMPLATE,
```
UlysseFG, line 48: This format will crash with the new way to forward the template to the frontend. You need to have the template in a list of objects. You can compare to the other templates once you rebase.

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -0,0 +1,92 @@
...
+from .._utils import float_to_decimal_and_none, input_to_date_and_none
+
+# {{TEMPLATE PROPERTY NAME : DB PROPERTY NAME}
+TEMPLATE_PROPERTY_MAPPING = {
+    "Measured Volume (uL)": "Measured Volume",
+    "Concentration (ng/uL)": "Concentration",
+    "RIN (for RNA only)": "RIN",
+    "Electrophoresis Instrument": "Electrophoresis Instrument",
+    "Quantitation Instrument": "Quantitation Instrument",
+    "Comment": "Comment"
```
UlysseFG, line 15: Looking at the Process_Measurement detail page made me realize this property is not useful. Its a duplicate of the one stored directly on the process_measurement. Remove it from the properties. We'll store Comments for the QC directly on the process_measurement.

backend/fms_core/viewsets/process_measurement.py
```diff
@@ -42,6 +42,12 @@ class ProcessMeasurementViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
             "template": [{"description": "Template to transfer samples","file": SAMPLE_TRANSFER_TEMPLATE}],
             "importer": TransferImporter,
         },
+        {
```
UlysseFG, line 45: Move this template action to the sample ... will be more logical to have sample QC on sample and library QC on the library page.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -178,6 +178,15 @@ const SampleDetailsContent = ({samplesByID, sampleKindsByID, containersByID, pro
               </Descriptions.Item>
               <Descriptions.Item label="Coordinates">{sample.coordinates || "—"}</Descriptions.Item>
               <Descriptions.Item label="Comment" span={3}>{sample.comment}</Descriptions.Item>
+              { // For flags there's a difference between having a false value (QC failed)
+                // and a null value (QC not performed yet)
+              }
+              <Descriptions.Item label="Quantity Flag">
+                { sample.quantity_flag ? "Passed" : (sample.quantity_flag ===  false ? "Failed" : "-") }
```
UlysseFG, line 185: Use em dash instead of en dash as in Concentration if you want to put a default.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -178,6 +178,15 @@ const SampleDetailsContent = ({samplesByID, sampleKindsByID, containersByID, pro
...
+              </Descriptions.Item>
+              <Descriptions.Item label="Quality Flag">
+                { sample.quality_flag ? "Passed" : (sample.quality_flag === false ? "Failed" : "-")}
```
UlysseFG, line 188: Use em dash instead of en dash as in Concentration if you want to put a default.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -178,6 +178,15 @@ const SampleDetailsContent = ({samplesByID, sampleKindsByID, containersByID, pro
               </Descriptions.Item>
               <Descriptions.Item label="Coordinates">{sample.coordinates || "—"}</Descriptions.Item>
               <Descriptions.Item label="Comment" span={3}>{sample.comment}</Descriptions.Item>
+              { // For flags there's a difference between having a false value (QC failed)
+                // and a null value (QC not performed yet)
+              }
+              <Descriptions.Item label="Quantity Flag">
```
UlysseFG, line 184: The flag are not visible from the list. We should make it visible somehow. I think 2 column for QC flags is too much ... I think we could combine them as QC flag on the list. Maybe we could have a mouse over details popup that give the quality and quantity flags separately.

UlysseFG, line 184: Also maybe create a component akin <SampleDepletion depleted=BOOL /> but that offer more options to be able to reuse it for the library QC and the list. I could see something like <QcFlag passed=[{key: BOOL}, {key: BOOL},...], tooltip=BOOL /> this would combine a number of flags into a single icon with or without tooltip giving the details by key.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -0,0 +1,83 @@
...
+
+                # Validate required RIN for RNA
+                if sample_obj.derived_sample_not_pool.sample_kind.name == 'RNA' and process_measurement_properties['RIN']['value'] is None:
+                    self.errors['RIN'] = 'RIN has to be specified for RNA.'
+
+                _, self.errors['properties'], self.warnings['properties'] = create_process_measurement_properties(
+                    process_measurement_properties,
+                    process_measurement_obj)
+
+                self.errors['non_optional_properties'], self.warnings['non_optional_properties'] = \
```
UlysseFG, line 82: I would validate before creating the properties ... No point in creating them if some required properties are missing.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -178,6 +178,15 @@ const SampleDetailsContent = ({samplesByID, sampleKindsByID, containersByID, pro
...
+                { sample.quantity_flag ? "Passed" : (sample.quantity_flag ===  false ? "Failed" : "—") }
```
UlysseFG, line 185: I would use the QC flag here too.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -178,6 +178,15 @@ const SampleDetailsContent = ({samplesByID, sampleKindsByID, containersByID, pro
...
+              </Descriptions.Item>
+              <Descriptions.Item label="Quality Flag">
+                { sample.quality_flag ? "Passed" : (sample.quality_flag === false ? "Failed" : "—")}
```
UlysseFG, line 188: I would use the QC flag here too.

frontend/src/components/samples/SamplesListContent.js
```diff
@@ -115,6 +116,18 @@ const getTableColumns = (containersByID, individualsByID, projectsByID, sampleKi
...
       width: 115,
     },
+    {
+      title: "QC Flag",
+      dataIndex: "qc_flag",
+      sorter: true,
+      render: (_, sample) => {
+        const flags = { quantity: sample.quantity_flag, quality: sample.quality_flag };
+        if (flags.quantity !== null && flags.quality !== null)
+          return <SampleQCFlag flags={flags}/>;
```
UlysseFG, line 126: The QC flag component should be usable for the library QC flags too. I would name it QCFlag simply for that reason.

frontend/src/components/samples/SampleQCFlag.js
```diff
@@ -0,0 +1,23 @@
...
+      <p> Quantity Flag: {flags.quantity ? "Passed" : "Failed"}</p>
+      <p> Quality Flag: {flags.quality ? "Passed" : "Failed"}</p>
+    </div>
+  )
+}
+
+export const SampleQCFlag = ({flags}) => {
+  return (
+    <Popover content={getFlagContent(flags)} title="QC Flags">
+      {(flags.quantity && flags.quality)
```
UlysseFG, line 18: use every() with the list of values to make this more generic

frontend/src/components/samples/SampleQCFlag.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+import {Popover, Button} from "antd";
+import {CheckOutlined, CloseOutlined} from "@ant-design/icons";
+
+const getFlagContent = (flags) => {
+  return  (
+    <div>
+      <p> Quantity Flag: {flags.quantity ? "Passed" : "Failed"}</p>
```
UlysseFG, line 9: Use the keys and values instead of harcoding the names of the flags here in order to keep this generic

backend/fms_core/migrations/0028_v3_5_0.py
```diff
@@ -0,0 +1,90 @@
...
+        }
+        protocol_content_type = ContentType.objects.get_for_model(Protocol)
+
+        for protocol_name in PROPERTY_TYPES_BY_PROTOCOL.keys():
+            protocol = Protocol.objects.create(name=protocol_name,
+                                               created_by_id=admin_user_id, updated_by_id=admin_user_id)
+            reversion.add_to_revision(protocol)
+
+            for (property, value_type) in PROPERTY_TYPES_BY_PROTOCOL[protocol_name]:
+                if any([property == 'Quantitation Instrument', property == 'Electrophoresis Instrument']):
```
UlysseFG, line 65: Add "Measured Volume" and Concentration as required...

backend/fms_core/models/full_sample.py
```diff
@@ -80,6 +86,14 @@ class FullSample(models.Model):
     updated_by = models.ForeignKey(User, null=False, blank=True, related_name="%(app_label)s_%(class)s_modification", on_delete=models.DO_NOTHING)
     deleted = models.BooleanField(default=False, help_text="Whether this instance has been deleted.")
 
+    @property
+    def extracted_from(self) -> "Sample":
```
UlysseFG, line 90: Maybe just call sample.extracted_from() ? If sample.extracted_from is not working we need to fix it ...

frontend/src/components/samples/SampleQCFlag.js
sebastianballesteros: the values are booleans, not strings

PR #190, opened 2021-12-07
WIP: Fixed template issues

sebastianballesteros:


backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -85,6 +85,9 @@ def process_row_inner(self, sample, container, project, parent_container, indivi
                                          cohort=individual['cohort'],
                                          mother=mother_obj,
                                          father=father_obj)
+        else:
+            self.warnings['individual'] = []
```
UlysseFG, line 89: you can initialize the self.warnings['individual'] outside the scope of the if so it is not repeated. in each case of the if ...

sebastianballesteros: it is not repeated, one is warnings and the other is errors but I can initialize both out of the scope to have consistency

UlysseFG: Oh right yeah it would be cleaner i think.

PR #192, opened 2021-12-08
Alias none template

UlysseFG:


sebastianballesteros (approved):
Do we want to fix the sample submission template (validation) in this PR or in another?

PR #206, opened 2021-12-17
Remove sample view

sebastianballesteros:
Got rid of Sample View to improve sample list performance


backend/fms_core/models/sample.py
```diff
@@ -62,6 +64,10 @@ def is_pool(self) -> bool:
     def derived_sample_not_pool(self) -> DerivedSample:
         return self.derived_samples.first() if not self.is_pool else []  # Forces crash if pool
 
+    @property
+    def biosample(self) -> Biosample:
+        return self.derived_sample_not_pool.biosample if not self.is_pool else None
```
UlysseFG, line 69: Do not use derived_sample_not_pool inside generic properties, the way we deal with pool is likely to be context specific and wrapping this in here will prevent this.

backend/fms_core/serializers.py
```diff
@@ -294,13 +286,13 @@ def get_location_barcode(self, obj):
     def get_current_volume(self, obj):
         return obj.volume
 
-class FullNestedSampleSerializer(serializers.ModelSerializer):
+class NestedSampleSerializer(serializers.ModelSerializer):
```
UlysseFG, line 289: Its is likely missing most "full sample" fields...

backend/fms_core/serializers.py
```diff
@@ -256,13 +241,20 @@ def get_depleted(self, obj):
 
 class SampleSerializer(serializers.ModelSerializer):
     extracted_from = serializers.SerializerMethodField()
+    sample_kind = serializers.CharField(read_only=True, source="derived_sample_not_pool.sample_kind.id")
```
UlysseFG, line 244: We need to have a clear idea how we'll have this working once we have pool or we may very well have to redesign the whole thing again ... From our previous discussion i think the word is to hide from list all pool data that is not from sample so this might actually work long term. Lets think it over some.

backend/fms_core/models/sample.py
```diff
@@ -62,6 +64,10 @@ def is_pool(self) -> bool:
     def derived_sample_not_pool(self) -> DerivedSample:
         return self.derived_samples.first() if not self.is_pool else []  # Forces crash if pool
 
+    @property
+    def biosample(self) -> Biosample:
```
UlysseFG, line 68: Might want to rename it similar to derived_sample_not_pool in order to make sure we remember not to use this on pools.

UlysseFG (approved):

PR #210, opened 2022-01-05
WIP: Implemented extraction rollback curation

sebastianballesteros:


UlysseFG:
This is not required as far as i know the current system works. Close this PR once we confirm this is not required.

PR #212, opened 2022-01-06
Add experiment run comments 

sebastianballesteros:


backend/fms_core/services/sample.py
```diff
@@ -194,7 +195,8 @@ def transfer_sample(process: Process,
                 coordinates=coordinates_destination if coordinates_destination else "",
                 creation_date=execution_date,
                 volume=volume_destination if volume_destination else volume_used,
-                depleted=False
+                depleted=False,
+                comment=comment if comment else {},
```
UlysseFG, line 199: This will override the sample based comment instead of setting it on the process_measurement. I do not think we should do that. Also, you should not put a dictionnary ({}) in a string variable. Use None or "".

UlysseFG:
There is only a small issue, but reading the code to find which comment goes where i realized the comments from extraction and transfer template are never inserted in the process_measurement as they should be. We should fix this in the same PR since its the same issue basically.

sebastianballesteros:
Indeed, I looked again at the code and transfer/extraction importers don't insert the right comment to process_measurements, I think they do it in the Process level. I will fix this in this PR. 

However, the problem with overriding the sample base comment you mean the base comment f"Automatically generated on .....etc"? In addition, I am using a dictionary to emphasize it is an optional argument. Using "" will just create and empty comment and None will give problems with the model constraint since it cannot be null. What do you think about using **(dict(comment=comment) if comment is not None else dict()) would include the comment just if it is not None, if it is None then it would not even add it to the arguments.

UlysseFG:
This seem alright. One last thing can you change the templates file names to make it so they are named after the version they were last changed ? This is annoying but it helps for people to keep track if they have the most recent version of the template. (These template also include an internal version number that we do not use yet. This might be good to update that one too.).

PR #215, opened 2022-01-26
Fixed and modified mandatory property validation 

sebastianballesteros:
For experiment run and sample qc

### DIFF
--- a/backend/fms_core/services/process_measurement.py
+++ b/backend/fms_core/services/process_measurement.py
@@ -1,10 +1,11 @@
 from datetime import datetime
 from django.core.exceptions import ValidationError
 
+from fms_core.services.property_value import validate_non_optional_properties
 from fms_core.models import ProcessMeasurement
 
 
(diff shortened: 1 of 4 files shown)


backend/fms_core/services/process_measurement.py
```diff
@@ -1,10 +1,11 @@
 from datetime import datetime
 from django.core.exceptions import ValidationError
 
+from fms_core.services.property_value import validate_non_optional_properties
 from fms_core.models import ProcessMeasurement
 
 
-def create_process_measurement(process, source_sample, execution_date, volume_used=None, comment=None):
+def create_process_measurement(process, source_sample, execution_date, volume_used=None, comment=None, process_measurement_properties=None,):
```
UlysseFG, line 8: You don't use this ... It is not a bad idea but i think we don't need it. as long as the test is done inside both create_process_properties and process_measurement_properties the general result is the same. If you want to prevent unnecessary processing as soon as when creating a process_measurement it is great but you have to actually add the parameter when you call the function ... At the moment, it appear to be deadcode.

PR #218, opened 2022-01-27
Template importer for qPCR protocol

UlysseFG:


backend/fms_core/template_importer/row_handlers/sample_selection_qpcr/sample.py
```diff
@@ -18,6 +18,11 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
 
         if sample_obj:
             # Update sample if Depleted
+            depleted = None
+            if sample['depleted'] == 'YES':
+                depleted = True
+            elif sample['depleted'] == 'NO':
```
UlysseFG, line 24: No is not a valid option... We can accept only YES... Anything else is None.

PR #221, opened 2022-01-31
Implemented qPCR status filter by overriding the queryset

sebastianballesteros:


frontend/src/components/filters/descriptions.js
```diff
@@ -114,6 +114,20 @@ export const SAMPLE_FILTERS = {
...
     detached: true,
   },
+  qPCR_status: {
+    type: FILTER_TYPE.SELECT,
+    key: "qPCR_status",
+    label: "qPCR Selection Status",
+    placeholder: "All",
+    mode: "multiple",
+    options: [
+      { label: "Positive", value: "Positive" },
```
UlysseFG, line 124: Move these status to the constants.js file. Make sure they have a name that distinguish them from Project status.

frontend/src/constants.js
```diff
@@ -37,11 +37,15 @@ export const TAXON = [
   "Sars-Cov-2",
 ]
 
-export const STATUS = [
+export const PROJECT_STATUS = [
   "Open",
   "Closed",
 ]
 
+export const PROPERTY_NAMES = [
```
UlysseFG, line 45: i would add "SEARCHABLE_" to make it clearer that it applies to filters.

PR #222, opened 2022-02-01
Added CT Value 2

sebastianballesteros:
** I realized we were not using sample_information in the template importer


backend/fms_core/template_importer/row_handlers/sample_selection_qpcr/sample.py
```diff
@@ -12,7 +12,7 @@ class SampleSelectionQPCRRowHandler(GenericRowHandler):
     def __init__(self):
         super().__init__()
 
-    def process_row_inner(self, sample, sample_information, process_measurement, process_measurement_properties):
+    def process_row_inner(self, sample, process_measurement, process_measurement_properties):
```
UlysseFG, line 15: Not sure this is the right way to do this every process_row_inner has a sample_information. We expect it to have one so if we call the function and pass parameter by position there is a high risk of error. Just pass an empty object if you want. but leave the method signature the same.

backend/fms_core/template_importer/importers/sample_selection_qpcr.py
```diff
@@ -69,7 +64,6 @@ def import_template_inner(self):
 
             sample_selection_qpcr_kwargs = dict(
                 sample=sample,
-                sample_information=sample_information,
```
UlysseFG, line 72: keep the function parameter just pass it an empty dictionary

backend/fms_core/template_importer/importers/sample_selection_qpcr.py
```diff
@@ -53,12 +54,6 @@ def import_template_inner(self):
                 'container': {'barcode': row_data['Sample Container Barcode']},
                 'depleted': row_data['Source Depleted'],
             }
-            sample_information = {
-                'qpcr_type': row_data['qPCR Type'],
-                'ct_value_experimental': float_to_decimal_and_none(row_data['CT Value (Experimental)']),
```
UlysseFG, line 58: It may be a good idea to process the numeric values in the input (ct values) in order to normalize them. If we don't do that we may store number with 8 decimals positions or numbers without 0 before the decimal.

PR #224, opened 2022-02-02
backend basic approach to properties export in process measurement

UlysseFG:


sebastianballesteros (approved):

backend/fms_core/serializers.py
```diff
@@ -165,16 +166,62 @@ class ProcessMeasurementExportSerializer(serializers.ModelSerializer):
...
+        # Instantiate the superclass normally
+        super(ProcessMeasurementWithPropertiesExportSerializer, self).__init__(*args, **kwargs)
+        # List all property fields that are tied to the protocol
+        self.property_types = self.list_property_types(self.instance)
+        for property_type in self.property_types:
+            try:
+                self.fields[property_type.name] = serializers.CharField(read_only=True)
+                self.Meta.fields = self.Meta.fields + (property_type.name,)
+            except Exception as err:
+                print(err)
```
sebastianballesteros, line 194: Maybe there's a better way of handling the error than just printing it

UlysseFG: yeah this was to debug but i'll replace it with a message to the frontend...

PR #226, opened 2022-02-08
Added process detailed page with properties

sebastianballesteros:


frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -47,7 +47,7 @@ const ProcessMeasurementsDetailContent = ({
 
     const isLoading = !isLoaded || processMeasurement.isFetching;
     const title =
-        `Sample Process ${[id, processMeasurement && protocolsByID[processMeasurement.protocol]?.name ].filter(Boolean).join(' - ')}`;
+        `Sample Process Measurement ${[id, processMeasurement && protocolsByID[processMeasurement.protocol]?.name ].filter(Boolean).join(' - ')}`;
```
UlysseFG, line 50: I don't think we should expose process measurement to users... it has no meaning for them. I think maybe differentiate the title using Process and Sample Process (for process measurement).

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -0,0 +1,71 @@
...
+                <Descriptions.Item label="Protocol" span={4}>{protocolsByID[process.protocol]?.name}</Descriptions.Item>
+                {process.parent_process &&
+                  <Descriptions.Item label="Parent Process" span={4}>{process.parent_process}</Descriptions.Item>
+                }
+                <Descriptions.Item label="Comment" span={4}>{process.comment}</Descriptions.Item>
+            </Descriptions>
+            { process?.children_properties?.length > 0 &&
+                <>
+                <Title level={3} style={{marginTop: '20px'}}>Properties</Title>
+                  <ExperimentRunsProperties
```
UlysseFG, line 60: We will have to rename and move that component if it is use in a generic manner.

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -0,0 +1,71 @@
...
+      listProcesses({id__in: id});
+    }
+
+    if (isLoaded && !propertiesAreLoaded) {
+      listPropertyValues({ object_id__in: id, content_type__model: "process" });
+    }
+
+    const isLoading = !isLoaded || process.isFetching;
+    const title =
+        `Sample Process ${[id, process && protocolsByID[process.protocol]?.name ].filter(Boolean).join(' - ')}`;
```
UlysseFG, line 44: Use Process here only to differentiate with Sample Process.

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -0,0 +1,71 @@
...
+        <PageContent loading={isLoading}>
+            <Title level={2}>Overview</Title>
+            <Descriptions bordered={true} size="small" column={4}>
+                <Descriptions.Item label="Protocol" span={4}>{protocolsByID[process.protocol]?.name}</Descriptions.Item>
+                {process.parent_process &&
+                  <Descriptions.Item label="Parent Process" span={4}>{process.parent_process}</Descriptions.Item>
+                }
+                <Descriptions.Item label="Comment" span={4}>{process.comment}</Descriptions.Item>
+            </Descriptions>
+            { process?.children_properties?.length > 0 &&
```
UlysseFG, line 57: The only issue that remains i think is that experiments run process do not have their properties displayed since the properties are attached to child processes ... Not sure it is necessary since its somewhere else already but it might be expected to appear here also. I think it would require to add Children_processes to the process endpoint (not such a bad idea i think).

frontend/src/components/experimentRuns/ExperimentRunsDetailContent.js
```diff
@@ -69,6 +69,7 @@ const ExperimentRunsDetailContent = ({
   if (isLoaded && !processesByID[experimentRun.process]) {
     // Need to be queried as a string, not as an array in order to work with DRF filters
     const processIDSAsStr = [experimentRun.process].concat(experimentRun.children_processes).join()
+    console.log(processIDSAsStr)
```
UlysseFG, line 72: not needed ...

UlysseFG (approved):

PR #227, opened 2022-02-08
Implemented batch filter for project

sebastianballesteros:
In addition, I made the filter generic since many attributes are using it


UlysseFG (approved):
Well done

PR #228, opened 2022-02-10
WIP: Index Model

sebastianballesteros:


backend/fms_core/migrations/0031_v3_7_0.py
```diff
@@ -0,0 +1,143 @@
...
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='index',
+            name='index_set',
+            field=models.ForeignKey(help_text='The set which this index belongs to', on_delete=django.db.models.deletion.PROTECT, related_name='indeces', to='fms_core.indexset'),
+        ),
+        migrations.AddField(
```
UlysseFG, line 118: As much as possible, do not add model fields apart from model creation. Makes tracking if everything is present harder.

backend/fms_core/models/index_structure.py
```diff
@@ -0,0 +1,29 @@
...
+
+__all__ = ["IndexStructure"]
+
+@reversion.register()
+class IndexStructure(TrackedModel):
+    name = models.CharField(unique=True, max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
+                            help_text="The name of the index structure.")
+
+    flanker_3prime_forward = models.ForeignKey("Sequence", related_name="flanker_3prime_forward",
+                                               on_delete=models.PROTECT, help_text="Flanker on the 3' forward direction")
```
UlysseFG, line 18: using single quote in the code is a bit risky. Might be safer to spell it as "prime".

PR #232, opened 2022-02-11
Implemented index creation via template submission

sebastianballesteros:


backend/fms_core/services/index.py
```diff
@@ -0,0 +1,99 @@
...
+
+    if set_name:
+        try:
+            index_set = IndexSet.objects.get(name=set_name)
+
+        except IndexSet.DoesNotExist:
+            try:
+                index_set = IndexSet.objects.create(name=set_name)
+                created_entity = True
+            # Pile up all validation error raised during the creation of the container
```
UlysseFG, line 19: Copy pasted comment not applicable here...

backend/fms_core/services/index.py
```diff
@@ -0,0 +1,99 @@
...
+            errors.append(f"Invalid index structure.")
+            return index, errors, warnings
+
+        try:
+            index = Index.objects.create(index_set=index_set, name=index_name, index_structure=index_structure_obj)
+        except ValidationError as e:
+            errors.append(';'.join(e.messages))
+
+        return index, errors, warnings
+
```
UlysseFG, line 46: Too many empty lines.

backend/fms_core/services/index.py
```diff
@@ -0,0 +1,99 @@
...
+
+    else:
+        errors.append(f"Index structure is needed to create an index.")
```
UlysseFG, line 49: This else statement has no return ... this would have a default return of None but this function typically returns more than one value so it's likely to cause failure if the else is called.

backend/fms_core/services/index.py
```diff
@@ -0,0 +1,99 @@
...
+def create_indices_3prime_by_sequence(index, index_3prime):
+    indices_3prime_by_sequence = None
+    errors = []
+    warnings = []
+
+    if not index:
+        errors.append(f"Index is required.")
+        return indices_3prime_by_sequence, errors, warnings
+
+    for value in index_3prime.split(','):
```
UlysseFG, line 61: This should be done outside the service the service should process already digested data. The comma separated format we receive the list is related to a limitation of the template submission. It should be converter at the level of the importer when data is captured.

backend/fms_core/template_importer/row_handlers/index_creation/index.py
```diff
@@ -0,0 +1,37 @@
...
+                if any([index['index_3prime'], index['index_5prime']]):
+                    indices_3prime_by_sequence, self.errors['index_3prime'], self.warnings['index_3prime'] = \
+                        create_indices_3prime_by_sequence(index=index_obj,
+                                                          index_3prime=index['index_3prime'])
+
+                    indices_5prime_by_sequence, self.errors['index_5prime'], self.warnings['index_5prime'] = \
+                        create_indices_5prime_by_sequence(index=index_obj,
+                                                          index_5prime=index['index_5prime'])
+                else:
+                    self.errors['indices_prime'] = 'At least one index prime is required.'
```
UlysseFG, line 37: Not sure we should call this an index prime ... the prime part tells us the position on the sequence. I am not sure what we should call this ... maybe "At least one index sequence is required."

backend/fms_core/services/index.py
```diff
@@ -0,0 +1,98 @@
...
+
+        try:
+            index = Index.objects.create(index_set=index_set, name=index_name, index_structure=index_structure_obj)
+        except ValidationError as e:
+            errors.append(';'.join(e.messages))
+
+        return index, errors, warnings
+    else:
+        errors.append(f"Index structure is needed to create an index.")
+        return index, errors, warnings
```
UlysseFG, line 47: Mmm that fixes the issue but it would be cleaner to just take the return value out of the if else statement.

backend/fms_core/utils.py
```diff
@@ -30,6 +31,12 @@
 
 TRUTH_VALUES = frozenset({"TRUE", "T", "YES", "Y"})
 
+def comma_separated_string_to_array(s):
+    """
+    Returns emtpy string if argument is a blank string or None,
+    otherwise it returns an array of strings
+    """
+    return s.replace(" ", "").split(',') if s else ""
```
UlysseFG, line 39: this is dangerous ... that replace may not be anticipated when we call comma_separated_string_to_array. In some condition we may want to preserve the spaces. Don't you want strip() instead ? less risky. Also, the return value is either a string or an array ... if there is no element it would be better to return either an empty list or a list with an empty string as sole element.

sebastianballesteros: strip doesn't remove white spaces in the middle of the string so "ABC ABC" would still be "ABC ABC". The replace is basically removing whitespaces, which if you want an array of strings I believe normally you would want the items of the array to be stripped.

frontend/src/components/DashboardPage.js
```diff
@@ -89,6 +92,15 @@ const DashboardPage = ({
               )}
             </Row>
           </Card>
+          <div style={{ display: 'flex', marginBottom: '1em' }}></div>
```
UlysseFG, line 95: The card placement on the Dashboard is a bit weird. Maybe place it in the second column below the samples ?

frontend/src/components/filters/descriptions.js
```diff
@@ -359,3 +359,26 @@ export const PROJECT_FILTERS = {
...
+    type: FILTER_TYPE.INPUT,
+    key: "id",
+    label: "ID",
+  },
+  name: {
+    type: FILTER_TYPE.INPUT,
+    key: "name",
+    label: "Name",
+  },
+  index_structure: {
```
UlysseFG, line 374: Ordering by structure and set do not work. Might be the same issue as we've had with the last version with the filter name not matching the key...

UlysseFG (approved):

PR #238, opened 2022-02-22
Check if logged in before fetching user.

UlysseFG:


frontend/src/components/App.js
```diff
@@ -144,7 +144,7 @@ const App = ({userID, usersByID, logOut, fetchInitialData, fetchSummariesData, g
   const isLoggedIn = userID !== null;
   const user = usersByID[userID];
 
-  if (!user)
+  if (!user && isLoggedIn)
```
sebastianballesteros, line 147: isn't it kind of redundant since user will always be null (usersByID[null]) when userID === null?

UlysseFG: This covers the user not logged in. It otherwise pop an error each time because it tries to get(null).

sebastianballesteros: ok

PR #242, opened 2022-02-28
Add Sequences to frontend index list.

UlysseFG:


frontend/src/components/DropdownListItems.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    ListItems && ListItems.length > 1 ? 
+    <Dropdown overlay={menuListItems} placement="bottomRight">
+      <div>
+        <MenuOutlined /> {ListItems && ListItems[0]}
+      </div>
+    </Dropdown>
+    :
+    <div>
+      {ListItems ? <LineOutlined /> : <CloseOutlined />} {ListItems && ListItems[0]}
```
sebastianballesteros, line 21: Here the CloseOutlined will never show, because even if there is no sequence (sequence value is "")  the ListItems will be true. You would need to check for empty string to show the CloseOutlined.

frontend/src/components/DropdownListItems.js
```diff
@@ -0,0 +1,26 @@
...
+    <Menu>
+      {ListItems ? ListItems.map((item, i) => <Menu.Item key={i.toString()}>{item}</Menu.Item>) : <Menu.Item/>}
+    </Menu>
+  );
+
+  return (
+    ListItems && ListItems.length > 1 ? 
+    <Dropdown overlay={menuListItems} placement="bottomRight">
+      <div>
+        <MenuOutlined /> {ListItems && ListItems[0]}
```
sebastianballesteros, line 16: If there are many sequences, maybe it would be better to show a "Expand sequences" message, instead of the first item of the list i.e. ListItems[0]

frontend/src/components/DropdownListItems.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import {Button, Dropdown, Menu} from "antd";
+import {CloseOutlined, LineOutlined, MenuOutlined} from "@ant-design/icons";
+
+const DropdownListItems = ({ListItems}) => {
```
sebastianballesteros, line 5: Naming convention for variables is camelCase. That would imply that ListItems is a component, but maybe I am confused.

frontend/src/components/DropdownListItems.js
UlysseFG: I did change that (in the test PR) now the sequences for the 3 prime and 5 prime index will have empty list instead of a list with an empty string.

frontend/src/components/DropdownListItems.js
UlysseFG: expand sequences is no information at all the first one is still some additional info ... I do not know really. The lines makes it already visible that it is a list... Not sure anything would be gained by not showing the first one.

frontend/src/components/DropdownListItems.js
UlysseFG: you are right, i am just used to have CamelCase variables.

frontend/src/components/DropdownListItems.js
sebastianballesteros: We can at least add a way to tell the user that they need to hover on it to be able to see the list. Like we did with the sampleQC flag in SampleList.

PR #252, opened 2022-03-10
Library preparation

sebastianballesteros:


backend/fms_core/services/library.py
```diff
@@ -0,0 +1,103 @@
...
+        library_type = LibraryType.objects.get(name=name)
+    except LibraryType.DoesNotExist as e:
+        errors.append(f"No Experiment Type named {name} could be found.")
+    except LibraryType.MultipleObjectsReturned as e:
+        errors.append(f"More than one Experiment Type was found with the name {name}.")
+
+    return instrument, errors, warnings
+
+
+def create_library_batch(library_type, library_size, platform, library_date,
```
UlysseFG, line 27: I would call this prepare_library on the same model as extract_sample and transfer_sample...

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,103 @@
...
+    warnings += process_warnings + properties_warnings
+
+    if not errors:
+        for library_info in library_rows_info:
+            source_sample = library_info['sample_obj']
+            library_volume = library_info['volume']
+            index = library_info['index']
+            volume_used = library_info['volume_used']
+            comment = library_info['comment']
+            volume_destination = 0  # prevents this sample from being re-used or re-transferred afterwards
```
UlysseFG, line 57: Don't forget to remove this it is not applicable for library prep (what use is a library with 0 volume.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,103 @@
...
+
+            try:
+                library = Library.objects.create(library_type=library_type,
```
UlysseFG, line 60: Add another service create_library to return better error message if some input is missing or incorrect.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,103 @@
...
+                                    sample_source=source_sample,
+                                    container_destination=container_obj,
+                                    volume_used=volume_used,
+                                    execution_date=library_date,
+                                    coordinates_destination=container_coordinates,
+                                    volume_destination=library_volume,
+                                    comment=comment)
+
+                if sample_destination:
+                    sample_destination.depleted = True  # deplete destination sample
```
UlysseFG, line 94: Do not deplete ...

UlysseFG: We should wipe the qc_flags here. The library will need to be QCed before those flags get repopulated.

backend/fms_core/templates.py
```diff
@@ -95,6 +96,29 @@
...
+      {
+          'name': 'Library',
+          'headers': ['Library Batch ID', 'Sample Container Barcode', 'Sample Container Coordinates', 'Library Container Barcode',
+                      'Library Container Coordinates',  'Library Container Name', 'Library Container Kind', 'Library Parent Container Barcode',
+                      'Library Parent Container Coordinates', 'Sample Volume Used (uL)', 'Library Volume  (uL)',
+                      'Index', 'Comment'],
+      },
+  ],
+  "prefill info": [
+      ("Samples", "Sample Container Barcode", "container__barcode"),
```
UlysseFG, line 118: If you rename the tabs in the template you need to match the first value of the tuple to the new tab name,

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def prepare_library(library_type, library_size, platform, library_date,
+                    library_rows_info, protocol, process_properties, comment):
+    library = None
+    errors = []
+    warnings = []
+
+    if not library_type:
+        errors['library_type'] = "Invalid library type"
```
UlysseFG, line 54: "Missing library type." maybe to describe what you are actually testing.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,118 @@
...
+        create_process(protocol=protocol,
+                       creation_comment=comment if comment else f"Automatically generated via library preparation "
+                                                                f"on {datetime.utcnow().isoformat()}Z",)
+
+    # Create process' properties
+    if not process_errors:
+        properties, properties_errors, properties_warnings = create_process_properties(process_properties,
+                                                                                       process)
+
+    errors += process_errors + properties_errors
```
UlysseFG, line 67: tabulation here and next line. this will fail if there is process_errors (properties_errors won't be initialized and "+" won't be defined for NoneType)

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,118 @@
...
+                                         library_size=library_size,
+                                         index=index,
+                                         platform=platform)
+    except ValidationError as e:
+        errors.append(';'.join(e.messages))
+
+    return library, errors, warnings
+
+
+
```
UlysseFG, line 46: too many empty lines.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,118 @@
...
+def prepare_library(library_type, library_size, platform, library_date,
```
UlysseFG, line 47: I would match the service signatures to the internal structure instead of the template structure.
The param matching should be done in the importer here the template should be abstracted as much as possible. Maybe we failed so far to do that in other services, but we need to do better. I would have all the information used to define each library in a single iterable object ... inside the source sample info(source_sample, volume_used) and and the library_info(library_type, library_size, platform, index, library_date, library_comment) the other parameter can be separated since they belong with the process (process_properties, protocol, process_comment). At the moment, the date and library_size cannot be set by library they need to be all the same for a process.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,118 @@
...
+    errors += process_errors + properties_errors
+    warnings += process_warnings + properties_warnings
+
+    if not errors:
+        for library_info in library_rows_info:
+            source_sample = library_info['sample_obj']
+            library_volume = library_info['volume']
+            index = library_info['index']
+            volume_used = library_info['volume_used']
+            comment = library_info['comment']
```
UlysseFG, line 76: Do not set the parameter comment inside the function for a different use (process_measurement level comment), this is hard to track and it's error prone. Local variables are cheap, make a new one.

backend/fms_core/template_importer/importers/library_preparation.py
```diff
@@ -0,0 +1,100 @@
...
+            for i, (key, val) in enumerate(row.items()):
+                if i < self.properties_starting_index:
+                    library_batch_dict[key] = row[key]
+                else:
+                    process_properties[key]['value'] = val
+
+            library_preparation_kwargs = dict(
+                # Library attributes data dictionary and related objects
+                library_type={'name' : library_batch_dict['Experiment Type']},
+                library_size=process_properties['Shearing Size (bp)']['value'],
```
UlysseFG, line 85: We'll have to check this with the lab. We'll have to clarify with the lab if library_size = shearing_size and in that case what would be called the size of the library including the index + adapters/primers.

backend/fms_core/template_importer/row_handlers/library_preparation/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .library_preparation import LibraryPreparationRowHandler
+from .sample import SampleRowHandler
+
+__all__ = [
+    "LibraryPreparationRowHandler",
+    "SampleRowHandler",
```
UlysseFG, line 6: Might want to rename if we rename the tab to library ...

backend/fms_core/template_importer/row_handlers/library_preparation/library_preparation.py
```diff
@@ -0,0 +1,36 @@
...
+from fms_core.services.library import get_library_type
+from fms_core.services.platform import get_platform
+
+
+class LibraryPreparationRowHandler(GenericRowHandler):
+    def __init__(self):
+        super().__init__()
+
+    def process_row(self, **kwargs):
+        if len(kwargs['sample_rows_info']) < 1:
```
UlysseFG, line 13: this has been renamed library_rows_info

backend/fms_core/template_importer/row_handlers/library_preparation/library_preparation.py
```diff
@@ -0,0 +1,36 @@
...
+            _, self.errors['library'], self.warnings['library'] = prepare_library(library_type_obj,
+                                                                                  library_size,
+                                                                                  platform_obj,
+                                                                                  library_date,
+                                                                                  library_rows_info,
+                                                                                  protocol,
+                                                                                  process_properties,
+                                                                                  comment)
+        else:
+            self.errors['library'] = f"Errors prevent the creation of the library."
```
UlysseFG, line 36: We are at the level of the process here, you should mention that the "Library Preparation" failed. Maybe assign the error to something else than library. maybe library_preparation or simply preparation.

UlysseFG (approved):
ok this will do until we can test it further.

PR #254, opened 2022-03-12
Library Viewsets and Frontend 

UlysseFG:


backend/fms_core/models/sample.py
```diff
@@ -60,6 +60,10 @@ def is_depleted(self) -> str:
     def is_pool(self) -> bool:
         return DerivedBySample.objects.filter(sample=self).count() > 1 # More than 1 DerivedBySample implies more than 1 DerivedSample
 
+    @property
+    def is_library(self) -> bool:
+        return True if self.derived_sample_not_pool.library else False
```
UlysseFG, line 65: This could already be coded with pool in mind. In the case of a pool i think we can assume all pooled samples will be libraries, we could test any([derived_sample.library for derived_sample in self.derived_samples]) that should cover the future case of pooled libraries.

backend/fms_core/serializers.py
```diff
@@ -360,6 +388,98 @@ def get_quantity_flag(self, obj):
...
+            return None
+        else:
+            return "Passed" if obj.derived_sample_not_pool.quality_flag else "Failed"
+
+    # TODO : update this formula to include RNA and single strand DNA
+    def get_concentration_nm(self, obj):
+        if not obj.derived_sample_not_pool.library or not obj.derived_sample_not_pool.library.library_size:
+            return None
+        else:
+            return (obj.concentration / obj.derived_sample_not_pool.library.library_size * 660) * 1000000
```
UlysseFG, line 428: We need to update this with the strandedness. We should create constants MW_SSDNA and MW_DSDNA in models/_constants.py and use one or the other in the formula depending on strandedness.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,113 @@
...
+
+def prepare_library(library_type, library_date, platform, index, strandedness, library_volume, library_comment, container,
+                    container_coordinates, source_sample, volume_used, protocol, process_properties, process_comment=None):
+    process = None
+    errors = []
+    warnings = []
+
+    if not protocol:
+        errors.append("Missing protocol.")
+        return process, errors, warnings
```
UlysseFG, line 65: We should either return at each error like here or after all errors are tested like create_library, but not both. This is applicable to all services ... We should decide and apply one way for all services.

backend/fms_core/services/library.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if not process_properties:
+        errors.append("Missing process properties.")
+        return process, errors, warnings
+
+    process_by_protocol, process_errors, process_warnings = \
+        create_process(protocol=protocol,
```
UlysseFG, line 72: One process for each library is too much. You can initialize it (process_by_protocol) in the importer and pass it along instead of the protocol

backend/fms_core/template_importer/importers/library_preparation.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+            LIBRARY BATCH SHEET
+        """
+        library_batch_sheet = self.sheets['Library Batch']
+
+        # Iterate through libraries rows
+        library_batch_rows_data = defaultdict(list)
+        for row_id, row in enumerate(library_batch_sheet.rows):
+            library_batch_dict = {}
+            process_properties = self.preloaded_data['process_properties']
```
UlysseFG, line 43: This may be an error (i am aware that it is the same in the experiment_run ...). If we create more than one process the process_properties will be modified in between each iteration leading to false content for empty rows. This would need to be tested. I see you did a deepcopy lower, fixing this would require a deepcopy first to create a new memory reference then you could just pass off the reference at the end (if we test it and it does fail we'll need to check the experiment_run code too).

backend/fms_core/template_importer/row_handlers/library_preparation/library_batch.py
```diff
@@ -0,0 +1,28 @@
...
+            get_library_type(library_type)
+
+        platform_obj, self.errors['platform'], self.warnings['platform'] = get_platform(platform)
+
+        if not library_type_obj:
+            self.errors['library_preparation'] = f"Library type is required for library preparation."
+
+        if not platform_obj:
+            self.errors['library_preparation'] = f"Platform is required for library preparation."
+
```
UlysseFG, line 23: You can create the process in here and return it as a row_object...

backend/fms_core/tests/test_template_importers/library_preparation_test.py
```diff
@@ -0,0 +1,220 @@
...
+        self.assertEqual(sample_library_1.volume, self.library_1['library_volume'])
+
+        # Library info tests
+        self.assertEqual(library_1.library_type, self.library_1['library_type'])
+        self.assertEqual(library_1.platform, self.library_1['platform'])
+        self.assertEqual(library_1.index, self.library_1['index'])
+        self.assertEqual(library_1.strandedness, self.library_1['strandedness'])
+
+        # Process and process measurements tests
+        pm_1 = ProcessMeasurement.objects.get(source_sample=source_sample,
```
UlysseFG, line 129: Test somewhere in here that the process for library_1 and library_2 is the same...

backend/fms_core/viewsets/library.py
```diff
@@ -0,0 +1,117 @@
...
+            context['labels'] = {i: i.replace('_', ' ').capitalize() for i in fields}
+        return context
+
+    @action(detail=False, methods=["get"])
+    def summary(self, _request):
+        """
+        Returns summary statistics about the current set of libraries in the
+        database.
+        """
+        return Response({"total_count": self.get_queryset().all().count()})
```
UlysseFG, line 95: I would add a breakdown by library_type (formatted like the process_measurement "protocol_count" summary).

backend/fms_core/viewsets/sample.py
```diff
@@ -62,6 +61,18 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePrefill
             "template": [SAMPLE_SELECTION_QPCR_TEMPLATE["identity"]],
             "importer": SampleSelectionQPCRImporter,
         },
+        {
+            "name": "Prepare Libraries",
+            "description": "Upload the provided template with up to 384 new samples.",
```
UlysseFG, line 66: Maybe change the description to remove the false limit and better describe the template content.

frontend/src/components/DashboardPage.js
```diff
@@ -149,6 +152,19 @@ const DashboardPage = ({
               </Col>
             </Row>
           </Card>
+          <div style={{ display: 'flex', marginBottom: '1em' }}></div>
+          <Card title="Libraries" {...CARD_PROPS}>
```
UlysseFG, line 156: I would swap this card with the indices one and add the library_type statistic breakdown.

frontend/src/utils/api.js
```diff
@@ -67,6 +81,28 @@ const api = {
...
+      request: (options, template) => get(`/libraries/prefill_template/`, {template: template, ...options}),
+    },
+    search: q => get("/libraries/search/", { q }),
+  },
+
+  libraryTypes: {
+    get: libraryTypeId => get(`/library-types/${libraryTypeId}/`),
+    list: (options, abort) => get("/library-types/", options, { abort }),
+  },
+
```
UlysseFG, line 105: Missing platforms api calls. The action reducers are still there and use api.platforms.list(). We should either keep platform API call or remove all traces of it.

backend/fms_core/tests/test_template_importers/library_preparation_test.py
```diff
@@ -0,0 +1,222 @@
...
+        pt_3 = PropertyType.objects.get(name='Library Kit Lot', object_id=pm_3.process.protocol.id)
+        p_3 = PropertyValue.objects.get(property_type_id=pt_3, object_id=pm_3.process.id)
+
+        pt_4 = PropertyType.objects.get(name='Shearing Technician Name', object_id=pm_3.process.protocol.id)
+        p_4 = PropertyValue.objects.get(property_type_id=pt_4, object_id=pm_3.process.id)
+
+        self.assertEqual(p_1.value, self.library_3['library_technician_name'])
+        self.assertEqual(p_2.value, self.library_3['library_kit_used'])
+        self.assertEqual(p_3.value, self.library_3['library_kit_lot'])
+        self.assertEqual(p_4.value, ' ')
```
UlysseFG, line 220: Unset properties are equal to one space character ? This is not ideal ...

UlysseFG:
The template platform dropdown still contains a lot of platforms that are not relevant to the center. Keep it short (ILLUMINA, DNBSEQ).

backend/fms_core/template_importer/importers/library_preparation.py
```diff
@@ -0,0 +1,104 @@
...
+                     },
+                'volume_used': float_to_decimal_and_none(row_data['Sample Volume Used (uL)']),
+                'comment': row_data['Comment'],
+                'container':
+                    {'barcode': row_data['Library Container Barcode'],
+                     'coordinates': row_data['Library Container Coordinates'],
+                     'name': row_data['Library Container Name'],
+                     'kind': row_data['Library Container Kind'],
+                     'parent_barcode': row_data['Library Parent Container Barcode'],
+                     'parent_coordinates': row_data['Library Container Coordinates']
```
UlysseFG, line 91: Wrong data column : "Library **Parent** Container Coordinates"

backend/fms_core/tests/test_template_importers/library_preparation_test.py
sebastianballesteros: I agree. When the importer was being developed I suggested that empty property values were not created, but it was the way it was set.

frontend/src/components/DashboardPage.js
```diff
@@ -149,6 +162,15 @@ const DashboardPage = ({
               </Col>
             </Row>
           </Card>
+          <div style={{ display: 'flex', marginBottom: '1em' }}></div>
```
UlysseFG, line 165: Indices might be better before the Other card in the right column.

backend/fms_core/models/_constants.py
```diff
@@ -3,6 +3,9 @@
 STANDARD_NAME_FIELD_LENGTH = 200
 STANDARD_SEQUENCE_FIELD_LENGTH = 500
 PROJECT_STATUS_CHOICES = ["Open", "Closed"]
+STRANDEDNESS_CHOICES = ["Double stranded", "Single stranded"]
```
UlysseFG, line 6: You should reference the constants in here so if we change them it propagates here too.

PR #259, opened 2022-03-24
Import samples as library

UlysseFG:


sebastianballesteros:
In the template you added two more platforms (Nanopore and PacBio), we will need to add them to the library prep template as well. The strings for Single stranded and Double stranded have the 'stranded' part not capitalized.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -97,13 +100,26 @@ def process_row_inner(self, sample, container, project, parent_container, indivi
...
+            # Create library objects
+            library_type_obj, self.errors['library_type'], self.warnings['library_type']  = get_library_type(library['library_type'])
+            index_obj, self.errors['index'], self.warnings['index']  = get_index(library['index'])
+            platform_obj, self.errors['platform'], self.warnings['platform']  = get_platform(library['platform'])
+
+            library_obj, self.errors['library'], self.warnings['library'] = create_library(library_type=library_type_obj,
+                                                                                           index=index_obj,
+                                                                                           platform=platform_obj,
+                                                                                           strandedness=library['strandedness'])
+        if not self.errors:
```
sebastianballesteros, line 114: This part will never be executed because self.errors was initialized already. Thus the sample will never be created.  Try a different  way.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -97,13 +100,26 @@ def process_row_inner(self, sample, container, project, parent_container, indivi
...
+            sample_obj, self.errors['sample'], self.warnings['sample'] = \
+                create_full_sample(name=sample['name'], volume=sample['volume'], collection_site=sample['collection_site'],
+                                  creation_date=sample['creation_date'], coordinates=sample['coordinates'], alias=sample['alias'],
+                                  concentration=sample['concentration'], tissue_source=sample['tissue_source'],
+                                  experimental_group=sample['experimental_group'], container=container_obj, individual=individual_obj,
+                                  library=library_obj, sample_kind=sample_kind_obj, comment=comment)
```
sebastianballesteros, line 120: Once this gets executed and the sample created you will have a problem with library_obj. If library is not submitted in the template the library_obj variable will be referenced before assignment. You'd have to assign it before

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
UlysseFG: Ugh did not see that. why would we do that ... maybe it is time to uninitialize them ...

sebastianballesteros:
Also, the indices under the set "IDT 10nt UDI TruSeq Adapter" are not quite correct. You can copy them from the library prep template. It is just basically that the 1 in "IDT_10nt_UDI_i7_001" is repeated throughout the list instead of being incremented.

PR #265, opened 2022-04-05
Move qc flags sample

sebastianballesteros:


backend/fms_core/serializers.py
```diff
@@ -427,16 +425,16 @@ class Meta:
...
-        if obj.derived_sample_not_pool.quality_flag is None:
+        if obj.quality_flag is None:
             return None
         else:
-            return "Passed" if obj.derived_sample_not_pool.quality_flag else "Failed"
+            return "Passed" if obj.quality_flag else "Failed"
 
     def get_quantity_flag(self, obj):
-        if obj.derived_sample_not_pool.quality_flag is None:
+        if obj.quality_flag is None:
```
UlysseFG, line 434: quantity ...

backend/fms_core/serializers.py
```diff
@@ -390,16 +388,16 @@ def get_mother_name(self, obj):
...
```
UlysseFG, line 397: quantity

backend/fms_core/serializers.py
```diff
@@ -460,16 +458,16 @@ class Meta:
...
```
UlysseFG, line 467: quantity ...

backend/fms_core/serializers.py
```diff
@@ -507,16 +505,16 @@ class Meta:
...
```
UlysseFG, line 514: quantity

backend/fms_core/services/sample.py
```diff
@@ -340,13 +340,14 @@ def prepare_library(process: Process,
                 coordinates=coordinates_destination if coordinates_destination else "",
                 creation_date=execution_date,
                 volume=volume_destination if volume_destination is not None else volume_used,
-                depleted=False
+                depleted=False,
```
UlysseFG, line 343: Should we add a depleted column to the lib prep template in case there is an issue and the resulting lib is junk ... ? Maybe we can ask that during our next meeting with the lab.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -39,7 +39,7 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
                               volume=new_volume,
                               concentration=sample_information['concentration'])
 
-            # Update derived_sample's flags with sample information
+            # Update the sample's flags with sample information
             _, self.errors['flags'], self.warnings['flags'] = update_qc_flags(sample_obj,
                                                                               sample_information['quantity_flag'],
```
UlysseFG, line 44: i would use named parameter here because it's very easy to confuse or intervert the order they are passed to the function.

backend/fms_core/tests/test_template_importers/library_preparation_test.py
```diff
@@ -100,9 +100,12 @@ def prefill_data(self):
...
 
-        create_full_sample(name=self.source_sample_name, volume=self.source_sample_initial_volume, concentration=25,
-                           collection_site='TestCaseSite', creation_date=datetime.datetime(2022, 1, 15, 0, 0),
-                           container=container, sample_kind=sample_kind)
+        (source_sample, errors, warnings) = \
+            create_full_sample(name=self.source_sample_name, volume=self.source_sample_initial_volume, concentration=25,
+                               collection_site='TestCaseSite', creation_date=datetime.datetime(2022, 1, 15, 0, 0),
+                               container=container, sample_kind=sample_kind)
+
+        update_qc_flags(source_sample, "Passed", "Passed")
```
UlysseFG, line 108: You could test if those are still set after the template ingestion.

UlysseFG:
There is an important part that is missing ... Since we no longer carry the flags with the derived sample the flag need to carry over in a number of situation from a source_sample to it's child. One case is the transfer and another is the infinium experiment. I don't think there is other one at the moment but Normalisation is likely to be one. The resetting (or not) of the flags should be part of the parameters of the functions that process the samples (the function should be told when creating a child sample if it copies or reset the flags)...

UlysseFG (approved):

PR #266, opened 2022-04-06
Added annotation and filter for a generic qc flag and fixed a display bug

sebastianballesteros:


backend/fms_core/viewsets/sample.py
```diff
@@ -23,7 +23,13 @@
 
 
 class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePrefillsMixin):
-    queryset = Sample.objects.select_related("container").all().distinct()
+    queryset = Sample.objects.annotate(
```
UlysseFG, line 26: apply the annotation on the previous query instead of inserting it in the chain. it would be more readable to have them separated (example
```
queryset = sample.objects.filter(name=bob).all()
queryset = queryset.annotate(patate_poil=Q()& Q()))
```

frontend/src/components/samples/SamplesListContent.js
```diff
@@ -120,7 +120,6 @@ const getTableColumns = (containersByID, individualsByID, projectsByID, sampleKi
     {
       title: "QC Flag",
       dataIndex: "qc_flag",
-      sorter: true,
```
UlysseFG, line 123: I would keep the sorter in case the user want to group up his multiple selection (if there was only 2 groups it would be useless to have both but for 3 groups it could be used).

frontend/src/components/filters/descriptions.js
```diff
@@ -79,6 +79,17 @@ export const SAMPLE_FILTERS = {
...
   },
+  qc_flag: {
+    type: FILTER_TYPE.SELECT,
+    key: "qc_flag",
+    label: "QC Flag",
+    placeholder: "All",
+    mode:"multiple",
+    options: [
+      { label: "Passed", value: "true" },
+      { label: "Failed",  value: "false"},
```
UlysseFG, line 90: filtering for null may be desirable (looking for samples that need to be QCed)

backend/fms_core/filters.py
```diff
@@ -41,6 +41,7 @@ class SampleFilter(GenericFilter):
     container__barcode = django_filters.CharFilter(field_name="container__barcode", method="batch_filter")
     qPCR_status__in = django_filters.CharFilter(method="process_measurement_properties_filter")
     projects__name = django_filters.CharFilter(method="batch_filter")
+    qc_flag__in = django_filters.CharFilter(method="qc_flag_filter")
```
UlysseFG, line 44: You can include the qc_flag filter in the library_filter too i think.

UlysseFG (approved):

PR #268, opened 2022-04-11
 Created model/migration and modified importer

sebastianballesteros:


backend/fms_core/models/sample_property.py
```diff
@@ -0,0 +1,19 @@
...
+from .tracked_model import TrackedModel
+from django.db import models
+from .sample import Sample
+
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+from ._validators import name_validator
+
+
+@reversion.register()
+class SampleProperty(TrackedModel):
```
UlysseFG, line 12: I would not call this property even if this is close in many way to the properties ... We will just get confused and it will also be hard to explain to the users why some properties can be accessed in a way in some other another way. We often call them additional meta-data with David, but this is a bit long ... We could invent a term for them Sample Addtributes ? MetAdData ? eheh well maybe it's time for a brainstorming meeting ...

backend/fms_core/models/sample_property.py
```diff
@@ -0,0 +1,19 @@
...
+    """ Class to store additional sample metadata. """
+
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
+                            help_text="The name of the property.")
+    value = models.CharField(max_length=200, help_text="The value of the property.s")
+    sample = models.ForeignKey(Sample, on_delete=models.PROTECT, related_name='properties',
```
UlysseFG, line 18: This needs to be stored at the biosample level ...

backend/fms_core/services/sample.py
```diff
@@ -443,10 +443,76 @@ def remove_qc_flags(sample):
...
+def remove_sample_properties(sample, properties):
+    num_objects_deleted = 0
+    errors = []
+    warnings = []
+
+    if sample and properties:
+        try:
+            for property_name in properties.keys():
+                property_value = properties[property_name]
+                num_objects_deleted, _ = SampleProperty.objects.filter(name=property_name, value=property_value, sample=sample).delete()
```
UlysseFG, line 505: I don't think this will call the model delete() which is the one we need to call to set the delete flag in the trackingfields so we record the current value in reversion. Since this can apply delete to multiple delete this will likely call the sampleProperty manager delete() instead (which we haven't overloaded). You can double check if you want to confirm and not go just by my word. I think calling first() after filter would remedy this.

backend/fms_core/models/sample_property.py
```diff
@@ -0,0 +1,19 @@
...
+from ._validators import name_validator
+
+
+@reversion.register()
+class SampleProperty(TrackedModel):
+    """ Class to store additional sample metadata. """
+
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
+                            help_text="The name of the property.")
+    value = models.CharField(max_length=200, help_text="The value of the property.s")
```
UlysseFG, line 17: I don't think there is a cost to longer max_length in the DB ... The fields are varchar and are allocated the length required. Since this field could contain anything i would recommend giving it longer than 200 ... maybe 2000 ? ... this is not the allocated size as far as i know only the max permitted by us.

backend/fms_core/models/sample_property.py
```diff
@@ -0,0 +1,19 @@
...
+
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+from ._validators import name_validator
+
+
+@reversion.register()
+class SampleProperty(TrackedModel):
+    """ Class to store additional sample metadata. """
+
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
```
UlysseFG, line 15: i would add a unique constraint with the sample id in order to have just one of a given metadata name by sample ...
```
class Meta:
    constraints = [
            models.UniqueConstraint(fields=['name', 'biosample_id'], name='unique_metadata_name_by_biosample_id')
      ]

```

backend/fms_core/models/sample_metadata.py
```diff
@@ -0,0 +1,19 @@
...
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+from ._validators import name_validator
+
+
+@reversion.register()
+class SampleMetadata(TrackedModel):
+    """ Class to store additional sample metadata. """
+
+    name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, validators=[name_validator],
+                            help_text="The name of the property.")
```
UlysseFG, line 16: Help text not updated.

backend/fms_core/templates.py
```diff
@@ -119,6 +120,21 @@
       ("Library", "Sample Container Coordinates", "coordinates"),],
 }
 
+SAMPLE_ADD_METADATA_TEMPLATE = {
```
UlysseFG, line 123: I would remove the _ADD in the name of the template it does not only add metadata.

backend/fms_core/template_importer/importers/sample_add_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        metadata_sheet = self.sheets['Metadata']
+
+        # Add the extra headers to the template sheet according to the metadata
+        metadata_headers = [key for i, (key, val) in enumerate(metadata_sheet.rows[0].items()) if i >= self.metadata_starting_index]
+        extra_headers = {
+            # sheet_number : extra metadata headers
+            0: metadata_headers
+        }
+        self.add_dynamic_headers(extra_headers)
```
UlysseFG, line 26: This appears to be redundant ... I think it may very well work by simply replacing self.headers by self.dataframe.columns in the generate_preview_info_from_rows_results() function. We would have to check if it does not break something else (experiment_run) but it would be more logical to have all input columns in the report.

backend/fms_core/services/sample.py
```diff
@@ -443,10 +443,88 @@ def remove_qc_flags(sample):
...
+def remove_sample_metadata(sample, metadata):
+    num_objects_deleted = 0
+    errors = []
+    warnings = []
+
+    if sample and metadata:
+        try:
+            # Retrieve the biosample of the given sample
+            biosample_obj = sample.biosample_not_pool
+            for name in metadata.keys():
```
UlysseFG, line 511: .items() will return the key, value pair

backend/fms_core/services/sample.py
```diff
@@ -443,10 +443,88 @@ def remove_qc_flags(sample):
...
+                value = metadata[name]
+                metadata_obj = SampleMetadata.objects.get(name=name, biosample=biosample_obj)
+                # Add warning if the new value is the same as the old value
+                if metadata_obj.value != value:
```
UlysseFG, line 515: I am not sure this will be sufficient and safe for this functionality. I am afraid someone will mix ADD, UPDATE and DELETE in a template. I would force them to type in "DELETE" in the deleted column to make sure or we could prevent them from mixing actions.

backend/fms_core/viewsets/sample_metadata.py
```diff
@@ -0,0 +1,17 @@
...
+
+from ._constants import FK_FILTERS
+
+class SampleMetadataViewSet(viewsets.ModelViewSet):
+    queryset = SampleMetadata.objects.all()
+    serializer_class = SampleMetadataSerializer
+    pagination_class = None
+    permission_classes = [IsAuthenticated]
+
+    filterset_fields = {
```
UlysseFG, line 15: Do not set the hardcoded value here put it in the _constants.py file.

backend/fms_core/template_importer/importers/sample_add_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+        self.metadata_starting_index = METADATA_STARTING_INDEX
+
+    def import_template_inner(self):
+        """
+            METADATA SHEET
+        """
+        metadata_sheet = self.sheets['Metadata']
+
+        # Add the extra headers to the template sheet according to the metadata
+        metadata_headers = [key for i, (key, val) in enumerate(metadata_sheet.rows[0].items()) if i >= self.metadata_starting_index]
```
UlysseFG, line 21: makes me think of metadata name validation ... where would it go ? what would happen if a name has a space ? ... I guess this could appear in each row. We should just make sure the message is clear enough to understand it points to the header row...

backend/fms_core/template_importer/importers/sample_add_metadata.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # Iterate through experiment rows
+        for row_id, row in enumerate(metadata_sheet.rows):
+            row_data = {}
+            metadata = {}
+            for i, (key, val) in enumerate(row.items()):
+                if i < self.metadata_starting_index:
+                    row_data[key] = row[key]
+                # Just add metadata if the cell is not empty, else ignore
+                elif row[key]:
```
UlysseFG, line 36: i would enforce lower case on the name of the metadata ... i guess this is where this would happen. This is to prevent having to test in the model if the name match another with different letter case. We could also create a new "lowercase name" validator to make sure we do not insert upper case names by a back door.

backend/fms_core/template_importer/importers/sample_add_metadata.py
sebastianballesteros: I think it might be better instead of replacing them too add the columns that are not part of the header.

backend/fms_core/models/sample_metadata.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+from ._validators import name_validator
+
+
+@reversion.register()
+class SampleMetadata(TrackedModel):
+    """ Class to store additional sample metadata. """
+
+    name = models.CharField(max_length=2000, validators=[name_validator],
```
UlysseFG, line 15: not 2000 here ... only for value... we do not want 2000 character long field name ;).

backend/fms_core/tests/test_template_importers/add_sample_metadata_test.py
```diff
@@ -0,0 +1,86 @@
...
+
+class SampleMetadataTestCase(TestCase):
+    def setUp(self) -> None:
+        self.importer = SampleAddMetadataImporter()
+        self.file = APP_DATA_ROOT / "Sample_metadata_v3_8_0.xlsx"
+        ContentType.objects.clear_cache()
+
+        self.sample1_name = 'SampleTestMetadata'
+        self.sample2_name = 'Sample2TestMetadata'
+        self.sample_metadata1_name = "MetadataName1"
```
UlysseFG, line 22: If you add the lower case conversion, test also that the name is lower(name). You can also test the case of the space replacement by underscore.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -298,6 +312,18 @@ const SampleDetailsContent = ({
           <SamplesAssociatedProjects sampleID={sample.id} />
         </TabPane>
 
+        <TabPane tab={`Metadata`} key="5" style={tabStyle}>
+          <Title level={5} style={{ marginTop: '1rem'}}> Metadata </Title>
+          <Descriptions bordered={true} size="small">
+            {
+              sampleMetadata.map(metadata => {
+                return  <Descriptions.Item label={metadata.name}>{metadata?.value} </Descriptions.Item>
```
UlysseFG, line 320: if metadata is not defined it will fail for the name too.

backend/fms_core/services/sample.py
```diff
@@ -443,10 +443,87 @@ def remove_qc_flags(sample):
...
+                    metadata_obj.delete()
+                    deleted = True
+        except SampleMetadata.DoesNotExist:
+            errors.append(f'Metadata with name [{name}] is not tied to sample [{sample.name}]')
+    else:
+        errors.append('Sample and metadata are required')
+
+    return deleted, errors, warnings
+
+
```
UlysseFG, line 524: Lots of empty lines

backend/fms_core/templates.py
```diff
@@ -119,6 +120,21 @@
...
+SAMPLE_METADATA_TEMPLATE = {
+  "identity": {"description": "Template to add metadata to samples", "file": static("submission_templates/Sample_metadata_v3_8_0.xlsx")},
+  "sheets info": [
+      {
+          'name': 'Metadata',
+          'headers': ['Action', 'Sample Name', 'Sample Container Barcode', 'Sample Container Coordinates'],
+      },
+  ],
+  "prefill info": [
+      ("Samples", "Sample Name", "name"),
```
UlysseFG, line 132: Tab name is "Metadata" currently.

backend/fms_core/template_importer/row_handlers/sample_add_metadata/sample_add_metadata.py
```diff
@@ -0,0 +1,57 @@
...
+
+from fms_core.template_importer.row_handlers._generic import GenericRowHandler
+
+from fms_core.services.sample import get_sample_from_container, add_sample_metadata, update_sample_metadata, remove_sample_metadata
+
+ADD_ACTION = "ADD"
+UPDATE_ACTION = "UPDATE"
+REMOVE_ACTION = "REMOVE"
+
+class SampleAddMetadataHandler(GenericRowHandler):
```
UlysseFG, line 13: SampleMetadataHandler would be better in my opinion

UlysseFG (approved):

PR #269, opened 2022-04-11
Fixed input_to_date_and_none() function and added additional format comment

sebastianballesteros:


backend/fms_core/template_importer/_utils.py
```diff
@@ -14,8 +14,8 @@ def input_to_date_and_none(date_input):
     elif isinstance(date_input, datetime.date):
         return date_input
     else:
-        year, month, day = date_input.split("-")
         try:
+            year, month, day = date_input.split("-")
```
UlysseFG, line 18: eheh yeah this would crash ... Good idea to pair this with an additional detail in the error message.

backend/fms_core/services/process_measurement.py
```diff
@@ -15,7 +15,7 @@ def create_process_measurement(process, source_sample, execution_date, volume_us
     if not source_sample:
         errors.append(f"Source sample is required for process measurement creation.")
     if not execution_date:
-        errors.append(f"Execution date is required for process measurement creation.")
+        errors.append(f"Execution date is required for process measurement creation. Format expected: YYYY-MM-DD")
```
UlysseFG, line 18: I would wrap this together in a single sentence, "An execution date with the format YYYY-MM-DD must be provided for process measurement creation."

UlysseFG (approved):

backend/fms_core/services/process_measurement.py
sebastianballesteros: Good call

PR #272, opened 2022-04-21
Last changes before release

sebastianballesteros:
Added thermocycler metadata, indices, modified column order, etc


UlysseFG:
In the template, thermocycler used drop down has "é" in the names. This is not very english ... we should remove the accents.

UlysseFG:
Ok Just saw a small thing that i would change but i'll leave it up to you to see if it is worth it. On the sample detail page for libraries the information specific to libraries are a bit randomly placed. I would place library_type, platform and index on the top row and the numeric values (which are sometime empty on the bottom row. (Library size, Concentration (nM), NA Quantity (ng)).

UlysseFG (approved):

PR #273, opened 2022-04-25
Update webpack for node 18

ckostiw:
Node 18 dropped support for md4 encryption, and webpack builds would fail if running node 18. Updating to the latest build of webpack 5 solves this.
This is a tentative fix, but cypress will also need to be updated, as the current version won't run on anything above node 15.


nafiz1001:
This is being addressed by the PR #325 😉

PR #275, opened 2022-04-29
Improve export performance

sebastianballesteros:


backend/fms_core/viewsets/sample.py
```diff
@@ -261,8 +264,16 @@ def update(self, request, *args, **kwargs):
 
     @action(detail=False, methods=["get"])
     def list_export(self, _request):
-        serializer = SampleExportSerializer(self.filter_queryset(self.get_queryset()), many=True)
-        return Response(serializer.data)
+        # Select related models in derived sample beforehand to improve performance and prefetch then in sample queryset
+        derived_samples = DerivedSample.objects.all().select_related('biosample', 'biosample__individual')
+
+        queryset = self.filter_queryset(self.get_queryset())
+        queryset = queryset.prefetch_related(Prefetch('derived_samples', queryset=derived_samples))
```
UlysseFG, line 271: I'm wondering if this would not work for just any sample queryset ? This could be inserted into the sample viewset get_queryset() by replacing self.get_queryset() by self.queryset ?

backend/fms_core/viewsets/sample.py
```diff
@@ -261,8 +264,16 @@ def update(self, request, *args, **kwargs):
 
     @action(detail=False, methods=["get"])
     def list_export(self, _request):
-        serializer = SampleExportSerializer(self.filter_queryset(self.get_queryset()), many=True)
-        return Response(serializer.data)
+        # Select related models in derived sample beforehand to improve performance and prefetch then in sample queryset
+        derived_samples = DerivedSample.objects.all().select_related('biosample', 'biosample__individual')
+
+        queryset = self.filter_queryset(self.get_queryset())
```
UlysseFG, line 270: what happens if you filter by a prefetched field ? Django would fetch it before it's prefetched to do the filtering ? Can filtering be moved last ?

backend/fms_core/viewsets/sample.py
sebastianballesteros: I think it could work, we can try including it in get_queryset() function. It is a good suggestion.

backend/fms_core/viewsets/sample.py
sebastianballesteros: That's a good question. As far as I am concerned, if you filter Django is going to generate an extra query instead of using the cached values. 

https://stackoverflow.com/questions/12973929/why-does-djangos-prefetch-related-only-work-with-all-and-not-filter/12974801#12974801

https://docs.djangoproject.com/en/3.0/ref/models/querysets/#prefetch-related

sebastianballesteros: `The prefetch_related('toppings') implied pizza.toppings.all(), but pizza.toppings.filter() is a new and different query. The prefetched cache can’t help here; in fact it hurts performance, since you have done a database query that you haven’t used. So use this feature with caution!`

UlysseFG: Yeah i read that and my understanding is that if you filter after the prefetch it's useless but if you prefetch on a filtered queryset you'll be actually prefetching on a smaller subset of entries ... I was wondering if there is anything gained by the prefetch in these cases ... Maybe it's possible to include the filtering columns in the Prefetch like mentionned in your first link to be able to prefetch on the correct subset in order to filter last ... not sure it would gain anything though... It might be worth it to test the actual query used in different situations.

UlysseFG: I don't want you to put too much time right away on this though. Lets wrap the main content of the release and we can include these changes and once we are happy with the rest let's come back to this and the migration work.

backend/fms_core/serializers.py
```diff
@@ -339,10 +341,48 @@ def get_quantity_flag(self, obj):
...
+def serialize_sample_export(sample: Sample) -> Dict[str, Any]:
+    first_derived_sample = sample.derived_sample_not_pool
+    biosample = first_derived_sample.biosample
+    return {
+        'sample_id': sample.id,
+        'sample_name': sample.name,
+        'biosample_id': biosample.id,
+        'alias': biosample.alias,
+        'sample_kind': first_derived_sample.sample_kind.name,
+        'tissue_source': first_derived_sample.tissue_source,
```
UlysseFG, line 353: after my PR merge ... 'tissue_source': first_derived_sample.tissue_source.name,

UlysseFG (approved):

PR #277, opened 2022-05-09
Create Lineage Component

nafiz1001:
Using react-d3-graph to create a graph to visualize lineage in the lineage tab

- [x] shows all children
- [ ] shows all all parents (it currently only handles single parent)
- [x] clicking node redirects to the correct sample page
- [x] clicking edge redirects to the correct process page
- [ ] correctly formatted graph (sometimes a random edge duplicate sticks around, most of the time the graph is oriented weirdly)

Screen Shot 2022-05-09 at 3 36 21 PM


UlysseFG:
Very nice ... Just a quick couple comments... 
- I think the arrow head are a bit overkill i think you are still juggling with how the graph shows up but it would be nice to have them linked from left to right where each step to the right means a progress. I don't know the module but it would be easier to follow than arrow heads.
- The red colored to indicate the current sample is a very good idea, but i think a alternative way of making the current node stand out might be better (maybe concentric circles like a target ? or a star instead of a circle...) we could use the red and green to display the QC flags instead.
- When navigating, it might be good to use history.push() as you move people around to samples or protocols.
- Do not bother too much with squashing commits unless its really second nature for you. It's actually nice to see what you tried on the way.

nafiz1001:
@UlysseFG 
> I think the arrow head are a bit overkill i think you are still juggling with how the graph shows up but it would be nice to have them linked from left to right where each step to the right means a progress. I don't know the module but it would be easier to follow than arrow heads.

I would like that too. If I were to do that, I will have to write an algorithm by hand to position nodes in certain ways, which I don't mind and look forward to do so.

Another alternative is to use simpler graphing libraries such as react-d3-tree (like i showed in my first version on Slack) which always assumes that the structure is a tree (meaning pools cannot be represented since they have multiple parents). There might be better libraries out there, but I haven't found them yet.

nafiz1001:
this PR is from my fork, I will close and reopen a PR

PR #278, opened 2022-05-09
Changed test on property creation to fix 0 being confounded for None.

UlysseFG:


sebastianballesteros (approved):
Changes seem ok. However, are we accepting zero values for all (numeric) property types? i.e. volume measured, RIN, etc? (I know that value zero for CT value should be valid but should all property types accept zero?)

PR #280, opened 2022-05-10
Check if taxon is set when trying to create individual.

UlysseFG:


backend/fms_core/services/individual.py
```diff
@@ -32,17 +32,6 @@ def get_or_create_individual(name, sex=None, taxon=None, pedigree=None, cohort=N
     if not name:
```
sebastianballesteros, line 32: Would it be easier to add  
```
if not taxon:
        errors.append(f"Taxon is required to create and individual")
```

just after the `if not name`

sebastianballesteros: Oh true

sebastianballesteros (approved):

PR #282, opened 2022-05-10
Added useful links for individual and project

sebastianballesteros:


backend/fms_core/migrations/0037_v3_9_0.py
```diff
@@ -0,0 +1,28 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='individual',
+            name='alias',
+            field=models.CharField(blank=True, null=True, help_text='Original individual name used by external client.', max_length=200),
+        ),
+        migrations.AddField(
+            model_name='project',
+            name='external_name',
+            field=models.CharField(blank=True, help_text='Original project name used by external client.', max_length=200),
```
UlysseFG, line 21: null=True ? i don't think those fields will be required.

backend/fms_core/migrations/0037_v3_9_0.py
```diff
@@ -0,0 +1,28 @@
...
+        ),
+        migrations.AddField(
+            model_name='project',
+            name='external_id',
+            field=models.CharField(blank=True, help_text='Identifier to connect to an external system.', max_length=200),
```
UlysseFG, line 26: null=True, attribute not required

backend/fms_core/serializers.py
```diff
@@ -381,6 +382,8 @@ def serialize_sample_export(sample: Sample) -> Dict[str, Any]:
         'quality_flag': None if sample.quality_flag is None else ("Passed" if sample.quality_flag else "Failed"),
         'quantity_flag': None if sample.quantity_flag is None else ("Passed" if sample.quantity_flag else "Failed"),
         'projects': ''.join([project.name for project in sample.projects.all()]) if sample.projects.all() else None,
+        'projects_external_id': ''.join([project.external_id for project in sample.projects.all()]) if sample.projects.all() else None,
```
UlysseFG, line 385: Not sure we need to export them. They should be in the project endpoint but i would not add them to samples or samples/list_export

backend/fms_core/services/individual.py
```diff
@@ -34,6 +34,7 @@ def get_or_create_individual(name, sex=None, taxon=None, pedigree=None, cohort=N
     else:
         individual_data = dict(
             name=name,
+            alias=alias,
```
UlysseFG, line 37: This is an optional field.

sebastianballesteros:
I don't know why the fms.log keeps getting created. For the Project list addition, do we want the external_project_id in the project name or the external_project_name?

UlysseFG:
> I don't know why the fms.log keeps getting created. 
you must add the directory instead of the file directly. You can add .log to gitignore if you want. i don't think we'll be adding log file to the app.

>For the Project list addition, do we want the external_project_id in the project name or the external_project_name?
I would put the ID because it will be shorter and more concise mostly the external name may be 500 character long and is not likely to be useful internally. We could also altogether not display any of the external project information into the list.

sebastianballesteros:
I will make the changes and we can discuss in during the meeting this afternoon.

frontend/src/components/individuals/IndividualsDetailContent.js
```diff
@@ -42,6 +42,7 @@ const IndividualsDetailContent = ({individualsByID, taxonsByID, get}) => {
             <Descriptions bordered={true} size="small" column={3}>
                 <Descriptions.Item label="ID">{individual.id}</Descriptions.Item>
                 <Descriptions.Item label="Name">{individual.name}</Descriptions.Item>
+                <Descriptions.Item label="Alias">{individual.alias}</Descriptions.Item>
```
UlysseFG, line 45: I would rearrange the fields to be in the following order to keep the father mother and pedigree together: 
ID  name alias
taxon sex cohort
pedigree mother father

UlysseFG (approved):

PR #285, opened 2022-05-17
Library Conversion

sebastianballesteros:


backend/fms_core/models/property_type.py
```diff
@@ -27,6 +27,9 @@ class PropertyType(TrackedModel):
     object_id = models.PositiveIntegerField()
     content_object = GenericForeignKey('content_type', 'object_id')
 
+    class Meta:
+        unique_together = ("name", "object_id")
```
UlysseFG, line 31: use uniqueconstraint instead 

> However, a notice has come with the release of Django 2.2 stating that ‘unique_together’ will soon be deprecated and replaced by ‘UniqueConstraint’.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+            errors.append(f"Volume used ({volume_used}) is invalid.")
+        if sample_source and volume_used > sample_source.volume:
+            errors.append(
+                f"Volume used ({volume_used}) exceeds the current volume of the library ({sample_source.volume}).")
+
+    if not isinstance(execution_date, date):
+        errors.append(f"Execution date is not valid.")
+
+    # Retrieve library object linked to the source sample
+    library_source_obj = sample_source.derived_sample_not_pool.library
```
UlysseFG, line 85: We are about to deal with pool and this will happen soon. Let's assume a library conversion can be applied to a pool an create a library for each derived_sample.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
         errors.append(';'.join(e.messages))
 
     return library, errors, warnings
+
+
+def convert_library(process, platform, sample_source, concentration, library_size, container_destination,
```
UlysseFG, line 56: Is there any reason to pass the library_size and concentration here ? Even if we do not reset them on conversion, we can still get them from the sample_source.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+                    coordinates_destination, volume_used, volume_destination, execution_date, comment):
+    library_destination = None
+    errors = []
+    warnings = []
+
+    if not process:
+        errors.append(f"Process is required.")
+    if not platform:
+        errors.append(f"Platform is required.")
+    if not library_size:
```
UlysseFG, line 66: I don't think there is a reason to force a library size here at the service level.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+        errors.append(f"Sample in container {sample_source.name} is not a library.")
+    elif library_source_obj.platform == platform:
+        errors.append(f"Source library platform and destination library platform can't be the same.")
+
+    if not errors:
+        # Create new destination library
+        library_destination, errors_library_destination, warnings_library_destinations = \
+            create_library(library_type=library_source_obj.library_type,
+                           # TODO: Verify this with Janick
+                           library_size=library_size,
```
UlysseFG, line 96: take it from sample_source.library.library_size

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+            sample_source.volume = sample_source.volume - volume_used
+            sample_source.save()
+
+            library_destination_data = dict(
+                container_id=container_destination.id,
+                coordinates=coordinates_destination if coordinates_destination else "",
+                creation_date=execution_date,
+                # TODO: Verify this with Janick
+                concentration=concentration if concentration is not None else None,
+                volume=volume_destination if volume_destination is not None else volume_used,
```
UlysseFG, line 110: if volume destination is None it is an error do not set it to something else. You can test it at the beginning to make sure.

backend/fms_core/templates.py
```diff
@@ -97,6 +97,29 @@
...
+                      'Destination Library Container Barcode', 'Destination Library Container Coordinates',
+                      'Destination Library Container Name', 'Destination Library Container Kind',
+                      'Destination Library Parent Container Barcode', 'Destination Library Parent Container Coordinates',
+                      'Library Source Concentration (ng/uL)', 'Library Size (bp)', 'Input used for conversion (ng)',
+                      'Volume Used (uL)', 'Volume (uL)', 'Comment'],
+      },
+  ],
+  "prefill info": [
+      ("Library", "Library Source Container Barcode", "container__barcode"),
+      ("Library", "Library Source Container Coordinates", "coordinates"),],
```
UlysseFG, line 120: We should prefill "Library Source Concentration (ng/uL) and "Library Size (bp)" too.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+            )
+
+            new_derived_sample_data = {
+                "library_id": library_destination.id
+            }
+            derived_samples_destination = []
+            volume_ratios = {}
+            for derived_sample_source in sample_source.derived_samples.all():
+                new_derived_sample, errors_inherit, warnings_inherit = inherit_derived_sample(derived_sample_source,
+                                                                                              new_derived_sample_data)
```
UlysseFG, line 124: each derived_sample will have its own library. This is the case here if derived_sample = 1, but otherwise it would convert all included libraries to the same one.

backend/fms_core/services/library.py
```diff
@@ -48,3 +51,96 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+                errors.extend(errors_inherit)
+                warnings.extend(warnings_inherit)
+
+                derived_samples_destination.append(new_derived_sample)
+                volume_ratios[new_derived_sample.id] = DerivedBySample.objects.get(sample=sample_source,
+                                                                                   derived_sample=derived_sample_source).volume_ratio
+
+            library_destination, errors_process, warnings_process = _process_sample(process,
+                                                                                   sample_source,
+                                                                                   library_destination_data,
```
UlysseFG, line 134: name is confusing ... might be better called sample_destination_data since we have a library object.

backend/fms_core/template_importer/row_handlers/library_conversion/library_batch.py
```diff
@@ -0,0 +1,37 @@
...
+
+        # Create process' properties
+        properties, self.errors['process_properties'], self.warnings['process_properties'] = \
+            create_process_properties(process_properties, process_by_protocol)
+
+        if not platform_obj:
+            self.errors['library_conversion'] = f"Destination platform is required for library conversion."
+
+        if not properties:
+            self.errors['library_conversion'] = f"Unable to process this library batch."
```
UlysseFG, line 30: Very nebulous message. You may want to mention "... without the required valid metadata columns." or something else to hint the user on this error.

backend/fms_core/template_importer/row_handlers/library_conversion/library_batch.py
```diff
@@ -0,0 +1,37 @@
...
+            create_process(protocol=protocol,
+                           creation_comment=comment if comment
+                           else f"Automatically generated via library conversion "f"on {datetime.utcnow().isoformat()}Z")
+
+        # Create process' properties
+        properties, self.errors['process_properties'], self.warnings['process_properties'] = \
+            create_process_properties(process_properties, process_by_protocol)
+
+        if not platform_obj:
+            self.errors['library_conversion'] = f"Destination platform is required for library conversion."
```
UlysseFG, line 27: I would mention "valid" and list [DNBSEQ] as the valid choice(s). People are likely to enter MGI and wonder what is wrong.

backend/fms_core/template_importer/row_handlers/library_conversion/library_conversion.py
```diff
@@ -0,0 +1,67 @@
...
+
+class LibraryRowHandler(GenericRowHandler):
+    def __init__(self):
+        super().__init__()
+
+    def process_row_inner(self, library_batch_info, library_source, volume_used, comment,
+                          container, concentration, library_size, volume):
+
+        if not library_batch_info:
+            self.errors['library_conversion'] = 'No batch is associated with this library.'
```
UlysseFG, line 17: you can bail out right there ... It will explode right and left if there is no process.

backend/fms_core/template_importer/row_handlers/library_conversion/library_conversion.py
```diff
@@ -0,0 +1,67 @@
...
+                container_parent_obj, self.errors['parent_container'], self.warnings['parent_container'] = \
+                    get_container(barcode=container['parent_barcode'])
+
+            container_obj, created, self.errors['library_container'], self.warnings['library_container'] = get_or_create_container(
+                name=container['name'],
+                barcode=container['barcode'],
+                kind=container['kind'],
+                container_parent=container_parent_obj if container_parent_obj else None,
+                coordinates=container['parent_coordinates'] if container_parent_obj else None,
+                creation_comment=comment)
```
UlysseFG, line 38: I don't think this comment is meant for the container creation ... You would be better to add an automatic message. Something like : Container created automatically while doing library conversion + timestamp.

UlysseFG:
The template has a full line format which will break the prefill validation fields. There is also 2 different tones of blue in the metadata section of the the batch sheet.

backend/fms_core/templates.py
```diff
@@ -97,6 +97,33 @@
   "prefill info": [],
 }
 
+LIBRARY_CONVERSION_TEMPLATE = {
+  "identity": {"description": "Template to convert libraries", "file": static("submission_templates/Library_conversion_v3_9_0.xlsx")},
+  "sheets info": [
+      {
+          'name': 'Conversion Batch',
+          'headers': ['Library Batch ID', 'Date (YYYY-MM-DD)', 'Platform', 'Comment',
+                      'Technician Name', 'Kit Used', 'Kit Lot', 'Thermocycler Used', 'PCR Cycles'],
```
UlysseFG, line 106: Some of the header name do not match the Template column names.

UlysseFG:
The template has some issues still. The formula for Platform only applies to the first line after the header. On the conversion batch sheet, The Comment has a dropdown that points to the platform. The Thermocycler dropdown only works for the first line after it is broken. On the Library sheet, THe Library Batch ID dropdown formula is broken.

backend/fms_core/services/library.py
UlysseFG: This is still not applied to a pool. I know this is not a case now without pools but it will soon enough and this is much simpler to fix while you have the code fresh in mind instead of in 2 months where we may forget it.

backend/fms_core/template_importer/importers/library_conversion.py
```diff
@@ -0,0 +1,101 @@
...
+
+        """
+            LIBRARIES SHEET
+        """
+        libraries_sheet = self.sheets['Library']
+        for i, row_data in enumerate(libraries_sheet.rows):
+            library_conversion_kwargs = {
+                'library_batch_info': library_batch_rows_data[row_data['Library Batch ID']],
+                'library_source':
+                    {'barcode': row_data['Library Source Container Barcode'],
```
UlysseFG, line 79: Maybe include right away the str_cast_and_normalize right away where it applies (str inputs). Elsewhere i skipped the Batch ID since it's only compared to itself.

UlysseFG (approved):

PR #286, opened 2022-05-17
Lineage backend

nafiz1001:


backend/fms_core/services/sample_lineage.py
```diff
@@ -25,3 +26,82 @@ def create_sample_lineage(parent_sample, child_sample, process_measurement):
...
+
+            nodes.extend(new_child_samples)
+            edges.extend(new_process_measurements)
+
+            stack.extend(new_child_samples)
+            visited.update(s.id for s in new_child_samples)
+        
+        def find_process_measurement(parent_sample, child_sample):
+            process_measurements = list(add_lineage(parent_sample.process_measurement.all()))
+            for p in process_measurements:
```
UlysseFG, line 83: searching iteratively in your queryset is very likely to be slower than using the queryset filtering capacity. This function seem to be processMeasurement.objects.filter(source_sample=parent_sample, child_sample=child_sample) with the extra annotations and select_related...

backend/fms_core/services/sample_lineage.py
```diff
@@ -25,3 +26,82 @@ def create_sample_lineage(parent_sample, child_sample, process_measurement):
...
+            return queryset
+
+        def query_set_iterator(f, l):
+            for x in l:
+                try:
+                    yield f(x)
+                except ObjectDoesNotExist:
+                    yield None
+
+        # fetch children
```
UlysseFG, line 54: I am not sure you need to start from the middle and basically repeat the operation up and down. All the samples in a lineage share the same biosample.id. So if you want all samples you can get them with a single query by testing this. To get the root sample(s), you can eliminate the ones that are child_sample in the lineage table. I haven't tracked your algorithm closely but it should be equivalent to putting on the stack all root samples in your fetch children section. This would simplify the graph traversal algorithm i think. You may even want to dodge the iterative traversal by listing all nodes using the biosample.id and listing all edges using the fact that all edges will be the ones that have a samplelineage.parent_sample and samplelineage.child sample in the node list. Up to you...

backend/fms_core/viewsets/sample_lineage.py
```diff
@@ -0,0 +1,23 @@
...
+    @action(detail=True, methods=["get"])
+    def graph(self, _request, pk):
+        parent_sample = Sample.objects.get(pk=pk)
+        nodes, edges, errors = create_sample_lineage_graph(parent_sample)
+
+        if errors:
+            raise ValidationError(errors)
+        else:
+            return Response({
+                "nodes": [SampleSerializer(s).data for s in nodes],
```
UlysseFG, line 21: This is very onerous for the usage we make of it ... your get() will fetch many samples and processMeasurement without putting them in the store. The largest part of the information gathered will never be used. I would suggest to go one of two direction maybe @ckostiw and @sebastianballesteros can help see the best way. I would either 
1) return just the id (+ parent and child for edges), label (either sample name for nodes or protocol for edges) and maybe QC for some eye candy, but the idea is to carry as little as needed in a throwaway approach. 

2) go for the reuse approach and use redux action reducer to store your samples and processMeasurements in the store for reuse when processing the graph request. 

Also, you could also mix those 2 options and have a skeleton graph with ids and use the existing sample and processMeasurement action reducers to get their full representation in a single get().

backend/fms_core/viewsets/sample_lineage.py
```diff
@@ -0,0 +1,23 @@
+from tkinter import W
```
UlysseFG, line 1: This causes an error and does not seem to be used.

backend/fms_core/services/sample_lineage.py
nafiz1001: The intention of my algorithm is that I want to explore the entire children of the middle node, but I'm not interested in all of the children of parents. My concern is that I might fetch too many samples.

backend/fms_core/viewsets/sample_lineage.py
sebastianballesteros: I tend to like the first option better. Given that there's no redux action/reducer at the moment, it might be better to just return the information needed (id, process measurements) to build the graph.

UlysseFG (approved):

sebastianballesteros:
The general implementation works great and the changes previously made in the backend are well done. My comments would be more in the frontend/UI side of the implementation. I like the addition of the legend. However, I don't really think the format and the style of the component match the current style we have on FreezeMan. You have a lot of components and resources provided by antd: https://ant.design. You might want to incorporate a style that goes along with the style we have currently on FreezeMan instead of just 2 big squares. 

In addition, and this is just personal preference, is it possible to change the design of the graph to be more "minimalistic". Like this: https://codesandbox.io/s/cvxu8. It is just an example, not a strict suggestion.

ckostiw (changes requested):
The code is nice and clean and I can't find anything I think needs to be changed. The only issue I ran into was that the graph renders outside the border that was created to hold the graph, so it looks a bit broken.
Screen Shot 2022-05-20 at 2 58 31 PM

ckostiw:
I like how generating the graph data on the backend simplified the rendering code  👍

ckostiw (approved):

ckostiw:
I added my approval in case I get sick and am unable to work next week. I only reported the one small issue about the graph going outside its border, which is no big deal.

ckostiw:
Noice. 

One nitpick: could the process labels be centered on the edge arrows? When the graph is small, the process text runs into the child node:

Screen Shot 2022-05-27 at 9 26 13 AM
?

ckostiw (approved):

ckostiw:
Another nitpick. The clip rect for the graph doesn't align with the tab content bounds, so the graph clipping looks a bit broken.
Screen Shot 2022-05-27 at 9 33 27 AM
Screen Shot 2022-05-27 at 9 33 15 AM
Screen Shot 2022-05-27 at 9 33 04 AM

nafiz1001:
> Another nitpick. The clip rect for the graph doesn't align with the tab content bounds, so the graph clipping looks a bit broken.
> Screen Shot 2022-05-27 at 9 33 27 AM
> Screen Shot 2022-05-27 at 9 33 15 AM
> Screen Shot 2022-05-27 at 9 33 04 AM

Screen Shot 2022-05-31 at 9 19 16 AM

Fixed it :D

ckostiw:
My OCD is relieved by the fix for the clipping issue, thanks. :)

PR #287, opened 2022-05-18
Update requirements

ckostiw:
These updates solved `pip install -r requirements.txt` issues, where install was failing due to build errors during the installation process.

### DIFF
--- a/backend/requirements.txt
+++ b/backend/requirements.txt
@@ -4,7 +4,7 @@ asn1crypto==1.4.0
 attrs==19.3.0
 CacheControl==0.12.6
 certifi==2020.6.20
-cffi==1.14.2
+cffi==1.15.0
 chardet==3.0.4
 codecov==2.1.8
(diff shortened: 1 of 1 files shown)


UlysseFG:
Can you specify what build issues this fixes ?

ckostiw:
> Can you specify what build issues this fixes ?

pip install was failing because of compilation errors in the sources of the modules I updated. Would you like me to roll back and record the error messages I ran in to?

UlysseFG:
>  Would you like me to roll back and record the error messages I ran in to?
Yikes nope. But i am just wondering because we do not have these issues so i was wondering if those were Mac only issues (those!) or something we may have overlooked. We cannot expect to bump versions each time Mac decide to be temperamental ;).

UlysseFG (approved):
Haven't found any issue. I'll merge so we can test these for as long as possible before release.

PR #288, opened 2022-05-18
Conserve Data validation after prefilling.

UlysseFG:
Replaced some formatting in the headers of some templates to fix the issue with Data validation being lost on prefilling.


backend/fms_core/template_prefiller/prefiller.py
```diff
@@ -14,7 +14,7 @@ def PrefillTemplate(template_path, template_info, queryset):
     """
     out_stream = BytesIO()
 
-    workbook = load_workbook(filename=template_path)
+    workbook = load_workbook(filename=template_path, keep_links=True, keep_vba=True)
```
sebastianballesteros, line 17: that was a quick fix ;)

sebastianballesteros (approved):

backend/fms_core/template_prefiller/prefiller.py
UlysseFG: That's not even the fix ... Just thought it would be nice to keep the vba macros if we ever have some in our templates. The fix is actually the template themselves. The 3 template had the issue with format applied row wide.

PR #290, opened 2022-05-24
Library qc

ckostiw:
Library QC template and import


backend/fms_core/templates.py
```diff
@@ -116,10 +117,34 @@
       },
   ],
   "prefill info": [
-      ("Library", "Sample Container Barcode", "container__barcode"),
-      ("Library", "Sample Container Coordinates", "coordinates"),],
+      # prefill_sheet_name, template_column, queryset_column
+      ("LibraryQC", "Library Container Barcode", "??"),
```
UlysseFG, line 121: Library QC should not change the library_preparation template.

backend/fms_core/migrations/0038_v3_9_0.py
```diff
@@ -0,0 +1,62 @@
...
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+        admin_user_id = admin_user.id
+
+        reversion.set_comment("Create objects related to protocol Library Quality Control")
+        reversion.set_user(admin_user)
+
+        # Create PropertyType and Protocols
+        PROPERTY_TYPES_BY_PROTOCOL = {
+            # note: property type values must be unique so these all need the Library prefix
```
UlysseFG, line 25: This is the case but Sebastian is changing that in his PR for library conversion. You can keep the same names. When Sebastian merge his branch, the unique constraint will be with both the name and the protocol id so they'll be unique.

backend/fms_core/templates.py
```diff
@@ -120,6 +121,29 @@
...
+        'name': 'LibraryQC',
+        'headers': ['Library Name', 'Library Container Barcode', 'Library Container Coord', 'Initial Volume (uL)',
+                    'Measured Volume (uL)', 'Volume Used (uL)', 'Strandedness', 'Library size (bp)', 'Concentration (nM)',
+                    'Concentration (ng/uL)', 'NA Quantity (ng)', 'Quality Instrument', 'Quality Flag',
+                    'Quantity Instrument', 'Quantity Flag', 'QC Date (YYYY-MM-DD)', 'Comment'],
+      },
+  ],
+  "prefill info": [
+      # TODO queryset columns
+      ("LibraryQC", "Library Name", "?"),
```
UlysseFG, line 137: This part is a bit magic. Template are based off different models the library and sample prefill assume we'll fill using information from the sample (container move and rename take them from container). This could be parametrized further at some point (we may have to here for the strandedness and library_size).

backend/fms_core/templates.py
```diff
@@ -120,6 +121,29 @@
...
+      ("LibraryQC", "Library Container Barcode", "?"),
```
UlysseFG, line 138: That would be "container__barcode"

backend/fms_core/templates.py
```diff
@@ -120,6 +121,29 @@
...
+      ("LibraryQC", "Library Container Coord", "?"),
```
UlysseFG, line 139: "coordinates"

backend/fms_core/templates.py
```diff
@@ -120,6 +121,29 @@
...
+      ("LibraryQC", "Initial Volume (uL)", "?"),
+      ("LibraryQC", "Strandedness", "?"),
```
UlysseFG, line 141: that would be more complicated i will have to think this one over ... The path would be "derived_sample_not_pool__library__strandedness" but the pooling thing may interfere So we may need to make a special accessor to this and the library_size which is also behind a many-to-many relationship.

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
```
UlysseFG, line 1: You can get rid of this. It does not belong in the repo.

backend/fms_core/templates.py
```diff
@@ -120,6 +121,29 @@
...
+                    'Quantity Instrument', 'Quantity Flag', 'QC Date (YYYY-MM-DD)', 'Comment'],
+      },
+  ],
+  "prefill info": [
+    # TODO queryset columns
+    ("LibraryQC", "Library Name", "name"),
+    ("LibraryQC", "Library Container Barcode", "container__barcode"),
+    ("LibraryQC", "Library Container Coord", "coordinates"),
+    ("LibraryQC", "Initial Volume (uL)", "volume"),
+    # ("LibraryQC", "Strandedness", "derived_samples__not_pool__library__strandedness"),   (doesn't work...)
```
UlysseFG, line 141: You can remove both for good no point keeping these commented out ;).

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def process_row_inner(self, sample_container, measures, process, process_measurement, process_measurement_properties):
+         # Get the library sample that was checked
+        source_sample_obj, self.errors['container'], self.warnings['container'] = \
+            get_sample_from_container(barcode=sample_container['container_barcode'], coordinates=sample_container['container_coord'])
+
+        if source_sample_obj is None:
+            self.errors['sample_source'] = 'Library sample for QC was not found at the specified container.'
```
sebastianballesteros, line 25: You might wanna include the container name or barcode to provide more information to the user. i.e. `at the specified container {container.barcode}` or something like that

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,116 @@
...
+                self.errors['concentration'] = 'A concentration in either nM or uL must be specified.'
+            else:
+                molecular_weight = source_sample_obj.derived_sample_not_pool.library.molecular_weight_approx
+                concentration = convert_concentration_from_nm_to_ngbyul(concentration, molecular_weight, library_size)
+                if concentration is None:
+                    self.errors['concentration'] = 'Concentration could not be converted from nM to ng/uL'
+                else:
+                    process_measurement_properties["Library Concentration"]['value'] = concentration
+
+        # update the sample volume and concentration
```
sebastianballesteros, line 65: Maybe is a good idea to check if there are no errors before continuing

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if measures['volume_used'] < 0:
+            self.errors['Volume used should be a positive number.']
+
+        final_volume = measures['measured_volume'] if measures['measured_volume'] is not None else measures['initial_volume']
+        final_volume -= measures['volume_used']
+
+        # concentration
+        if measures['concentration_nm'] is not None and measures['concentration_uL'] is not None:
+            self.warning['concentration'] = 'The concentration should be specified in nM or uL, not both. nM will be ignored.'
```
UlysseFG, line 45: We try not to make choices for the users. The ultimate goal in Freezeman is not to make the user life easier it is to make sure the data is correctly saved. The user QoL comes second. In this case maybe the user actually converted the nM (real measurement) into ng/ul which would have come at a conversion cost. Maybe he trust his nM value more than the converted value ... In any case we want to leave the decision in his hand. It would be better to raise an error and refuse the template and let the user decide. If you want to get more fancy you can actually convert the nm to ng/ul and check if both values and equal if they are you can downgrade the error to a warning to let the user know they are doing more work than necessary.

.vscode/launch.json
ckostiw: I'll add that to .gitignore then

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,104 @@
...
+from fms_core.templates import LIBRARY_QC_TEMPLATE
+from collections import defaultdict
+from .._utils import float_to_decimal_and_none, input_to_date_and_none
+
+PROPERTIES_STARTING_INDEX = 5 #TODO verify that this index is correct
+
+# {{TEMPLATE PROPERTY NAME : DB PROPERTY NAME}
+TEMPLATE_PROPERTY_MAPPING = {
+    "Measured Volume (uL)": "Library Measured Volume",
+    "Volume Used (uL)": "Library Volume Used",
```
UlysseFG, line 16: Volume used is stored at the processMeasurement level (it is duplicating information to store it too). This should not be a property. On the other hand you are not storing the value of the library size which is not stored in a permanent storage. You should add the library size here instead. We save the library size in the library model but a subsequent QC would override the value and the historical value would not be available to users anymore.

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,104 @@
...
+            #     'quality_flag': row_data['Quality Flag'],
+            #     'quantity_instrument': row_data['Quantity Instrument'],
+            #     'quantity_flag': row_data['Quantity Flag'],
+            #     'qc_date': input_to_date_and_none(row_data['QC Date (YYYY-MM-DD)']),
+            #     'comment': row_data['Comment']
+            # }
+
+            process_measurement = {
+                'execution_date': input_to_date_and_none(row_data['QC Date (YYYY-MM-DD)']),
+                'volume_used': float_to_decimal_and_none(row_data['Volume Used (uL)']),
```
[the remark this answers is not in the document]
UlysseFG: I would remove the one from measures since that value is not intended to be saved as a property.

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,104 @@
...
+                'container_coord': row_data['Library Container Coord']
+            }
+
+            measures = {
+                'initial_volume': float_to_decimal_and_none(row_data['Initial Volume (uL)']),
+                'measured_volume': float_to_decimal_and_none(row_data['Measured Volume (uL)']),
+                'volume_used': float_to_decimal_and_none(row_data['Volume Used (uL)']),
+                'concentration_nm': float_to_decimal_and_none(row_data['Concentration (nM)']),
+                'concentration_uL' : float_to_decimal_and_none(row_data['Concentration (ng/uL)']),
+                'library_size': float_to_decimal_and_none(row_data['Library size (bp)']).to_integral_value()
```
UlysseFG, line 72: Here i think we have to check this with the lab. The model for library was made before we really had any values to store into the library_size, but if i recall the library_size is an average which mean that even if it actually represent an integral size, it would use decimals. We need to check with Janick or Tony if they have decimals to store (since it's used for conversion we'll take as much precision as possible).

backend/fms_core/template_importer/importers/library_qc.py
ckostiw: Okay. I don't see a property type that I could use for that though. Should I add a new "Library Size" property in my migration?

backend/fms_core/template_importer/importers/library_qc.py
ckostiw: Following our discussion with the lab, I will leave this as a decimal with zero decimal digits. I will add a `digits` parameter to the `float_to_decimal_and_none` function so that I can specify zero digits after the decimal.

backend/fms_core/coordinates.py
```diff
@@ -77,13 +77,14 @@ def validate_and_normalize_coordinates(coords: str, spec: CoordinateSpec) -> str
...
     parent using a queryset, assuming that the queried model has a coordinates
     field which specifies possibly-overlapping item locations.
     """
-    existing = queryset.exclude(pk=obj.pk).get(coordinates=obj.coordinates)
-    raise CoordinateError(f"Parent container {parent} already contains {obj_type} {existing} at "
-                          f"coordinates {obj.coordinates}")
+    error = None
+    exists = queryset.exclude(pk=obj.pk).filter(coordinates=obj.coordinates).exists()
+    if exists:
+        error = f"Parent container {parent} already contains {obj_type} at coordinates {obj.coordinates}"
```
UlysseFG, line 89: You removed the object identifier from the error message. You would have to make a second query here to get it.

backend/fms_core/models/container.py
```diff
@@ -87,14 +87,9 @@ def add_error(field: str, error: str):
                 if not errors.get("coordinates") and not errors.get("location") \
                         and not parent_spec.coordinate_overlap_allowed:
                     # Check for coordinate overlap with existing child containers of the parent
-                    try:
-                        check_coordinate_overlap(self.location.children, self, self.location)
-                    except CoordinateError as e:
-                        add_error("coordinates", str(e))
-                    except Container.DoesNotExist:
```
UlysseFG, line 94: Despite me disapproving with that behaviour some operations are expected to be performed (and are faster) using exception catching in python. You may have to accept that some things are done that way. There is as far as i know always a clean way around but it might force some extra operations.

backend/fms_core/models/container.py
```diff
@@ -87,14 +87,9 @@ def add_error(field: str, error: str):
                 if not errors.get("coordinates") and not errors.get("location") \
                         and not parent_spec.coordinate_overlap_allowed:
                     # Check for coordinate overlap with existing child containers of the parent
-                    try:
-                        check_coordinate_overlap(self.location.children, self, self.location)
```
UlysseFG, line 91: This is a risky change just to make your debugger happy. I would recommend removing the change from this PR and moving it to its own branch for future inclusion and creating a ticket so it gets properly tested when we have some time.

backend/fms_core/template_importer/importers/library_qc.py
UlysseFG: You create a new one in the migration yeah.

backend/fms_core/services/library.py
```diff
@@ -143,3 +143,23 @@ def convert_library(process, platform, sample_source, container_destination, coo
             errors.append(e)
 
     return library_destination, errors, warnings
+    
+def set_library_size(sample, library_size = None):
```
UlysseFG, line 147: I would create a more multipurpose function here ... update_library(sample, library_data) maybe ... Library_data would hold a dictionary of values to update in the model. That way you can omit attributes if you don't want to change them or pass None if you want to set it to None. We are likely to have to set many values in the library over the time and it would be easier to have a more generic call instead. Maybe kwargs could be used too might be more elegant and easy to use.

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,104 @@
...
+
+            process_measurement_properties = self.preloaded_data['process_properties']
+
+             #Populate process properties
+            for i, (key, val) in enumerate(row_data.items()):
+                if key in TEMPLATE_PROPERTY_MAPPING.keys():
+                    process_measurement_properties[TEMPLATE_PROPERTY_MAPPING[key]]['value'] = val
+
+            sample_container = {
+                'container_barcode': row_data['Library Container Barcode'],
```
UlysseFG, line 62: I've applied str_cast_and_normalize in another PR to row_data values to make sure the input is a string for string inputs. It could be a good idea to do it right away instead of fixing this later.

backend/fms_core/models/container.py
ckostiw: I've reverted the commit

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,116 @@
...
+        if measures['initial_volume'] is None and measures['measured_volume'] is None:
+            self.errors['measured_volume'] = 'Either measured or initial volume must be specified.'
+
+        if measures['volume_used'] is None:
+            self.errors['volume_used'] = 'Volume used must be specified.'
+
+        if measures['volume_used'] < 0:
+            self.errors['Volume used should be a positive number.']
+
+        final_volume = measures['measured_volume'] if measures['measured_volume'] is not None else measures['initial_volume']
```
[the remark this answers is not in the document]
ckostiw: I'm not sure the delta approach always works. It depends on why the volume is different and if that change in volume was recorded in freezeman or not.

It works if the initial volume matches the volume in freezeman, meaning that some volume was removed and the change was not tracked in freezeman:

```
initial: 	100
measured:	80
volume used:	10
freezeman:	100*. (current volume stored in freezeman)

final: 		100* - (100 - 80) - 10 = 70
```

If the change *was* tracked in freezeman, then the delta will end up subtracting too much. Here, 20 uL was removed in a step that was tracked by freezeman, so the freezeman volume is 80uL. With the delta calculation, we end up subtracting that 20uL again, and end up with the wrong volume:

```
initial:	100
measured:	80
volume used:	10
freezeman	80* (change was already recorded in freezeman)

final:		80* - (100 - 80) - 10 = 50	
```

Am I missing something?

backend/fms_core/coordinates.py
ckostiw: This change was reverted

backend/fms_core/migrations/0039_v3_9_0.py
```diff
@@ -0,0 +1,51 @@
...
+        reversion.set_comment("Add property type 'Library Size' for library QC")
+        reversion.set_user(admin_user)
+
+        # Add a "Library Size" property type for library QC process measurements,
+        # associated with the Sample Quality Control protocol
+
+        protocol_content_type = ContentType.objects.get_for_model(Protocol)
+        sample_qc_protocol = Protocol.objects.get(name='Sample Quality Control')
+       
+        property_type = PropertyType.objects.create(name="Library Size",
```
UlysseFG, line 29: You have to create all the properties not only the ones that are different from sample QC even if they share the same name.

backend/fms_core/services/library.py
```diff
@@ -143,3 +144,71 @@ def convert_library(process, platform, sample_source, container_destination, coo
             errors.append(e)
 
     return library_destination, errors, warnings
+
+def set_library_size(sample, library_size = None):
```
UlysseFG, line 148: You can remove the set_library_size function if you make the update library function.

backend/fms_core/services/library.py
```diff
@@ -143,3 +144,71 @@ def convert_library(process, platform, sample_source, container_destination, coo
...
+
+    if sample is None:
+        errors.append('Missing sample')
+    elif not sample.is_library:
+        errors.append('Sample {sample.name} is not a library')
+    else:
+        try:
+            library = sample.derived_sample_not_pool.library
+
+            if 'library_type' in kwargs:
```
UlysseFG, line 180: this could be a very long list of arguments... i would make that more compact by using  the kwargs dict in a model update like in the update sample viewset. You can add some type checking or validation in a compact copy from kwargs to a local dict for the update.

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,99 @@
...
+from fms_core.templates import LIBRARY_QC_TEMPLATE
+from fms_core.template_importer.row_handlers.library_preparation import LibraryRowHandler, LibraryBatchRowHandler
+from fms_core.template_importer.row_handlers.library_qc.library_qc import LibraryQCRowHandler
+
+
+PROPERTIES_STARTING_INDEX = 5 #TODO verify that this index is correct
+
+# {{TEMPLATE PROPERTY NAME : DB PROPERTY NAME}
+TEMPLATE_PROPERTY_MAPPING = {
+    "Measured Volume (uL)": "Measured Volume",
```
UlysseFG, line 17: Library_size is missing.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def process_row_inner(self, sample_container, measures, process, process_measurement, process_measurement_properties):
+         # Get the library sample that was checked
+        source_sample_obj, self.errors['container'], self.warnings['container'] = \
+            get_sample_from_container(barcode=sample_container['container_barcode'], coordinates=sample_container['container_coord'])
+
+        if source_sample_obj is None:
+            self.errors['sample_source'] = f"Library sample for QC was not found in container {sample_container['container_barcode']} at {sample_container['coordinates']}."
```
UlysseFG, line 23: use either library or sample to describe samples, for the user they are 2 different things and are clear enough. combining them is likely to bring confusion.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -0,0 +1,146 @@
...
+                comment=process_measurement['comment'],
+            )
+
+         # Create process measurement's properties
+        if process_measurement_obj:
+            properties_obj, self.errors['properties'], self.warnings['properties'] = create_process_measurement_properties(
+                process_measurement_properties,
+                process_measurement_obj)
+
+
```
UlysseFG, line 132: That is a lot of empty lines ...

backend/fms_core/migrations/0039_v3_9_0.py
ckostiw: Ah, okay. I misunderstood. I thought we wanted to use the same protocol and properties for sample QC and library QC.

backend/fms_core/template_importer/importers/library_qc.py
```diff
@@ -0,0 +1,99 @@
...
+    SHEETS_INFO = LIBRARY_QC_TEMPLATE['sheets info']
+
+    def __init__(self):
+        super().__init__()
+        self.properties_starting_index = PROPERTIES_STARTING_INDEX
+        self.initialize_data_for_template()
+
+    def initialize_data_for_template(self):
+        #Get protocol for SampleQC, which is used for samples and libraries
+        protocol = Protocol.objects.get(name='Sample Quality Control')
```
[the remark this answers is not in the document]
ckostiw: Fixed

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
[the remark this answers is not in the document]
UlysseFG: It all come down to Initial being correct ... if someone measure 80 while Freezeman reported 100 it means there was already 20 uL that were incorrectly stored and are now corrected ... If initial is incorrect it is indeed going to lead to values that are wrong.

backend/fms_core/templates.py
```diff
@@ -147,6 +148,27 @@
       ("Library", "Sample Container Coordinates", "coordinates"),],
 }
 
+LIBRARY_QC_TEMPLATE = {
+  "identity": {"description": "Template to add library QC", "file": static("submission_templates/Library_QC_v3_9_0.xlsx")},
```
UlysseFG, line 152: Small stuff, the front end uses a different icon if you include quality in the description. It would be better too to match the sample quality control description. Use something like : Template to perform library quality control

UlysseFG: Prefiling the template seem to be killing the drop downs. Try removing all content and formats from the "empty" lines between the Title and the header in the template. Hopefully this is going to fix it.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -15,43 +15,69 @@ def __init__(self):
...
 
     def process_row_inner(self, sample_container, measures, process, process_measurement, process_measurement_properties):
-         # Get the library sample that was checked
-        source_sample_obj, self.errors['container'], self.warnings['container'] = \
-            get_sample_from_container(barcode=sample_container['container_barcode'], coordinates=sample_container['container_coord'])
 
+        barcode = sample_container['container_barcode']
+        coordinates = sample_container['container_coord']
+
+        if barcode is None:
```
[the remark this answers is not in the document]
ckostiw: I had several exceptions caused by None values (eg. comparing a volume to None) and that ruins the validation and just displays an unhelpful error message in the freezeman page, so I decided to be more aggressize and do my None checks once, rather than having to have None checks all over the place. These checks are to avoid unexpected exceptions blowing up the validation.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -15,43 +15,69 @@ def __init__(self):
...
+
+        # If library sample cannot be found then bail.
         if source_sample_obj is None:
-            self.errors['library'] = f"Library sample for QC was not found in container {sample_container['container_barcode']} at {sample_container['coordinates']}."
+            self.errors['library'] = f"Library sample for QC was not found in container ({barcode}) at ({coordinates})"
+            return
 
         if not source_sample_obj.is_library:
-            self.errors['sample'] = f'The sample at the specified location is not a library: ${source_sample_obj.name}'
+            self.errors['sample'] = f'The sample at {barcode}@{coordinates} is not a library ({source_sample_obj.name})'
```
UlysseFG, line 40: move the name after "The sample"  maybe to make it flow more naturally  ?

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -15,43 +15,69 @@ def __init__(self):
...
+        if initial_volume is None:
+            self.errors['initial_volume'] = 'Initial volume must be specified'
+        if measured_volume is None:
+            self.errors['measured_volume'] = 'Measured volume must be specified'
+        if volume_used is None:
+            self.errors['volume_used'] = 'Volume used must be specified'
+        if sample_volume is None:
+            self.errors['library volume'] = 'Library volume is missing'
+
+        if not None in (initial_volume, measured_volume, volume_used, sample_volume):
```
UlysseFG, line 58: This is an ugly test ... you should check if there is errors instead ... Its much easier to understand and you are less likely to let it proceed because you missed something.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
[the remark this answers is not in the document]
ckostiw: Can I add a check in the service to explicitly check for a None value for the barcode?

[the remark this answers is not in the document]
UlysseFG: Yeah this could improve the actual error message returned and prevent a useless call to get().

UlysseFG (approved):

PR #291, opened 2022-05-24
Patch sample update

UlysseFG:


backend/fms_core/template_importer/row_handlers/sample_update/sample.py
```diff
@@ -28,11 +28,11 @@ def process_row_inner(self, delta_volume, sample, sample_updated, process_measur
             if sample_updated['volume']:
                 volume_used = sample_to_update.volume - sample_updated['volume']
```
sebastianballesteros, line 29: In the case when they want to increase the volume. i.e. from 200 to 220 for example. The volume_used would be negative (200-220). If that's the desired behaviour then perfect. Just want to check I understand.

UlysseFG: yep when increasing the volume there should be a negative volume_used. This is not normally possible in real protocols but update can be used to fix some small volume discrepancies.

sebastianballesteros (approved):

PR #294, opened 2022-05-30
Added a direct call to confirm() in Range filters Done buttons.

UlysseFG:


ckostiw (approved):
Phew!! The code for the Ant Design table component is incredibly convoluted (as, I think, all table components end up being), so I had to work hard to figure out why your fix solved the problem.

Here is the code for the `confirm` function, from the Ant Design table component sources (the function name is actually `doFilter` but gets named `confirm` in our code):

https://github.com/ant-design/ant-design/blob/cfd25f9650f65a320b9cabfbf7eaee5e0ff0c083/components/table/hooks/useFilter/FilterDropdown.tsx#L251
```
const doFilter = ({ closeDropdown } = { closeDropdown: true }) => {
    if (closeDropdown) {
      triggerVisible(false);
    }
    internalTriggerFilter(getFilteredKeysSync());
  };
```
Before your fix, we were passing `doFilter` directly to onClick, resulting in `doFilter` receiving the mouse click event as an argument (from React). This meant that `doFilter` was receiving an object (the event) that does not include a `closeDropdown` property. So, the check for `closeDropdown` was false, and `triggerVisible` was not called.
With your change, we are calling `confirm` with no arguments, so the default argument is used `{ closeDropdown: true }`, and `triggerVisible` gets called, closing the date picker.

PR #295, opened 2022-06-01
Fix Issue #939

nafiz1001:
Fixes issue #939 by making units one decimal place after the decimal point.

https://206.12.92.46/issues/939

### DIFF
--- a/frontend/src/components/samples/details/SampleDetailsContent.js
+++ b/frontend/src/components/samples/details/SampleDetailsContent.js
@@ -41,6 +41,7 @@ import {
   withIndex
 } from "../../../utils/withItem";
 import ExperimentRunsListSection from "../../shared/ExperimentRunsListSection";
+import { NumberFormat } from "../../../utils/functions";
 
 const { Title, Text } = Typography;
 const { TabPane } = Tabs;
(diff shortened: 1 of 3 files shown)


frontend/src/utils/functions.js
```diff
@@ -1 +1,6 @@
 export const constVal = x => () => x;
+
+export function NumberFormat(number, afterDecimal = 0, round = true) {
+    const regex = RegExp(`(^[0-9]+\.[0-9]{0,${afterDecimal}})|(^[0-9]+)`, "g")
+    return number.toFixed(afterDecimal + (round ? 0 : 1)).match(regex)[0]
```
UlysseFG, line 5: This is heavy handed for cases you want to round. I would test on the round param and do the regexp matching only when needed to truncate.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -125,8 +126,8 @@ const SampleDetailsContent = ({
   let processMeasurements = []
   let experimentRunsIDs = []
   const library = librariesByID[id]
-  const quantity = library && library.quantity_ng ? parseFloat(library.quantity_ng).toFixed(3) : undefined
-  const concentration_nm = library && library.concentration_nm ? parseFloat(library.concentration_nm).toFixed(3) : undefined
+  const quantity = library && library.quantity_ng ? NumberFormat(parseFloat(library.quantity_ng), 1) : undefined
+  const concentration_nm = library && library.concentration_nm ? NumberFormat(parseFloat(library.concentration_nm), 1) : undefined
```
UlysseFG, line 130: I am not sure all numeric values should share the same precision (this would need to be checked with the lab. Also if we apply a precision in one list it should be done everywhere. The library list and project sample list should also share the same changes.

PR #296, opened 2022-06-03
Fixing the prefilled template not being openable in excel.

UlysseFG:


ckostiw (approved):
We can use keep_links, if we need it. Only keep_vba causes an error in excel

PR #297, opened 2022-06-06
fix centering bug in the lineage

nafiz1001:
fix centering bug in the lineage and a few others.

i only discover bugs when I'm confident about my code :'(


frontend/src/components/samples/details/SampleDetailsLineage.js
```diff
@@ -67,6 +68,30 @@ const SampleDetailsLineage = ({
     marginy: 50
   }
 
+  const adjustGraphData = (graphData) => {
+    const { nodes, links } = graphData
+    let dx = 0
+    if (nodes.length > 0 && sample.id !== undefined) {
```
nafiz1001, line 74: hmm... i should do `sample?.id` just to be safe.

ckostiw (approved):

PR #298, opened 2022-06-07
Cast and normalize template input strings

UlysseFG:


backend/fms_core/template_importer/importers/container_creation.py
```diff
@@ -14,12 +16,12 @@ def import_template_inner(self):
         for row_id, row_data in enumerate(containers_sheet.rows):
             container = {
                 'kind': row_data['Container Kind'],
```
sebastianballesteros, line 18: Do we wanna `str_cast_and_normalize` Container Kind?

UlysseFG: If this is a string we should i think. This is unlikely someone will input a number in there but better safe than sorry.

ckostiw:
```suggestion
'kind': str_cast_and_normalize(row_data['Container Kind']),
```

ckostiw:
```suggestion
                'kind': str_cast_and_normalize (row_data['Container Kind']),
```

UlysseFG: Oh right ... i should apply that !

backend/fms_core/template_importer/importers/sample_submission.py
```diff
@@ -19,49 +20,49 @@ def import_template_inner(self):
...
-                'alias': row_data['Individual Alias'],
-                'sex': row_data['Sex'],
-                'pedigree': row_data['Pedigree'],
-                'taxon': row_data['NCBI Taxon ID #'],
-                'cohort': row_data['Cohort'],
+                'name': str_cast_and_normalize(row_data['Individual ID']),
+                'alias': str_cast_and_normalize(row_data['Individual Alias']),
+                'sex': str_cast_and_normalize(row_data['Sex']),
+                'pedigree': str_cast_and_normalize(row_data['Pedigree']),
+                'taxon': int(row_data['NCBI Taxon ID #']),
```
ckostiw, line 36: int(None) and int('') will throw an exception. Should the int conversion be put in a function that handles the exception and returns None?

ckostiw (changes requested):

backend/fms_core/template_importer/importers/sample_submission.py
UlysseFG: yeah i'll add something to the utils for that...

ckostiw (approved):

PR #299, opened 2022-06-08
Add documentation system for python codes.

nafiz1001:


backend/.gitignore
```diff
@@ -21,3 +21,5 @@ htmlcov/
 ~$*
 .DS_Store
 venv
+
+docs
```
UlysseFG, line 25: Not sure we want that ... I was hoping to store the version current docs in the repo instead of having to force someone to install the repo in order to generate the docs.

nafiz1001: wouldn't that double the size of our repo?

backend/fms_core/management/commands/createdocs.py
```diff
@@ -0,0 +1,35 @@
...
+
+class Command(BaseCommand):
+    help = "Generates documentation for python modules"
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument(
+            "--output",
+            help="The directory to output generated documents to (default: ./docs)",
+            type=str,
+            default="./docs"
```
UlysseFG, line 12: You might want to register the default as a path relative to the repo root. This default value may lead to an output being created at undesirable position depending on the location of the execution. Maybe use the settings.py file to set the default at the App level.

PR #304, opened 2022-06-14
Store template file

UlysseFG:
TODO : 
Add the link to the stored template in the process detail page in the frontend.


backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -46,14 +57,30 @@ def import_template(self, file, format, dry_run):
                         reversion.set_comment("Template import - dry run")
                     transaction.set_rollback(True)
                 else:
-                    self.import_template_inner()
+                    # Save template on disk with modified name (append timestamp and id) and insert path in imported_file
+                    if user is not None:
+                        os.environ["TZ"] = settings.LOCAL_TZ
+                        time.tzset()
+                        new_file_name = f"{file_name}_{time.strftime('%Y-%m-%d_%H:%M:%S')}_{user.username}{file_format}"
```
ckostiw, line 64: The `:` character is illegal on OS X and Windows, so a different separator is needed. 
FYI:
The following characters are invalid as file or folder names on Windows using NTFS:
/ ? < > \ : * | " and any character you can type with the Ctrl key

The only illegal character for file and folder names in Mac OS X is the colon ":"

The only illegal character for file and folder names in Linux is "/"

backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -1,12 +1,19 @@
...
 import reversion
+import os
 
 from ..sheet_data import SheetData
 from .._utils import blank_and_nan_to_none
 from fms_core.utils import str_normalize
 
+from fms_core.models import ImportedFile
+
+UPLOAD_PATH = os.path.join(settings.MEDIA_ROOT, 'uploads/')
```
ckostiw, line 15: Maybe this should be defined in settings.py? Also, I'd be tempted to name this UPLOADED_TEMPLATES_PATH, and name the folder under /media "templates", just in case freezeman has other types of uploads in the future.

UlysseFG: Excellent suggestions ... yep

backend/fms_core/template_importer/importers/_generic.py
UlysseFG: My bad some dash should do it.

backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -46,14 +57,30 @@ def import_template(self, file, format, dry_run):
...
+                            self.base_errors.append(err)
+                    self.import_template_inner()                            
                     reversion.set_comment("Template import")
 
             # Add processed rows with errors/warnings/diffs to self.previews_info list
             for sheet in list(self.sheets.values()):
                 preview_info = sheet.generate_preview_info_from_rows_results(rows_results=sheet.rows_results)
                 self.previews_info.append(preview_info)
 
+            # Save the template on the server if the template is valid.
```
ckostiw, line 78: I wonder if this should be wrapped in a try block. What kind of response will the user get if an exception occurs? Has the user's data been commited to the DB, or will it roll back if an exception is raised?

backend/fms_core/template_importer/importers/sample_qc.py
```diff
@@ -44,6 +44,11 @@ def initialize_data_for_template(self):
     def import_template_inner(self):
         sample_qc_sheet = self.sheets['SampleQC']
 
+        # Add the template to the process
+        if self.imported_file is not None:
```
ckostiw, line 48: Service?

backend/fms_core/template_importer/importers/sample_selection_qpcr.py
```diff
@@ -42,6 +42,11 @@ def initialize_data_for_template(self):
     def import_template_inner(self):
         sample_qpcr_sheet = self.sheets['Samples']
 
+        # Add the template to the process
+        if self.imported_file is not None:
```
ckostiw, line 46: Service?

backend/fms_core/template_importer/importers/sample_update.py
```diff
@@ -23,6 +23,11 @@ def initialize_data_for_template(self):
     def import_template_inner(self):
         sampleupdate_sheet = self.sheets['SampleUpdate']
 
+        # Add the template to the process
+        if self.imported_file is not None:
```
ckostiw, line 27: Service?

backend/fms_core/template_importer/importers/transfer.py
```diff
@@ -23,6 +23,11 @@ def initialize_data_for_template(self):
     def import_template_inner(self):
         sheet = self.sheets['SampleTransfer']
 
+        # Add the template to the process
+        if self.imported_file is not None:
```
ckostiw, line 27: Service?

backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -16,12 +23,16 @@ def __init__(self):
...
+        self.format = None
+        self.imported_file = None
         self.sheets = {}
         self.previews_info = []
 
-    def import_template(self, file, format, dry_run):
+    def import_template(self, file, dry_run, user = None):
         self.file = file
-        self.format = format
+        file_name, file_format = os.path.splitext(file.name)
```
ckostiw, line 33: I can't believe the python api uses this typo for a function name. It should be "splittext" or "split_text". :)

UlysseFG: I think it is split_ext as in split extension. But yeah it is not that friendly

backend/fms_core/template_importer/importers/extraction.py
```diff
@@ -27,6 +27,11 @@ def initialize_data_for_template(self):
     def import_template_inner(self):
         sheet = self.sheets['ExtractionTemplate']
 
+        # Add the template to the process
+        if self.imported_file is not None:
+            self.preloaded_data['process'].imported_template_id = self.imported_file.id
+            self.preloaded_data['process'].save()
```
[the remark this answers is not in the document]
UlysseFG: DB stuff in services is not a religion ... i do not think it deserves a service since there is really nothing here to reuse. Service are more to buffer the internal core from input from templates and to gather commonly used functionality for complex objects.

backend/fms_core/template_importer/importers/_generic.py
UlysseFG: Lets say we do not have enough disk space to receive the file ? I think it will return a 500 internal server error. I think it will force a rollback of the whole request. I think we work with a system that wrap each request in a transaction. I am not sure though i will do some testing (writing to the same file name and see what happens). Having a try block will likely not prevent the issue but at least we can inform the user on the issue (the default message is likely very unhelpful).

sebastianballesteros (approved):

PR #305, opened 2022-06-15
add table of Process Measurements in ProcessProperties

nafiz1001:
Resolves ticket: https://206.12.92.46/issues/941

* Add table in Process->Properties page that shows Process Measurements
* Add table in Process Measuerments -> Properties that shows Process properties


frontend/src/modules/processMeasurements/actions.js
```diff
@@ -31,6 +32,21 @@ export const list = (options) => async (dispatch, getState) => {
...
+    filters = serializeFilterParams(filters, PROCESS_MEASUREMENT_FILTERS)
+    const ordering = serializeSortByParams(sortBy)
+    const options = { limit, offset, ordering, ...filters}
+
+    return await dispatch(networkAction(LIST_FILTER,
+        api.processMeasurements.list(options, abort),
+        { meta: { ...options } }
+    ));
+};
+
```
nafiz1001, line 49: Should I look into creating a High-Order Function so we can replicate this for each models?
Not sure how often we will repeat this though.

nafiz1001: Not now 👎🏼

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -76,15 +76,11 @@ const ProcessMeasurementsDetailContent = ({
...
-                  <>
-                  <Title level={3} style={{marginTop: '20px'}}>Properties</Title>
-                    <ProcessProperties
-                        propertyIDs={processMeasurement.properties}
-                        protocolName={protocolsByID[processMeasurement.protocol]?.name}
-                    />
-                  </>
-               }
+              <Title level={3} style={{marginTop: '20px'}}>Properties</Title>
+              <ProcessProperties
```
sebastianballesteros, line 80: It would be useful to display the Process properties in the Process Measurement Detail Content page as well. Like that we have both: Process' properties and process measurement's properties in both pages (process measurement detail content and process detail content).

frontend/src/components/shared/ProcessProperties.js
```diff
@@ -46,11 +49,19 @@ const ProcessProperties = ({
...
+      {
+        propertiesAreLoaded &&
+          <Table columns={columns} dataSource={data} pagination={false} className="table-min-width" />
+      }
+
+      <Divider/>
+
+      {
+        showProcessMeasurements && <>
+          <Title level={5} style={{ marginTop: "20px" }}>Measurements</Title>
```
nafiz1001, line 62: switch from `Measurements` to `Protocol Applications`?

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -8,12 +8,13 @@ const {Title} = Typography;
 import AppPageHeader from "../AppPageHeader";
 import PageContent from "../PageContent";
 import ProcessProperties from "../shared/ProcessProperties";
+import ProcessAssociatedMeasurements from "../shared/ProcessAssociatedMeasurements"
 import TrackingFieldsContent from "../TrackingFieldsContent";
-import {listProcesses, listPropertyValues} from "../../modules/experimentRuns/actions";
+import {listProcesses, listPropertyValues, listProcessProperties} from "../../modules/experimentRuns/actions";
 import {download as templateDownload} from "../../modules/importedFiles/actions";
```
ckostiw, line 14: At runtime I get an error saying `download` does not exist in the importedFiles actions module. Maybe we could remove this import as part of your PR?

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -31,7 +33,10 @@ const ProcessMeasurementsDetailContent = ({
   samplesByID,
   usersByID,
```
ckostiw, line 34: usersById is not used

frontend/src/modules/experimentRuns/actions.js
```diff
@@ -119,7 +119,39 @@ export const listTemplateActions = () => (dispatch, getState) => {
     return dispatch(networkAction(LIST_TEMPLATE_ACTIONS, api.experimentRuns.template.actions()));
 };
 
+export const listProcessProperties = (id) => async (dispatch, getState) => {
+    if (getState().propertyValues.isFetching)
+        return;
 
+    const { itemsByID, propertyValuesByID } = getState().processes
```
ckostiw, line 126: propertyValuesByID is not part of the redux state and the var is always undefined. I think you want `propertyValues.itemsByID`.

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
nafiz1001: I wrote that? 🤔

frontend/src/components/processes/ProcessDetailContent.js
nafiz1001: I feel like it makes more sense to make it a separate PR. And then I merge your PR after your PR is merged.

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
ckostiw: Ah, no I guess not. But, it's nice to follow the Boy Scout Principal - leave the code cleaner than when you found it (if the cleanup is easy).

frontend/src/modules/experimentRuns/actions.js
nafiz1001: I think I resolved it https://github.com/c3g/freezeman/pull/305/commits/ce584414740e3b3fdc0e7aa87a8f2fef2ac3ffdf?

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -55,16 +51,8 @@ const ProcessDetailContent = ({
...
-      // Need to be queried as a string, not as an array in order to work with DRF filters
-      const processIDSAsStr = [id].concat(process.children_processes).join()
-      listProcesses({id__in:processIDSAsStr});
-    }
-
-    if (isLoaded && childrenProcessesAreLoaded && !allPropertiesAreLoaded) {
-      // Need to be queried as a string, not as an array in order to work with DRF filters
-      const processIDSAsStr = [id].concat(process.children_processes).join()
-      listPropertyValues({ object_id__in: processIDSAsStr, content_type__model: "process" });
+    if (isLoaded && !isProcessPropertiesLoaded(processesByID, propertyValuesByID, id)) {
```
ckostiw, line 54: Any code which causes the redux state associated with the component to change needs to be wrapped in `useEffect`. Otherwise, we are modifying state during render, and that generates a react error.

frontend/src/components/shared/ProcessAssociatedMeasurements.js
```diff
@@ -0,0 +1,121 @@
...
+  id,
+}) => {
+  const filterKey = PROCESS_MEASUREMENT_FILTERS.process.key
+
+  const measurementsWithMissingProperties = processMeasurements.filter((id) => id in processMeasurementsByID)
+                                                   .map((id) => processMeasurementsByID[id])
+                                                   .filter((p) => !hasEveryProperties(p, propertyValuesByID))
+
+  if (measurementsWithMissingProperties.length > 0) {
+    listPropertyValues({ object_id__in: measurementsWithMissingProperties, content_type__model: "processmeasurement" })
```
ckostiw, line 81: Needs to be wrapped in useEffect()

frontend/src/modules/processes/actions.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const listProcessProperties = (id) => async (dispatch, getState) => {
+    if (getState().propertyValues.isFetching)
+        return;
+
+    const { itemsByID: processesByID } = getState().processes    
+    const { itemsByID: propertyValuesByID } = getState().propertyValues
+
+    if (!(id in processesByID)) {
+        return await dispatch(listProcesses({ id__in: id }))
```
ckostiw, line 11: I don't think you want to `return` here. You just want to wait until the processes have been loaded and then continue to the next step.

frontend/src/modules/processes/actions.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    const process = processesByID[id];
+
+    if (!process?.children_processes?.every(process => process in processesByID)) {
+        return await dispatch(listProcesses({ id__in: process.children_processes.join() }))
```
ckostiw, line 16: Don't return here.

frontend/src/modules/processes/actions.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    const propertiesAreLoaded = process?.children_properties?.every(property => property in propertyValuesByID)
+    const childrenPropertiesAreLoaded = process?.children_processes?.every(process => processesByID[process]?.children_properties?.every(property => property in propertyValuesByID))
+    
+    if (!(propertiesAreLoaded && childrenPropertiesAreLoaded)) {
+        const processIDSAsStr = [id].concat(process.children_processes).join()
+        return await dispatch(listPropertyValues({ object_id__in: processIDSAsStr, content_type__model: "process" }))
```
ckostiw, line 24: No need to return here.

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -46,6 +52,15 @@ const ProcessMeasurementsDetailContent = ({
       listPropertyValues({ object_id__in: processMeasurement.id, content_type__model: "processmeasurement" });
     }
 
+    const processIsLoaded = processMeasurement?.process in processesByID;
+    const process = processesByID[processMeasurement?.process] || {};
+    if (!processIsLoaded) {
```
ckostiw, line 57: This block of code which is retrieving data and updating the redux state needs to be wrapped in useEffect(). We are not supposed to update state during render.

ckostiw (changes requested):

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
```diff
@@ -46,6 +52,15 @@ const ProcessMeasurementsDetailContent = ({
...
+      listProcesses({id__in: processMeasurement?.process});
```
UlysseFG, line 58: i would use a "get"  here since there is a single one...

frontend/src/components/processes/ProcessDetailContent.js
nafiz1001: Huh. I was told it was unnecessary. That's good to know 👍🏼

frontend/src/components/processMeasurements/ProcessMeasurementsDetailContent.js
nafiz1001: is there overhead associated with fetching a single item with `list`?

ckostiw:
I wonder if this table should contain a link to the process measurement. If they need to know the volume used, or the process measurement comment then they have to go back to the Protocols table and click on the pm link there. Or maybe the volume used and comment should be included in this table so there is no need to navigate to the process measurement?

Screen Shot 2022-07-19 at 9 50 12 AM

nafiz1001:
> Also, why are Properties displayed in a separate tab? Wouldn't it be easier for the user if they were just displayed below the Tracking Details in the Overview tab?

@UlysseFG 👀 ?

ckostiw:
Something is a bit broken. I did a library QC step on 3 samples and then went to the Protocols page. I clicked on the protocol I just ran, and clicked the Properties tab. The table showed my three samples, but all of the properties columns were empty - only the sample links in the first column showed up. Later, I went back to the same process and the properties were displayed normally. There may be a race condition or something that triggers an error. 

The console log contained an error message, but it's unclear if it's related to the problem (it might be a completely different issue).
Screen Shot 2022-07-20 at 4 37 18 PM

frontend/src/utils/api.js
```diff
@@ -115,6 +115,7 @@ const api = {
   },
 
   processes: {
+    get: processId => get(`/processes/${processId}/`),
     list: (options, abort) => get("/processes", options, { abort }),
```
UlysseFG, line 119: missing a "/" at the end of the path to match the other signatures.

frontend/src/modules/processes/actions.js
```diff
@@ -0,0 +1,57 @@
...
+        { meta: { ...options} }
+    ));
+};
+
+export const listProperties = (id) => async (dispatch, getState) => {
+    if (getState().propertyValues.isFetching)
+        return;
+
+    const { itemsByID: processesByID } = getState().processes    
+    const { itemsByID: propertyValuesByID } = getState().propertyValues
```
UlysseFG, line 31: aren't you creating twice the same constant with 2 different values ?
I don't know this notation but to my understanding what you do is const itemsByID = ... and then const itemsByID = ...

nafiz1001: that actually would be much simpler 😂

ckostiw:
I think that the "Process Properties" and "Process By Sample" strings should be greyed if there is no data to display.

Screen Shot 2022-07-25 at 9 54 35 AM

frontend/src/components/processes/ProcessDetailContent.js
```diff
@@ -94,31 +79,14 @@ const ProcessDetailContent = ({
...
-                        />
-                      </>
-                  )
-                })
-              }
-              </TabPane>
-            </Tabs>
-        </PageContent>
+          <TabPane tab="Properties" key="2" style={{ marginTop: 8 }}>
+            <Title level={3} style={{ marginTop: '20px' }}>Process Properties</Title>
```
UlysseFG, line 83: Try using these 2 titles for each property tab 
Shared process properties => for process level properties
Sample process properties => for processMeasurement  properties

ckostiw (changes requested):

frontend/src/components/shared/ProcessAssociatedMeasurements.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * 
+ * @param {any} processMeasurements 
+ * @param {any} processMeasurementsByID 
+ * @param {any} propertyValuesByID 
+ * @returns An array of property values where the property type is unique.
+ */
+const getPropertyValuesFromMeasurements = (processMeasurements, processMeasurementsByID, propertyValuesByID) => {
```
ckostiw, line 25: I think this should be named `getPropertyTypesFromMeasurements`, not values.

ckostiw (changes requested):

ckostiw (approved):
Approved, but we need a second story to cleanup the process measurements in the redux state when they are no longer needed.

ckostiw (approved):

PR #306, opened 2022-06-15
Added both index sets IDT TruSeq V1-V2 to the template merged in one set

sebastianballesteros:


UlysseFG:
In the template, the the first line in the Index columns has the Strandedness dropdown and the first line in the strandedness column no longer has a drop down. Looks like you cut and pasted something here ... Other than that the core, is ok ...

sebastianballesteros:
Good catch

UlysseFG:
Mmmm what i see is that now index 2 first lines are wrong. The strandedness drop down are ok now.

UlysseFG:
I still have the same file which date 2 hours back ... Did you push the right version ? The 2 first line of index give me the strandedness drop down. Have you tried copy and paste special - format ?

UlysseFG (approved):

PR #318, opened 2022-06-21
Data Delivery - Create FMS Backend Models and viewset, frontend to accept Dataset release flag.

nafiz1001:
image

image


backend/fms_core/migrations/0041_auto_20220623_1801.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.1 on 2022-06-23 18:01
```
sebastianballesteros, line 1: You might want to include those 2 auto-migrations into one 0040 migration file for simplicity.

ckostiw (changes requested):

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,21 @@
...
+from .experiment_run import ExperimentRun
+from .container import Container
+
+from ._constants import STANDARD_NAME_FIELD_LENGTH
+
+@reversion.register()
+class Dataset(TrackedModel):
+    """ Class to store information about the datasets of data deliveries. """
+
+    # project = models.ForeignKey(Project, null=True, blank=True, on_delete=models.PROTECT, help_text="The project the dataset belongs to")
```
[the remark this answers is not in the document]
nafiz1001: I treat it more like a TODO since in the near future we plan to actually have foreign keys to projects.

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,21 @@
...
+    project_name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, help_text="Project name.")
+
+    # run = models.ForeignKey(ExperimentRun, null=True, blank=True, on_delete=models.PROTECT, help_text="The experiment run in which the dataset was generated from")
```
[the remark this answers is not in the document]
nafiz1001: Similar to before.

frontend/src/components/datasets/DatasetsListContent.js
```diff
@@ -0,0 +1,149 @@
...
+        {
+            title: "Lane",
+            dataIndex: "lane",
+            sorter: true,
+        },
+        {
+            title: "Files",
+            dataIndex: "files",
+            render: (files, _) => {
+                // TODO: make this a link to the files
```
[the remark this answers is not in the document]
nafiz1001: Nope. I don't think I'm expected to create another page for each dataset files.

frontend/src/components/datasets/DatasetsListContent.js
```diff
@@ -0,0 +1,149 @@
...
+            render: (_, dataset) => {
+                return <>{dataset?.files?.find((f) => f?.completion_date)?.completion_date ?? "Unknown"}</>
+            }
+        },
+        {
+            title: "Validation Date",
+            dataIndex: "validation_date",
+            render: (_, dataset) => {
+                return <>{findValidationDate(dataset) ?? "N/A"}</>
+            }
```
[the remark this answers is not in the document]
nafiz1001: It was @UlysseFG 's idea to to use validation date as a flag 😉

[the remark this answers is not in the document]
UlysseFG: validation_date is a not as precise a name as i would like. The idea is to have a timestamp of the release time as flag. There is no need for a validation date right now. Maybe calling this data_release_timestamp would be clearer ...

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,21 @@
...
+    run_name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, help_text="Run name.")
+
+    lane = models.CharField(max_length=10, blank=True, help_text="Coordinates of the lane in a container")
```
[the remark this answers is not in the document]
UlysseFG: Might be useful to add a validator to make sure the lane is a number.

frontend/src/components/datasets/DatasetsListContent.js
```diff
@@ -0,0 +1,149 @@
...
+            dataIndex: "completion_date",
+            render: (_, dataset) => {
+                return <>{dataset?.files?.find((f) => f?.completion_date)?.completion_date ?? "Unknown"}</>
+            }
+        },
+        {
+            title: "Validation Date",
+            dataIndex: "validation_date",
+            render: (_, dataset) => {
+                return <>{findValidationDate(dataset) ?? "N/A"}</>
```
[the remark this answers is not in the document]
UlysseFG: This is because these dates are not dates ;) they are stored as datetimes. They would need some formatting to look purty.

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,195 @@
...
+from fms_core.models.dataset_file import DatasetFile
+from fms_core.models.dataset import Dataset
+
+def convertToModelDate(value: Union[datetime, date, str, None]) -> Union[datetime, None]:
+    if isinstance(value, date):
+        return datetime.combine(value, datetime.min.time())
+    elif isinstance(value, datetime):
+        return value
+    elif isinstance(value, str):
+        return datetime.fromisoformat(value[:10])
```
UlysseFG, line 15: I think this would provide you with a date rather than a datetime.

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,195 @@
...
+            sample_name=sample_name,
+            completion_date=completion_date,
+            validation_date=validation_date,
+        )
+    except ValidationError as e:
+        errors.append(';'.join(e.messages))
+
+    return (dataset_file, errors, warnings)
+
+def create_from_run_processing(run_processing_metrics: Dict[str, Any], completion_date: str, validation_date: str) -> Tuple[List[Dataset], List[DatasetFile], List[str], List[str]]:
```
UlysseFG, line 75: I would this out of the services. Services are a layer we want to keep a bit insulated from the incoming structure of data in order to be able to use and reuse it from different sources. This would belong to somewhere similar to the template importers in a more generic fashion.

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,195 @@
...
+        raise ValueError(f"Invalid date: {value}")
+
+def create_dataset(project_name: str, run_name: str, lane: str, files: List[Dict[str, Any]] = []) -> Tuple[Union[Dataset, None], List[str], List[str]]:
+    dataset = None
+
+    errors = []
+    warnings = []
+
+    try:
+        dataset = Dataset.objects.create(
```
UlysseFG, line 28: Try to anticipate the issues you could get from missing parameters and catch them before hitting the try as much as possible to be able to provide more useful error messages. This is unlikely to be propagated to the user in this case but we'll want to know what's wrong if it crashes.

frontend/src/components/datasets/DatasetsListContent.js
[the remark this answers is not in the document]
nafiz1001: @UlysseFG, where would I look for the data_release_timestamp? Would it be the time when it was submitted?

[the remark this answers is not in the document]
UlysseFG: The data release is empty until data is released manually by a user. We'll need to add a form / button / toggle in the table to release files. Hold a bit on that i think we'll have to check with PM how they feel about the functionality.

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,195 @@
...
+        if not errors:
+            dataset_file, newerrors, newwarnings = create_dataset_file(dataset, file['file_path'], file['sample_name'])
+            errors.extend(newerrors)
+            warnings.extend(newwarnings)
+
+            if dataset_file:
+                dataset_files.append(dataset_file)
+
+    if errors:
+        Dataset.objects.filter(id__in=dataset if dataset else []).delete()
```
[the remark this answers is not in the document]
nafiz1001: But what about datasets that have been created in the previous iteration? Shouldn't I delete those?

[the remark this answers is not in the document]
UlysseFG: For a second API call containing dataset data that fits an existing dataset pk, you can delete the dataset and corresponding dataset_files using the model delete (you using get()) and create new ones.

frontend/src/modules/datasets/actions.js
```diff
@@ -0,0 +1,137 @@
...
+});
+
+export const setReleaseFlags = (id, releaseFlag) => async (dispatch, getState) => {
+    const dataset = getState().datasets.itemsByID[id]
+    if (dataset?.isFetching)
+        return;
+
+    await dispatch(api.datasets.setReleaseFlags(id, releaseFlag))
+    return await dispatch(listFiles({ id__in: dataset?.files?.join(",") ?? "" }))
+};
```
nafiz1001, line 108: > btw, I think that code would be easier to read if you did a null check on dataset,
```js
export const setReleaseFlags = (id, releaseFlag) => async (dispatch, getState) => {
  const dataset = getState().datasets.itemsByID[id]
  if(dataset) {
    if (dataset.isFetching)
      return;

    await dispatch(api.datasets.setReleaseFlags(id, releaseFlag))
    return await dispatch(listFiles({ id__in: dataset.files?.join(",") ?? "" }))
  }
};
```

backend/fms_core/models/dataset_file.py
```diff
@@ -0,0 +1,22 @@
...
+from .container import Container
+
+from ._constants import STANDARD_FILE_PATH_LENGTH, STANDARD_NAME_FIELD_LENGTH, ReleaseFlag
+
+@reversion.register()
+class DatasetFile(TrackedModel):
+    """ Class to store information about the files associated with their dataset for data deliveries. """
+
+    dataset = models.ForeignKey(Dataset, on_delete=models.PROTECT, help_text="The dataset of the file", related_name="files")
+    file_path = models.CharField(max_length=STANDARD_FILE_PATH_LENGTH, help_text="File path to the dataset")
```
ckostiw, line 16: "File path to the dataset"? Isn't it the path to the file?

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,65 @@
...
+                        ) -> Tuple[Union[DatasetFile, None], List[str], List[str]]:
+    dataset_file = None
+    errors = []
+    warnings = []
+
+    try:
+        dataset_file = DatasetFile.objects.create(
+            dataset=dataset,
+            file_path=file_path,
+            sample_name=sample_name,
```
ckostiw, line 58: Should the service check if a sample with the given sample_name exists in the db, or is the name just informative?

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,35 @@
...
+
```
UlysseFG, line 23: Add a constraint that ensure project_name + run_name + lane is unique. It will prevent duplicate submissions.

backend/fms_core/services/dataset.py
nafiz1001: For now it is informative.

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,36 @@
...
+    def save(self, *args, **kwargs):
+        errors = {}
+
+        try:
+            int(self.lane)
+        except ValueError as e:
+            errors["ValueError"] = "The lane must be a positive integer"
```
UlysseFG, line 28: you do not test for negative ... using -5 will pass.

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,36 @@
...
+    run_name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, help_text="Run name.")
+
+    lane = models.CharField(max_length=10, blank=True, help_text="Coordinates of the lane in a container")
+
+    def save(self, *args, **kwargs):
+        errors = {}
+
+        try:
+            int(self.lane)
+        except ValueError as e:
```
UlysseFG, line 27: These validations should be in the clean() instead of the save()

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,36 @@
...
+            errors["ValueError"] = "The lane must be a positive integer"
+        
+        if Dataset.objects.filter(project_name=self.project_name, run_name=self.run_name, lane=self.lane).exists():
```
UlysseFG, line 30: This is a clean way to prevent adding existing duplicates using the app, but it does not prevent the injection of erroneous row in the DB using SQL. It would still be safer to include a constraint to make sure we do not make an error during a DB curation. It also does not prevent the creation of Dataset that have different case for project or run, but are otherwise identical.

frontend/src/modules/shared/actions.js
```diff
@@ -45,6 +47,8 @@ export const fetchInitialData = () => async (dispatch, getState) => {
         Libraries.listTable,
         Projects.listTable,
         ProcessMeasurements.listTable,
+        Datasets.listTable,
+        DatasetFiles.listTable,
```
UlysseFG, line 51: Not sure this is useful. I don't think we'll ever list all files without first going through the Datasets.

frontend/src/components/datasets/DatasetsPage.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { Redirect, Route, Switch } from "react-router-dom"
+import PageContainer from "../PageContainer"
+import DatasetDetailContent from "./DatasetDetailContent";
+import DatasetsListContent from "./DatasetsListContent";
+
+const DatasetsPage = () => <PageContainer>
+    <Switch>
```
UlysseFG, line 8: Might be worth it to merge your other PR and update the content here to be compatible with the new react requirements.

frontend/src/components/datasets/DatasetDetailContent.js
```diff
@@ -0,0 +1,136 @@
...
+    }, [datasetsById])
+
+    return <>
+    <AppPageHeader
+        title={`Dataset ${dataset?.run_name}`}
+        onBack={() => history.push("/datasets/list")}
+    />
+
+    <PageContent>
+        <Descriptions bordered={true} size="small">
```
UlysseFG, line 112: This is not pretty. I would try to fit all 4 items in a single row with the order of information being ID, Project, Run, Lane. If you can't fit all of them in a single row. Let the ID alone on it's row and put the other 3 on the second row.

frontend/src/modules/shared/actions.js
nafiz1001: If I don't fetch them initially, then it will be difficult to load the datasets into the table. PaginatedTable always assumes that the models have been fetched already.

backend/fms_core/models/dataset.py
nafiz1001: Do we allow lane number 0? Just in case they are programmers?

backend/fms_core/models/dataset.py
nafiz1001: > It would still be safer to include a constraint to make sure we do not make an error during a DB curation.
Which constraint am I missing?

backend/fms_core/services/dataset.py
```diff
@@ -0,0 +1,60 @@
...
+                        file_path: str,
+                        sample_name: str,
+                        release_flag: int = ReleaseFlag.BLOCK
+                        ) -> Tuple[Union[DatasetFile, None], List[str], List[str]]:
+    dataset_file = None
+    errors = []
+    warnings = []
+
+    if release_flag not in [value for value, _ in ReleaseFlag.choices]:
+        errors.append("The release flag can only be 1 (Blocked) or 2 (Released)")
```
UlysseFG, line 43: Use the releaseFlag.choices list in the error message in case the choices change.

backend/fms_core/viewsets/dataset.py
```diff
@@ -0,0 +1,122 @@
...
+        **_dataset_filterset_fields,
+    }
+
+    filter_class = DatasetFilter
+
+    @action(detail=False, methods=["post"])
+    def add_run_processing(self, request, *args, **kwargs):
+        data = request.data
+
+        def func(rpm):
```
UlysseFG, line 44: Is that function (func) really necessary ?

nafiz1001: it's mainly so i can return easily
maybe an exception would have been better...

backend/fms_core/migrations/0041_auto_20220816_1946.py
```diff
@@ -0,0 +1,58 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Dataset',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_at', models.DateTimeField(auto_now_add=True, help_text='Date the instance was created.')),
+                ('updated_at', models.DateTimeField(auto_now=True, help_text='Date the instance was modified.')),
+                ('deleted', models.BooleanField(default=False, help_text='Whether this instance has been deleted.')),
+                ('project_name', models.CharField(help_text='Project name.', max_length=200)),
```
UlysseFG, line 24: We will not be getting project names... we would be better off calling this external_project_id (external to make sure we do not confuse with FMS project_id and id so we know this is not a name...).

backend/fms_core/migrations/0041_auto_20220816_1946.py
```diff
@@ -0,0 +1,58 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+import fms_core.schema_validators
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('fms_core', '0040_v3_10_0'),
```
UlysseFG, line 13: Rename your file to match the naming scheme you might have to update your dependency (0041 may or may not already exist...).

backend/fms_core/models/_constants.py
```diff
@@ -11,3 +12,11 @@
 INDEX_READ_FORWARD = "FORWARD"
 INDEX_READ_REVERSE = "REVERSE"
 INDEX_READ_DIRECTIONS_CHOICES = ((INDEX_READ_FORWARD, INDEX_READ_FORWARD), (INDEX_READ_REVERSE, INDEX_READ_REVERSE))
+RELEASE_FLAG = 1
```
UlysseFG, line 15: RELEASE_FLAG and BLOCK_FLAG are not used (by chance they seem to be conspicuously similar).

backend/fms_core/models/dataset.py
```diff
@@ -0,0 +1,47 @@
...
+
+    project_name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, help_text="Project name.")
+
+    run_name = models.CharField(max_length=STANDARD_NAME_FIELD_LENGTH, help_text="Run name.")
+
+    lane = models.PositiveIntegerField(help_text="Coordinates of the lane in a container")
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(fields=["project_name", "run_name", "lane"], name="unique_dataset")
```
UlysseFG, line 24: use a more descriptive name for the constraint so we know what it represent (it is typical to list the unique fields e.g. "unique_externalprojectid_runname_lane")

frontend/src/components/datasets/DatasetDetailContent.js
```diff
@@ -0,0 +1,227 @@
...
+        title={`Dataset ${loading(dataset?.project_name)} - ${loading(dataset?.run_name)} - ${loading(dataset?.lane)}`}
+    />
+
+    <PageContent>
+        <Descriptions bordered={true} size={"small"} column={4}>
+            <Descriptions.Item label={"ID"}>{loading(dataset?.id)}</Descriptions.Item>
+            <Descriptions.Item label={"Project"}>{loading(dataset?.project_name)}</Descriptions.Item>
+            <Descriptions.Item label={"Run Name"}>{loading(dataset?.run_name)}</Descriptions.Item>
+            <Descriptions.Item label={"Lane"}>{loading(dataset?.lane)}</Descriptions.Item>
+            <Descriptions.Item label={"Total Files"}>{loading(dataset?.files?.length)}</Descriptions.Item>
```
UlysseFG, line 218: span = {2} for total files and released

frontend/src/components/datasets/DatasetsListContent.js
```diff
@@ -0,0 +1,123 @@
...
+        {
+            title: "Files Released",
+            dataIndex: "release_flag_count",
+            sorter: true,
+            render: (release_flag_count, dataset) => {
+                return `${release_flag_count}/${dataset?.files?.length}`
+            }
+        },
+        {
+            title: "Last Release Time",
```
UlysseFG, line 53: I would rename this Release Date and format it to only display the date part.
Also, In order to give an indication that the whole dataset is released (state either Release or Block no remaining Available) i would only display this date if there is no more Available (or whatever the default name we give) file in the dataset. That would give an indication there is no more action to be done for this dataset (in case someone forget to release a file for some reason).

frontend/src/components/App.js
```diff
@@ -270,6 +277,11 @@ const App = ({userID, usersByID, logOut, fetchInitialData, fetchSummariesData, g
                 <About />
               </PrivateNavigate>
             }/>
+            <Route path="/datasets/*" element={
+              <PrivateNavigate>
+                <DatasetsPage/>
```
UlysseFG, line 282: I would move this menu item between the Indices and Users.

ckostiw (approved):

sebastianballesteros (approved):

PR #325, opened 2022-06-29
Upgrade some packages in the front-end

nafiz1001:
# Migrations

## `react-redux`

<https://github.com/reduxjs/react-redux/releases/tag/v8.0.0>

> In most cases, it's very likely that the only change you will need to make is bumping the package version to "react-redux": "^8.0".

- [x] bumping the package version to "react-redux": "^8.0"

## `react-router-dom`

[<https://github.com/remix-run/react-router/discussions/8753>](https://github.com/remix-run/react-router/discussions/8753#discussion-3978470)

- [x] Upgrade your app to React 16.8+
- [x] Read from v6 useParams() instead of v5 props.match
- [x] Read from v6 useLocation() instead of v5 props.location
- [x] Use navigate instead of history
- [x] Update links to use relative to values
- ~[ ] Update nav links (optional)~
- [x] Convert `<Switch>` to `<Routes>`
- [x] Add splat paths to any <Route> with a descendant <Routes>
- [x] Convert route paths from absolute to relative paths

## `webpack-cli`

- [x] update package.json
  - `npm start` is missing `--mode=development`

## `webpack-dev-server`

<https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md>

- [x] webpack-dev-server v3 and webpack-dev-server v4 automatically apply HotModuleReplacementPlugin plugin when you set hot: true, so please check you don't have HotModuleReplacementPlugin in your plugins if you have hot: true/hot: "only"
- ~[ ] webpack-dev-server v3 and webpack-dev-server v4 automatically inject webpack/hot/dev-server in your entry option when you set hot: true (except when you use injectHot for webpack-dev-server v3), please check you don't have webpack/hot/dev-server in your entry option~
- ~[ ] webpack-dev-server v3 and webpack-dev-server v4 automatically inject webpack-dev-server/client/index.js in your entry option (except when you use injectClient for webpack-dev-server v3), please check you don't have webpack-dev-server/client/index.js in your entry option~
- [x] The hotOnly option was removed, if you need hot only mode, use { hot: 'only' } value.
- [x] contentBase/contentBasePublicPath/serveIndex/watchContentBase/watchOptions/staticOptions options were moved to static option 
- [x] The rest are clearly not relevant to our code

## React/ReactDOM

- [x] Make necessary upgrades to the dependencies of react-d3-graph (fork). Or maybe don't use react-d3-graph 😨.

### 16 -> 17

https://reactjs.org/blog/2020/10/20/react-v17.html

- ~[ ] Changes to Event Delegation (hopefully this is not a problem, chris might be more qualified to judge this)~
- ~[ ] The onScroll event no longer bubbles to prevent [common confusion](https://github.com/facebook/react/issues/15723).~
- ~[ ] React onFocus and onBlur events have switched to using the native focusin and focusout events under the hood, which more closely match React’s existing behavior and sometimes provide extra information~
- ~[ ] Capture phase events (e.g. onClickCapture) now use real browser capture phase listeners~
- ~[ ] timing of the useEffect cleanup function more consistent (hopefully useEffect is being used properly 😒)~

### 17 -> 18

https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

- [x] `ReactDOM.render` is no longer supported in React 18. Use `createRoot` instead.
- [x] `unmountComponentAtNode` to `root.unmount`

## Cypress

- [x] Update Cypress (according to https://github.com/c3g/freezeman/pull/273#issue-1214984921)


frontend/package.json
```diff
@@ -21,56 +21,56 @@
...
     "dagre": "^0.8.5",
-    "date-fns": "^2.15.0",
-    "debounce": "^1.2.0",
-    "dompurify": "^2.2.7",
-    "dotenv": "^10.0.0",
-    "fuse.js": "^6.4.0",
+    "date-fns": "^2.28.0",
+    "debounce": "^1.2.1",
+    "dompurify": "^2.3.8",
+    "dotenv": "^16.0.1",
```
nafiz1001, line 35: I'm not sure if I should worry about this upgrade, do we even use dotenv?

nafiz1001:
We can take this a step further by forking react-d3-graph, and upgrading the relevant packages which I've already done 😉. That way we can upgrade React to version 18 😉.

frontend/src/components/PrivateRoute.js
```diff
@@ -1,16 +0,0 @@
-import React from "react";
-import {connect} from "react-redux";
-import {Redirect, Route} from "react-router-dom";
-
-const PrivateRoute = ({isAuthenticated, children, ...rest}) => (
```
UlysseFG, line 5: I'm not sure you had to change all these PrivateRoutes into PrivateNavigate.

nafiz1001: Maybe not, but I couldn't fix examples of how one could use something like `PrivateRoutes` in the new `Route` component. It also seems to be the more correct way. For example: <https://dev.to/iamandrewluca/private-route-in-react-router-v6-lg5>

UlysseFG: Not as much the renaming than the practicality of having a wrapper to encompass the route instead of having just a wrapper to the auth. Looks like you did your research and Its ok to change it, but it seem to be a step backward in code readability and simplicity. I don't see how this element prop fit in the react way of doing things.

nafiz1001: I believe it's because React developers prefers if the children of `Route` are `Route`s. Otherwise, they will have to parse the `children` prop to figure out if it's a `<Route>` or `<div>, etc.

sebastianballesteros:
There's a problem when accessing the actions routes, i.e. `http://localhost:9000/samples/actions/0/` maybe related to static files?

frontend/src/components/ActionContent.js
```diff
@@ -38,8 +38,8 @@ const submitRequests = {
 }
 
 const ActionContent = ({token, templateType, templateActions}) => {
-  const history = useHistory();
-  const match = useRouteMatch();
+  const history = useNavigate();
+  const match = useMatch();
```
sebastianballesteros, line 42: `const { action : actionId } = useParams();

  const actionIndex = parseInt(actionId, 10) || 0;`

sebastianballesteros, line 42: `const goBack = () => {
    history(-1)
  }`

ckostiw:
As reported in Slack, there is a routing issue wit the user details page. Whenever we route to the page the page immediately routes back to wherever we routed from. For example, in the Users table, clicking on a user will route to the user details but then immediately route back to the Users table.

ckostiw:
I pushed a fix for the "expected static flag was missing" react error, based on this issue: https://github.com/facebook/react/issues/24391
The problem is that we were checking for an undefined `container` value and bailing before `useEffect()` was called. The fix is to first call `useEffect()` and then bail if necessary. Otherwise, when React renders the component, sometimes useEffect is called and sometimes it isn't, and that is not allowed in React.

ckostiw:
Now we have this error in Container Details
Screen Shot 2022-08-03 at 4 44 55 PM

ckostiw:
Also in Container Details

Screen Shot 2022-08-03 at 4 50 16 PM

ckostiw (approved):

PR #327, opened 2022-06-30
Fixed sammple export individual problem

sebastianballesteros:


ckostiw (approved):
A sample without an individual can now be exported without error.

PR #329, opened 2022-07-05
Added normalization protocol

sebastianballesteros:


backend/fms_core/migrations/0040_v3_10_0.py
```diff
@@ -0,0 +1,58 @@
...
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+        admin_user_id = admin_user.id
+
+        reversion.set_comment("Add properties' type for Normalization")
+        reversion.set_user(admin_user)
+
+        PROPERTY_TYPES_BY_PROTOCOL = {
+            "Normalization": [
+                ("Volume", "str"),
```
UlysseFG, line 23: maybe call these "target volume" and "target concentration" or "final ..." or ... Just to make sure the content is clear enough for the user (and for us).

backend/fms_core/templates.py
```diff
@@ -168,6 +169,23 @@
...
+        'headers': ['Source Container Barcode', 'Source Container Coord', 'Destination Container Barcode', 'Destination Container Coord',
+                    'Destination Container Name', 'Destination Container Kind', 'Destination Parent Container Barcode',
+                    'Destination Parent Container Coord', 'Source Depleted', 'Volume Used (uL)', 'Volume (uL)',
+                    'Conc. (ng/uL)', 'Conc. (nM)', 'Normalization Date (YYYY-MM-DD)', 'Comment'],
+      },
+  ],
+  "prefill info": [
+    ("Normalization", "Source Container Barcode", "container__barcode"),
+    ("Normalization", "Source Container Coord", "coordinates"),
+    ],
```
UlysseFG, line 186: too indented

backend/fms_core/templates.py
```diff
@@ -168,6 +169,23 @@
...
+      {
+        'name': 'Normalization',
+        'headers': ['Source Container Barcode', 'Source Container Coord', 'Destination Container Barcode', 'Destination Container Coord',
+                    'Destination Container Name', 'Destination Container Kind', 'Destination Parent Container Barcode',
+                    'Destination Parent Container Coord', 'Source Depleted', 'Volume Used (uL)', 'Volume (uL)',
+                    'Conc. (ng/uL)', 'Conc. (nM)', 'Normalization Date (YYYY-MM-DD)', 'Comment'],
+      },
+  ],
+  "prefill info": [
+    ("Normalization", "Source Container Barcode", "container__barcode"),
```
UlysseFG, line 184: Make me think we may be better to add an optional "source sample name" column to the template for prefilling.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+        destination_container_dict = destination_sample['container']
+
+        parent_barcode = destination_container_dict['parent_barcode']
+        if parent_barcode:
+            container_parent, self.errors['parent_container'], self.warnings['parent_container'] = get_container(
+                barcode=parent_barcode)
+        else:
+            container_parent = None
+
+        if source_sample and ((parent_barcode and container_parent) or not parent_barcode):
```
UlysseFG, line 28: This is an overly complicated test it seem to be sufficient to test source_sample and (container_parent or not parent_barcode) in principle to get the same result and a simple logic.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+            destination_container, _, self.errors['container'], self.warnings['container'] = get_or_create_container(
+                barcode=destination_container_dict['barcode'],
+                kind=destination_container_dict['kind'],
+                name=destination_container_dict['name'],
+                coordinates=destination_container_dict['coordinates'],
+                container_parent=container_parent)
+
+            source_depleted = check_truth_like(source_sample['depleted']) if source_sample['depleted'] else None
```
UlysseFG, line 36: I'm unsure if we should keep the source_depleted conversion here instead of in the importer. It does not really change the performance or readability but it is more of a logical organization. I think of the importer as the place to do simple preprocessing and normalization in order to have more or less final values ready to be passed along. It's not that important just something to think about.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+
+            if destination_sample['concentration_nm'] is not None and destination_sample['concentration_uL'] is not None:
+                self.errors['concentration'] = 'Concentration must be specified in either nM or ng/uL, not both.'
+
+            concentration = destination_sample['concentration_uL']
+            if concentration is None:
+                concentration = destination_sample['concentration_nm']
+                if source_sample_obj.is_library:
+                    library = source_sample_obj.derived_sample_not_pool.library
+                    if library.library_size:
```
UlysseFG, line 61: It is very deep into the processing to be checking this ... we would save a lot of processing if we would to it at the top.

backend/fms_core/template_importer/importers/normalization.py
```diff
@@ -0,0 +1,89 @@
...
+            source_sample = {
+                'container': {'barcode': str_cast_and_normalize(row_data['Source Container Barcode'])},
+                'coordinates': str_cast_and_normalize(row_data['Source Container Coord']),
+                'depleted': str_cast_and_normalize(row_data['Source Depleted']),
+            }
+
+            destination_sample = {
+                'coordinates': str_cast_and_normalize(row_data['Destination Container Coord']),
+                'volume': float_to_decimal_and_none(row_data['Volume (uL)']),
+                'concentration_uL': float_to_decimal_and_none(row_data['Conc. (ng/uL)']),
```
UlysseFG, line 58: _ngul instead of _uL maybe ?

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # Update concentration
+            if destination_sample['concentration_nm'] is None and destination_sample['concentration_uL'] is None:
+                self.errors['concentration'] = 'A concentration in either nM or ng/uL must be specified.'
+
+            if destination_sample['concentration_nm'] is not None and destination_sample['concentration_uL'] is not None:
+                self.errors['concentration'] = 'Concentration must be specified in either nM or ng/uL, not both.'
+
+            concentration = destination_sample['concentration_uL']
+            if concentration is None:
```
UlysseFG, line 57: Add comment to mention this is to test the case the input is in nM.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+                concentration = destination_sample['concentration_nm']
```
UlysseFG, line 58: Do not put anything in concentration as long as it's not converted.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,91 @@
...
+                    library = source_sample_obj.derived_sample_not_pool.library
+                    if library.library_size:
+                        concentration = convert_concentration_from_nm_to_ngbyul(concentration,
+                                                                                library.molecular_weight_approx,
+                                                                                library.library_size)
+                    else:
+                        self.errors['library'] = 'Library size has not been set for this library.'
+
+                    if concentration is None:
+                        self.errors['concentration'] = 'Concentration could not be converted from nM to ng/uL'
```
UlysseFG, line 69: I would use "failed to convert" since this looks like an unexpected failure.

backend/fms_core/viewsets/sample.py
```diff
@@ -89,6 +89,12 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePrefill
             "template": [SAMPLE_METADATA_TEMPLATE["identity"]],
             "importer": SampleMetadataImporter,
         },
+        {
+            "name": "Normalize Samples",
```
UlysseFG, line 93: Samples or Libraries

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,92 @@
...
+            if concentration is not None:
+                _, self.errors['concentration_update'], self.warnings['concentration_update'] = \
+                    update_sample(sample_to_update=resulting_sample, concentration=concentration)
+
+            # Set the process measurement properties
+            process_measurement_properties['Final Volume']['value'] = destination_sample['volume']
+            process_measurement_properties['Final Concentration']['value'] = concentration
+
+            # Create process measurement's properties
+            process_measurement_obj = ProcessMeasurement.objects.get(source_sample=source_sample_obj)
```
UlysseFG, line 82: get() returned more than one ProcessMeasurement -- it returned 3!
This filter is not enough specific ...

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,92 @@
...
+                sample_source=source_sample_obj,
+                container_destination=destination_container,
+                volume_used=process_measurement['volume_used'],
+                execution_date=process_measurement['execution_date'],
+                coordinates_destination=destination_sample['coordinates'],
+                volume_destination=destination_sample['volume'],
+                source_depleted=source_sample['depleted'],
+                comment=process_measurement['comment'])
+
+            concentration = None
```
UlysseFG, line 54: Move the whole concentration calculation before the container creation and  before moving forward make a validation that the volume used, source_sample concentration, final volume and final concentration add up. It will likely not be an exact match but you can round both values up to 1 decimal (when comparing theoretical to submitted concentration) and make sure it match at this level...

backend/fms_core/services/sample.py
```diff
@@ -525,6 +525,36 @@ def remove_sample_metadata(sample, metadata):
     return deleted, errors, warnings
 
 
+def validate_normalization(initial_volume, initial_concentration, final_volume, desired_concentration, tolerance=1):
```
UlysseFG, line 528: I would set the default tolerance to 0.01

backend/fms_core/services/sample.py
```diff
@@ -525,6 +525,36 @@ def remove_sample_metadata(sample, metadata):
...
+        errors.append(f"Final volume is required for validation.")
+    if desired_concentration is None:
+        errors.append(f"Final concentration is required for validation.")
+
+    if not errors:
+        # Calculate the current amount to be able to calculate final concentration
+        solute_amount = initial_concentration * initial_volume
+        computed_concentration = solute_amount / final_volume
+
+        delta_concentration = round(computed_concentration, 2) - round(desired_concentration, 2)
```
UlysseFG, line 548: We should not round the tolerance will deal with the any jitter.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -0,0 +1,100 @@
...
+                        self.errors['concentration'] = 'Failed to convert concentration from nM to ng/uL'
+                else:
+                    self.errors['concentration'] = 'Concentration specified in nM should be only for libraries.'
+
+            is_concentration_valid, self.errors['concentration_validation'], self.warnings['concentration_validation'] \
+                = validate_normalization(initial_volume=process_measurement['volume_used'],
+                                         initial_concentration=source_sample_obj.concentration,
+                                         final_volume=destination_sample['volume'],
+                                         desired_concentration=concentration,
+                                         tolerance=1)
```
UlysseFG, line 60: You can leave it to default here we'll tweak the default to be a good compromise for normalization and if we need to reuse we'll use the param to get it right if the default is not good enough.

UlysseFG:
Just a remainder when you have a minute we need a "Robot Source Container" and "Robot Source Coord" after "Source Container Coord" (same section) and "Robot Destination Container" and "Robot Destination Coord" after "Destinatiion Container Coord". Also it would be nice to have some indicators for required columns and a Sample name optional column at the beginning of the template for prefilling.

sebastianballesteros:
Wouldn't it be better to put the two columns "Robot Destination Container" and "Robot Destination Coord" after "Destination Container Kind" instead of "Destinatiion Container Coord"

UlysseFG:
> Wouldn't it be better to put the two columns "Robot Destination Container" and "Robot Destination Coord" after "Destination Container Kind" instead of "Destinatiion Container Coord"
Possibly not too sure... Those columns are meant to help someone map lab container to robot "container" i imagine if i would check this i would like the columns to be close together ... In the end its not that important unless we get feedback from the users ...

backend/fms_core/services/sample.py
```diff
@@ -525,6 +525,36 @@ def remove_sample_metadata(sample, metadata):
...
+    if not errors:
+        # Calculate the current amount to be able to calculate final concentration
+        solute_amount = initial_concentration * initial_volume
+        computed_concentration = solute_amount / final_volume
+
+        delta_concentration = computed_concentration - desired_concentration
+        if abs(delta_concentration) <= tolerance:
+            is_valid = True
+        else:
+            errors.append(f'Desired concentration [{desired_concentration}] is not valid.')
```
UlysseFG, line 552: i would add " given the dilution ratio (source volume used / final volume)." to make it clear that validation is linked to the volumes submitted in the template too.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
UlysseFG: After testing 1 is much too permissive... you should use the default of 0.01, we could possibly go lower ... but it may get annoying. Our calculation at the third decimal should always pass at the 0,01 level and they will prevent gross errors at least.

backend/fms_core/viewsets/sample.py
```diff
@@ -89,6 +89,12 @@ class SampleViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePrefill
             "template": [SAMPLE_METADATA_TEMPLATE["identity"]],
             "importer": SampleMetadataImporter,
         },
+        {
```
UlysseFG, line 92: I would add the action to the library viewset too in order to have the option in the library or sample as well as to the library prefilling. Change the name and description to be specific on the action for the library (the current one is good since you can do both at the sample level).

UlysseFG: The dropdown for the Robot columns should be removed in the template.

UlysseFG (approved):

UlysseFG:
The Robot column dropdown are still there if you can remove them it would be great...

PR #331, opened 2022-07-07
Added canis lupus taxon

sebastianballesteros:


ckostiw (changes requested):
Shouldn't canis lupus be added to the sample submission template as well?

ckostiw:
@sebastianballesteros If users now enter a taxon id instead of picking the taxon name from a dropdown then I can just approve this PR.

PR #332, opened 2022-07-07
Fix for get_taxon() plus tests

ckostiw:
I found a small bug in the get_taxon() function in the Individual services, where the `name` parameter was not used properly. It's not critical since we only get taxons by ID in freezeman, so the name argument is never used, but I fixed it to avoid problems in the future. 
I also added some tests, to be able to test my fix.

### DIFF
--- a/backend/fms_core/tests/test_services/individual_test.py
+++ b/backend/fms_core/tests/test_services/individual_test.py
@@ -6,15 +6,16 @@ class IndividualServicesTestCase(TestCase):
 
     def test_get_taxon(self):
         HOMO_SAPIENS_NAME = 'Homo sapiens'
+        HOMO_SAPIENS_NON_NORMALIZED_NAME = 'HOMO SAPIENS'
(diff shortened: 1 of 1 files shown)


sebastianballesteros (approved):

backend/fms_core/tests/test_services/individual_test.py
```diff
@@ -0,0 +1,38 @@
...
+from fms_core.services.individual import get_taxon
+
+class IndividualServicesTestCase(TestCase):
+
+    def test_get_taxon(self):
+        HOMO_SAPIENS_NAME = 'Homo sapiens'
+        HOMO_SAPIENS_NCBI_ID = 9606
+
+        # Check with normalized name
+        taxon, _, _ = get_taxon(name='Homo sapiens')
```
UlysseFG, line 12: You could use the constant as input too and make a different one for the non-normalized test.

UlysseFG (approved):

PR #333, opened 2022-07-11
optimize `list_export`

nafiz1001:


sebastianballesteros:
I think you don't need the first values() in both sample_values and derived_sample_values, with the annotation() and the second values() might suffice.

backend/fms_core/serializers.py
```diff
@@ -352,42 +352,52 @@ def get_quantity_flag(self, obj):
...
+    def to_list(obj: Union[Any, List[Any]]) -> List[Any]:
+                try:
+                    return list(obj)
+                except TypeError:
+                    if obj is None:
+                        return []
+                    else:
+                        return [obj]
+
+    derived_sample = derived_samples[sample["first_derived_sample"]]
```
UlysseFG, line 365: This is an unsafe function in the way that there is no sign this function need pre-existant annotations to be done. 
I would at the very least add a comment in the function (of course !) that warns against calling this without the proper annotations. This is an extention to the code from the viewset. 
I think a cleaner way would be to move all relevant annotation and processing into this function from the viewset... Other than the initial queryset filtering and returning data the rest is mostly directly related to serializing data.

backend/fms_core/viewsets/sample.py
```diff
@@ -273,8 +275,117 @@ def update(self, request, *args, **kwargs):
...
+                'container_name': sample["container__name"],
+                'container_barcode': sample["container__barcode"],
+                'coordinates': sample["coordinates"],
+                'location_barcode': sample["container__location__barcode"] or "",
+                'location_coord': sample["container__coordinates"] or "",
+                'current_volume': sample["volume"],
+                'concentration': sample["concentration"],
+                'creation_date': sample["creation_date"],
+                'collection_site': derived_sample["biosample__collection_site"],
+                'experimental_group': derived_sample["experimental_group"],
```
UlysseFG, line 369: Experimental_group return nothing. Maybe doing a json dumps might help.

backend/fms_core/serializers.py
nafiz1001: I'm not sure if this is still relevant to the current state of the code.

UlysseFG (approved):

backend/fms_core/serializers.py
UlysseFG: If we are not using it, you can remove it.

backend/fms_core/utils.py
```diff
@@ -112,4 +112,14 @@ def convert_concentration_from_nm_to_ngbyul(concentration_nm, molecular_weight,
         return None
     concentration = (Decimal(concentration_nm) * Decimal(molecule_count) * Decimal(molecular_weight)) / Decimal(1000000)
 
-    return concentration
\ No newline at end of file
+    return concentration
+
+def ensure_iterable(obj: Union[Any, None, Iterable[Any]]) -> Iterable[Any]:
```
UlysseFG, line 117: Is it intended that you ensure_iterable returns a generator ?

backend/fms_core/utils.py
```diff
@@ -114,7 +114,24 @@ def convert_concentration_from_nm_to_ngbyul(concentration_nm, molecular_weight,
...
     return concentration
 
-def ensure_iterable(obj: Union[Any, None, Iterable[Any]]) -> Iterable[Any]:
+T = TypeVar("T")
+def ensure_iterable(obj: Union[T, None, Iterable[T]]) -> Iterable[T]:
+    """Ensures that ManyToMany fields such as the `obj` passed are iterable.
+    None is turned into an empty iterable,
+    non-None objects are turned into an iterable with a single element,
+    and iterable objects remain the same.
+    It's meant to handle to the fact that a ManyToMany field is not a list if it has less than two elements.
```
nafiz1001, line 123: Uuugh... typo

backend/fms_core/utils.py
```diff
@@ -114,7 +114,24 @@ def convert_concentration_from_nm_to_ngbyul(concentration_nm, molecular_weight,
 
     return concentration
 
-def ensure_iterable(obj: Union[Any, None, Iterable[Any]]) -> Iterable[Any]:
+T = TypeVar("T")
```
nafiz1001, line 117: @UlysseFG Is this ok?

UlysseFG: eheh you can remove that TypeVar thing ... This does not help clarify things. Your comment give the gist of it ... I would personally call it "make_generator" or something of the kind but it's really a personal preference, You can merge when you are happy with what you have...

UlysseFG (approved):

PR #334, opened 2022-07-18
Removed reference to download action-reducer which was removed.

UlysseFG:


nafiz1001 (approved):
The less code the better 😄

PR #335, opened 2022-07-18
Prevent templates with dates in the future (process_measurement and sample)

UlysseFG:


backend/fms_core/utils.py
```diff
@@ -88,6 +88,13 @@ def get_normalized_str(d: dict, key: str, default: str = "") -> str:
     """
     return str_cast_and_normalize(d.get(key) or default)
 
+
+def is_date_after_today(date):
+    if not isinstance(date, datetime.date):
+      return None
```
ckostiw, line 94: I wonder if we should raise an exception here - it might help catch bugs.

backend/fms_core/models/sample.py
```diff
@@ -154,6 +155,10 @@ def add_error(field: str, error: str):
         if self.volume is not None and self.volume < Decimal("0"):
             add_error("volume", "Current volume must be positive.")
 
+        # Make sure the creation date is not in the future 
+        if datetime.combine(self.creation_date, datetime.min.time()) > timezone.make_naive(timezone.now()):
```
ckostiw, line 159: Shouldn't the util function be used here?

backend/fms_core/models/process_measurement.py
```diff
@@ -42,11 +44,15 @@ def add_error(field: str, error: str):
...
-                add_error("volume_used", f'volume_used by processes for protocol {self.protocol_name} must be specified')
+                add_error("volume_used", f'volume_used by processes for protocol {self.protocol_name} must be specified.')
 
         else:
             if self.protocol_name not in PROTOCOLS_WITH_NEGATIVE_VOLUME_USED_ALLOWED and self.volume_used <= Decimal("0"):
-                add_error("volume_used", f'volume_used {self.volume_used} must be positive for protocol {self.protocol_name}')
+                add_error("volume_used", f'volume_used {self.volume_used} must be positive for protocol {self.protocol_name}.')
+
+        # Make sure the execution date is not in the future 
+        if datetime.combine(self.execution_date, datetime.min.time()) > timezone.make_naive(timezone.now()):
```
ckostiw, line 54: Shouldn't this be using the `is_date_after_today` util function?

ckostiw (changes requested):

backend/fms_core/utils.py
```diff
@@ -88,6 +88,13 @@ def get_normalized_str(d: dict, key: str, default: str = "") -> str:
     """
     return str_cast_and_normalize(d.get(key) or default)
 
+
+def is_date_after_today(date):
+    if not isinstance(date, datetime.date):
```
ckostiw, line 93: Should this also support `datetime` objects, and use `date.date()` to convert to a simple date?

ckostiw (changes requested):

backend/fms_core/utils.py
UlysseFG: it's an utility i don't want to propagate errors each time i use this. If it breaks we'll fix it.

ckostiw (changes requested):

backend/fms_core/utils.py
```diff
@@ -89,10 +87,10 @@ def get_normalized_str(d: dict, key: str, default: str = "") -> str:
     return str_cast_and_normalize(d.get(key) or default)
 
 
-def is_date_after_today(date):
+def is_date_or_time_after_today(date: datetime) -> bool or None:
```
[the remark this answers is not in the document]
UlysseFG: Union is pretty ;) these are annotations we can put whatever we want ;)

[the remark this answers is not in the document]
UlysseFG: datetime.datetime is a datetime, datetime is either a date or a datetime...

ckostiw (approved):

PR #339, opened 2022-07-21
Add sample name prefill to templates

sebastianballesteros:


backend/fms_core/templates.py
```diff
@@ -108,7 +110,7 @@
       },
       {
           'name': 'Library',
-          'headers': ['Library Batch ID', 'Library Source Container Barcode', 'Library Source Container Coordinates',
+          'headers': ['Library Batch ID', 'Library Source Name', 'Library Source Container Barcode', 'Library Source Container Coordinates',
```
UlysseFG, line 113: "Library Source Name" ... i think we can do better "Source Library Name" is already better. It could also just mention "Source Sample Name" since it is a sample.

sebastianballesteros: I kept it like this for consistency.  All of the header are "Library Source ... "

UlysseFG (approved):

PR #340, opened 2022-07-21
Fix container type casing error for "Tube" type

ckostiw:
Fixes redmine issue: [Feature #1185](https://206.12.92.46/issues/1185)

Container kind strings are stored in lower case in the db. In the sample submission form the drop down uses "Tube" as the type. Container kind strings need to be converted to lowercase by the importer.

### DIFF
--- a/backend/fms_core/utils.py
+++ b/backend/fms_core/utils.py
@@ -81,7 +81,7 @@ def str_cast_and_normalize(s) -> str:
 
 def str_cast_and_normalize_lower(s) -> Union[str, None]:
     """
-    Cases a value to a string, normalizes it and then converts to lower case.
+    Casts a value to a string, normalizes it and then converts to lower case.
     """
     result = str_cast_and_normalize(s)
     return result.lower() if result is not None else result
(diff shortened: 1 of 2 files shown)


ckostiw:
Before this fix, `get_or_create_container` would compare "Tube" with "tube" and complain that about the container kind not matching here: https://github.com/c3g/freezeman/blob/03f5cdc910e94a85ca5d6f9bd1b7e4e02c875196/backend/fms_core/services/container.py#L48

backend/fms_core/utils.py
```diff
@@ -79,6 +79,14 @@ def str_cast_and_normalize(s) -> str:
     return str_normalize(str(s) if s is not None else s)
 
 
+def str_cast_and_normalize_lower(s) -> Union[str, None]:
+    """
+    Casts a value to a string, normalizes it and then converts to lower case.
+    """
+    result = str_cast_and_normalize(s)
+    return result.lower() if result is not None else result
```
UlysseFG, line 87: would be clearer with "else None" as it clearly state the return value

ckostiw: I hesitated between returning None or result but I agree - returning None would be more readable.

PR #342, opened 2022-07-21
Added warning when using an existing name

sebastianballesteros:


backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -129,6 +129,11 @@ def process_row_inner(self, sample, library, container, project, parent_containe
                                                                                            index=index_obj,
                                                                                            platform=platform_obj,
                                                                                            strandedness=library['strandedness'])
+
+        # Check if there's a sample with the same name
+        if Sample.objects.filter(name=sample['name']).exists():
+            self.warnings['name'] = f'Sample with the same name [{sample["name"]}] already exists.'
```
UlysseFG, line 135: I would explain what will be the resulting action taken if they keep going ("A new sample with the same name will be created.)

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -129,6 +129,11 @@ def process_row_inner(self, sample, library, container, project, parent_containe
                                                                                            index=index_obj,
                                                                                            platform=platform_obj,
                                                                                            strandedness=library['strandedness'])
+
+        # Check if there's a sample with the same name
+        if Sample.objects.filter(name=sample['name']).exists():
```
UlysseFG, line 134: Another warning would be nice testing with iexact to see if there is another sample with the same name with different case. We could suggest checking the name.

UlysseFG:
All of these warning are very nice and useful but i fear users will see warning too often and will little attention. With these added checks i would also add a warning counter in the warning component of the frontend to let user know there is 1, 2, 3 errors. That way they can flag a column where there is more warning than expected (there is always the warning for plates for example if there is other warning it should be visible without hovering.)

UlysseFG (approved):

PR #346, opened 2022-07-26
WIP: Normalization Planning

sebastianballesteros:


backend/fms_core/template_importer/row_handlers/normalization_planning/normalization_planning.py
```diff
@@ -0,0 +1,128 @@
...
+        if sum([measurements['concentration_nm'] is not None,
+                measurements['concentration_ngul'] is not None,
+                measurements['na_quantity'] is not None]) != 1:
+            self.errors['concentration'] = 'Only one option must be specified out  of the following: NA quantity, conc. ng/uL or conc. nM.'
+
+        source_sample_obj, self.errors['sample'], self.warnings['sample'] = get_sample_from_container(
+            barcode=source_sample['container']['barcode'],
+            coordinates=source_sample['coordinates'])
+
+        if source_sample_obj.concentration is None:
```
sebastianballesteros, line 44: We might wanna ad an `if source_sample_obj: `and then continue after line 44. If source sample is None then this will crash.

sebastianballesteros:
Also I realized that the code does not validate if the sample given is actually a library when library is chosen as `norm_choice `

PR #352, opened 2022-07-29
Fix Add a simpler way to access templates stored in Freezeman

nafiz1001:
https://206.12.92.46/issues/1189


frontend/src/components/samples/details/SampleDetailsProcessMeasurements.js
```diff
@@ -13,15 +13,20 @@ const mapStateToProps = state => ({
 const SampleDetailsProcessMeasurements = ({processMeasurements, usersByID, protocolsByID,}) => {
     const columns = [
       {
-        title: '',
+        title: 'ID',
```
UlysseFG, line 16: I would change this to "Sample Process ID" to make it clearer of the ID this designate.

UlysseFG (approved):

PR #353, opened 2022-07-29
Switch to pytest

nafiz1001:


nafiz1001:
For now I'm stuck with only 182 tests being run when on master it's actually 183. Not sure which tests did not get run...

UlysseFG (approved):

nafiz1001:
Can you check if you can debug with vscode with this new test runner? @ckostiw

ckostiw (approved):
Tests can still be debugged 👍

PR #354, opened 2022-08-02
Fix process properties

nafiz1001:


backend/fms_core/serializers.py
```diff
@@ -231,6 +233,17 @@ class Meta:
...
         return PropertyValue.objects.filter(object_id=obj.id, content_type=process_content_type).values_list('id', flat=True)
+    
+    def get_sample_property_types(self, obj):
+        protocol_content_type = ContentType.objects.get_for_model(Protocol)
+        property_types = PropertyType.objects.filter(object_id=obj.protocol.id, content_type=protocol_content_type)
+        
+        process_measurement_type = ContentType.objects.get_for_model(ProcessMeasurement)
+        property_values = PropertyValue.objects.filter(property_type__in=property_types, content_type=process_measurement_type)
+
+        property_types = property_types.filter(Exists(property_values.filter(property_type=OuterRef("pk"))))
```
UlysseFG, line 244: I'm not sure its a good idea to filter this. Endpoints should not solve a specific punctual problem you have on the frontend ... It should give you the tools to solve the problems. The more business logic you put in at this level the more different end point you'll need and the harder it will be to actually reuse the endpoint.

There is a few things here i would change :
- Set this on the protocol ... the types are semantically tied to the protocol not the process.
- List all of them no matter they have values or not for your frontend case.
- Create a property_type serializer and nest it optionally within the protocol (see NestedSampleSerializer).
- Add the model they are tied to (going through a single property_value) that way you can tell if they are shared(process) or specific(processMeasurement). You could also make 2 endpoint with telling names to make it clear which one you interrogate, but it's not as generic.

UlysseFG (approved):

PR #358, opened 2022-08-04
Configure redux logger

ckostiw:
Configure the redux logger to log actions as 'info' messages so that we can show or hide them in the browser console. By default, the redux logger logs them as 'log' messages which cannot be filtered in the console, so the console is flooded with noise, making it tiresome to check for errors and warnings. With this change, we can hide the redux action logging by disabling 'info' messages in the console.


ckostiw:
Closing pull request. In prod, redux is logged as info, so it's only in dev that the console is noisy. I have to see how to change that for dev.

PR #360, opened 2022-08-04
Fix corrupt imported template downloads RM#1199

ckostiw:
Fixes redmine ticket https://206.12.92.46/issues/1199

The original problem seems to be caused by the Content-Type being set to None for the downloaded file.

We are using the `mimetypes` python module to determine a content type for the template file. By default, mimetypes gets the mime types from the system on which it is running. On our dev machines, it is getting it from MacOS or Linux. I guess that on the server there is no mapping for xlsx files, and the `mimetypes.guess_types()` call is returning None.

My simple fix is to default to `application/octet-stream` if guess_types returns None.

I am also going to see how complicated it is to configure the mimetype for xlsx files, to return the proper type without defaulting to system types. `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`

### DIFF
--- a/backend/fms_core/viewsets/imported_file.py
+++ b/backend/fms_core/viewsets/imported_file.py
@@ -45,7 +45,7 @@ class ImportedFileViewSet(viewsets.ModelViewSet):
         try:
             with open(file_path, "rb") as file:
                 response = HttpResponse(content=file)
-                response["Content-Type"] = mime_type_guess[0]
+                response["Content-Type"] = mime_type_guess[0] if mime_type_guess[0] != None else 'application/octet-stream'
                 response["Content-Disposition"] = "attachment; filename=" + filename
         except Exception as err:
             print(err)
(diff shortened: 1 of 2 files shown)


ckostiw:
Update: I played with mimetypes in the console and discovered that if you don't `init()` mimetypes then it only has a small list of common mimetypes. If you call `init()` then it has a much larger list which includes a mapping for `.xlsx` files.
I've added a call to `mimetypes.init()` in the the constructor to ensure the types are initialized. According to the mimetypes documentation, we can call `init()` as often as we like.

backend/fms_core/viewsets/imported_file.py
```diff
@@ -45,7 +50,7 @@ def download(self, _request, pk) -> Response:
         try:
             with open(file_path, "rb") as file:
                 response = HttpResponse(content=file)
-                response["Content-Type"] = mime_type_guess[0]
+                response["Content-Type"] = mime_type_guess[0] if mime_type_guess[0] != None else 'application/octet-stream'
```
UlysseFG, line 53: We could also replace the mime_type_guess and send everything as an octet-stream and let the frontend sort.

UlysseFG (approved):

PR #363, opened 2022-08-08
Added a new column for the full location in the sample export

sebastianballesteros:


UlysseFG:
Looks good, but i am not certain about the verbose format ... I find it a bit hard to parse ... you actually need to read in backward also to track the position. 

I think the flow would be hard to reverse efficiently but maybe we can separate each level more clearly with "<<" and maybe use "@" instead of "at".
e.g. EQ305208358 (tube) << HostSeq_3393@B01 (tube rack 8x12) << ... 

Naming the column "Container Full Location" would also explain why the sample position is not there.

One last thing ... would this be useful to the lab in the container export too ?

sebastianballesteros:
I totally agree with you. I will send you Anne-Marie's answer and we can discuss it.

PR #364, opened 2022-08-09
Use Hash in URL for Tabs

nafiz1001:


UlysseFG:
It is looking good. The only thing i saw that was a bit strange is that when you backtrack in the browser the previous tab stays bluish instead of becoming grey... If you click the page it goes away. Not sure that is really an issue.
I would add that hash ref to all the sections (container, sample, individual, project, experiment, ...).

nafiz1001:
> It is looking good. The only thing i saw that was a bit strange is that when you backtrack in the browser the previous tab stays bluish instead of becoming grey... If you click the page it goes away. Not sure that is really an issue. I would add that hash ref to all the sections (container, sample, individual, project, experiment, ...).

It happens to me when i backtrack and mouse click the tabs at the same time.

ckostiw:
I applied this to all of the pages with <Tab> elements. It should be ready to merge.

ckostiw (commented):
Ready for a final review. All <Tabs> views are now hooked up to the url.

UlysseFG (approved):

PR #367, opened 2022-09-02
Quick fix for infinite loop of item requests

ckostiw:
This is a quick fix that solves the immediate problem of a bad filter value causing an infinite loop of requests from a PaginatedTable, but which just makes the whole system a bit more complicated and difficult to understand. This could be merged if a better fix is going to take a long time to implement.

### DIFF
--- a/backend/fms_core/viewsets/_constants.py
+++ b/backend/fms_core/viewsets/_constants.py
@@ -42,6 +42,7 @@ _individual_filterset_fields: FiltersetFields = {
 }
 
 _user_filterset_fields: FiltersetFields = {
+    "id": PK_FILTERS,
     "username": FREE_TEXT_FILTERS,
     "email": FREE_TEXT_FILTERS,
 }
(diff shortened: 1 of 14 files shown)


UlysseFG (approved):
Made a couple small additions to make ID search work for all sections.

ckostiw:
I added a new FILTER_TYPE.INPUT_OBJECT_ID to use explicitly for object id filters, and added validation on the input to block the user from setting a garbage value and generating a bad request error from the server.

UlysseFG:
> I added a new FILTER_TYPE.INPUT_OBJECT_ID to use explicitly for object id filters, and added validation on the input to block the user from setting a garbage value and generating a bad request error from the server.
FILTER_TYPE.INPUT_NUMBER was created for IDs ;) I am not sure it is helpful to have a filter type for each column eheh. The validation idea is nice and applies to any input_number.

PR #370, opened 2022-09-08
Protocol pooling

UlysseFG:


ckostiw (commented):
For the template, in the SamplesToPool sheet, Pool Name should probably be the first column. It's a bit strange that it is in between "Source Container Coord" and "Source Depleted".

backend/fms_core/services/sample.py
```diff
@@ -304,6 +304,108 @@ def extract_sample(process: Process,
...
+                if volume_used <= 0:
+                    errors.append(f"Volume used ({volume_used}) is invalid.")
+                if sample_obj and volume_used > sample_obj.volume:
+                    errors.append(f"Volume used ({volume_used} uL) exceeds the current volume of the sample ({sample_obj.volume} uL).")
+
+            sample["Volume Ratio"] = volume_used / volume # Calculate the volume ratio of each sample in the pool
+            sample["Fractional Concentration"] = sample["Volume Ratio"] * sample_obj.concentration
+            sample_obj.volume = decimal_rounded_to_precision(sample_obj.volume - volume_used)  # Reduce the volume of source samples by the volume used
+            sample_obj.save()
+        concentration = sum(sample["Fractional Concentration"] for sample in samples_info) # Calculate pool concentration using fractional concentration
```
ckostiw, line 345: Is it possible that the lab will want to normalize the pool concentration during the pooling step?

backend/fms_core/services/sample.py
```diff
@@ -304,6 +304,108 @@ def extract_sample(process: Process,
...
+        concentration = decimal_rounded_to_precision(concentration)
+        # Create a a new sample
+        pool_data = dict(
+                name=pool_name,
+                volume=volume,
+                creation_date=execution_date,
+                container=container_destination,
+                comment=f"Automatically generated by sample pooling on {datetime.utcnow().isoformat()}Z",
+                **(dict(depleted=sample["Source Depleted"]) if sample["Source Depleted"] is not None else dict()),
```
ckostiw, line 354: My linter complains that the sample variable here is unbound - is this a copy/paste error?

backend/fms_core/template_importer/importers/sample_pooling.py
```diff
@@ -0,0 +1,79 @@
...
+
+            (result, row_object) = self.handle_row(
+                row_handler_class=SamplesToPoolRowHandler,
+                sheet=samplestopool_sheet,
+                row_i=i,
+                **samplestopool_kwargs,
+            )
+            pools_dict[str_cast_and_normalize(row_data['Pool Name'])].append(row_object)
+
+        if not result['validation_error'].messages: # prevent the repetition of error messages for volume_used at the level of pools.
```
sebastianballesteros, line 44: This will crash because result is initialized inside the loop, so here it is referenced before assignment

backend/fms_core/template_importer/importers/sample_pooling.py
```diff
@@ -0,0 +1,79 @@
...
+                'comment': str_cast_and_normalize(row_data['Comment']),
+            }
+
+            (result, row_object) = self.handle_row(
+                row_handler_class=SamplesToPoolRowHandler,
+                sheet=samplestopool_sheet,
+                row_i=i,
+                **samplestopool_kwargs,
+            )
+            pools_dict[str_cast_and_normalize(row_data['Pool Name'])].append(row_object)
```
sebastianballesteros, line 42: Maybe adding a validation that the Pool Name cell is filled so you don't populate pools_dict with invalid values.

backend/fms_core/services/sample.py
UlysseFG: It is more like people will want to pool after/during the normalization protocol. We'll be adding the pooling to the normalization in the next sprint. Would not make much sense to also add the normalization to the pooling.

backend/fms_core/services/sample.py
UlysseFG: This is not the right place to set this ... this should be in the loop above ... Your linter is right ... You can pat it on the back...

backend/fms_core/template_importer/importers/sample_pooling.py
```diff
@@ -0,0 +1,89 @@
...
+            (result, row_object) = self.handle_row(
+                row_handler_class=SamplesToPoolRowHandler,
+                sheet=samplestopool_sheet,
+                row_i=i,
+                **samplestopool_kwargs,
+            )
+            if pool_name is not None:
+                pools_dict[pool_name].append(row_object)
+
+        if result and not result['validation_error'].messages: # prevent the repetition of error messages for volume_used at the level of pools.
```
sebastianballesteros, line 55: There's a validation problem here. The `result` variable is only referring to the last row of samples validated (since it is overwritten at each iteration), if there's a sample object not found in any line but the last line it will continue and there will be problems and crashes in the row handler i.e. when accessing sample.is_library because sample is going to be None.

backend/fms_core/template_importer/row_handlers/sample_pooling/pool.py
```diff
@@ -0,0 +1,62 @@
...
+from fms_core.services.container import get_container, get_or_create_container
+
+class PoolsRowHandler(GenericRowHandler):
+    def __init__(self):
+        super().__init__()
+
+    def process_row_inner(self, protocol, imported_template, samples_info, pool, pooling_date, comment):
+        # Ensure there is samples_tied to the pool
+        if samples_info is None:
+            self.errors["source_sample"].append(f"Cannot find samples for pool {pool['name']}. "
```
sebastianballesteros, line 15: In addition, this is a different row handler so self.errors is an empty object and self.errors["source_sample"] is undefined (not an empty list) so you won't be able to append to it and it will crash.

backend/fms_core/template_importer/importers/sample_pooling.py
UlysseFG: yeah this is part of the issue we need to fix when we have the time. We need to have row_handler combine errors into the importer own errors to have a proper error hierarchy in there.

backend/fms_core/template_importer/row_handlers/sample_pooling/pool.py
```diff
@@ -0,0 +1,68 @@
...
+
+class PoolsRowHandler(GenericRowHandler):
+    def __init__(self):
+        super().__init__()
+
+    def process_row_inner(self, protocol, imported_template, samples_info, pool, pooling_date, comment):
+        # Ensure there is samples_tied to the pool
+        if samples_info is None:
+            self.errors["source_sample"] = (f"Cannot find samples for pool {pool['name']}. "
+                                            f"Make sure SampleToPool sheet Pool Name column values "
```
ckostiw, line 16:
```suggestion
                                            f"Make sure SamplesToPool sheet Pool Name column values "
```

ckostiw (changes requested):

backend/fms_core/models/sample.py
```diff
@@ -69,10 +69,18 @@ def is_library(self) -> bool:
...
-    def avg_library_size(self) -> Union[Decimal, None]:
-        return None if not self.is_library or any(derived_by_sample.derived_sample.library.library_size is None for derived_by_sample in self.derived_by_samples.all()) \
-                    else decimal_rounded_to_precision(sum(derived_by_sample.volume_ratio * derived_by_sample.derived_sample.library.library_size \
-                    for derived_by_sample in self.derived_by_samples.all()))
+    def sample_kind_name(self) -> str:
+        POOL_KIND_NAME = "POOL"
+
+        if self.is_library:
+            if self.is_pool:
+                sample_kind_name = POOL_KIND_NAME
```
sebastianballesteros, line 77: It might be better to add POOL_KIND_NAME to constants so it can be used elsewhere when calling this property and checking if it is a pool (like in sampleQC). Also, this will return "POOL" so we will still need to check the sample kind of the pool.

sebastianballesteros (approved):

ckostiw (approved):

PR #371, opened 2022-09-09
Pool protocol compatibility

sebastianballesteros:


backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -84,11 +86,14 @@ def process_row_inner(self, sample_container, measures, process, process_measure
...
             self.errors['concentration'] = 'Concentration must be specified in either nM or ng/uL, not both.'
 
+        # TODO: if it is a pool, library size is not provided. Should we force concentration in uL?
         concentration = measures['concentration_uL']
         if concentration is None:
             concentration = measures['concentration_nm']
             if concentration is not None:
-                molecular_weight = source_sample_obj.derived_sample_not_pool.library.molecular_weight_approx
+                # TODO: for pool, should we take the first one?
+                molecular_weight = source_sample_obj.derived_samples.first().library.molecular_weight_approx
```
sebastianballesteros, line 95: To check with Janick

backend/fms_core/template_importer/row_handlers/normalization_planning/normalization_planning.py
```diff
@@ -60,13 +61,16 @@ def process_row_inner(self, source_sample, destination_sample, measurements, rob
...
                 combined_concentration_nguL = concentration_nguL
             elif measurements['concentration_nm'] is not None:
-                if not source_sample_obj.is_library:
-                    self.errors['concentration'] = 'Concentration in nM cannot be used to normalize samples that are not libraries.'
+                if not source_sample_obj.is_library and not source_sample_obj.is_pool_of_libraries:
+                    self.errors['concentration'] = 'Concentration in nM cannot be used to normalize samples that are not libraries or pool of libraries.'
                 else:
                     concentration_nm = measurements['concentration_nm']
+                    # TODO: for pool, should we take the first one?
+                    library_obj = source_sample_obj.derived_samples.first().library
```
sebastianballesteros, line 69: To check with Janick

backend/fms_core/models/sample.py
```diff
@@ -66,8 +66,13 @@ def is_pool(self) -> bool:
 
     @property
     def is_library(self) -> bool:
+        # TODO: any?
```
UlysseFG, line 69: I think this is the definition i would keep..

backend/fms_core/models/sample.py
```diff
@@ -66,8 +66,13 @@ def is_pool(self) -> bool:
...
         return True if any([derived_sample.library is not None for derived_sample in self.derived_samples.all()]) else False
 
+    @property
+    def is_pool_of_libraries(self) -> bool:
```
UlysseFG, line 73: Not sure we need that since its exactly is_pool and is_library.

backend/fms_core/models/sample_lineage.py
```diff
@@ -31,16 +31,17 @@ def add_error(field: str, error: str):
...
+                if not child_derived.sample_kind.is_extracted:
+                    add_error("sample_kind", "Extracted sample need to be a type of Nucleic Acid.")
+                if parent_derived.sample_kind.is_extracted:
+                    add_error("extracted_from",
+                              "Extraction process cannot be run on samples of extracted kinds like DNA and RNA.")
+                if not child_derived.tissue_source:
+                    add_error("tissue_source", "Extracted sample need to have a tissue source.")
+                elif child_derived.tissue_source != parent_derived.sample_kind:
+                    add_error("tissue_source", "Extracted sample tissue_source must match parent sample_kind.")
         elif protocol_name == "Transfer":
```
UlysseFG, line 45: We would need to add validation here for "Sample Pooling" too. We should check that parent.derived_sample are a subset of the child.derived_sample. Also "Experiment Run" should share the current validation with "Transfer".

backend/fms_core/models/sample_lineage.py
```diff
@@ -31,16 +31,17 @@ def add_error(field: str, error: str):
...
-            if parent_derived.sample_kind.is_extracted:
-                add_error("extracted_from",
-                          "Extraction process cannot be run on samples of extracted kinds like DNA and RNA.")
-            if not child_derived.tissue_source:
-                add_error("tissue_source", "Extracted sample need to have a tissue source.")
-            elif child_derived.tissue_source != parent_derived.sample_kind:
-                add_error("tissue_source", "Extracted sample tissue_source must match parent sample_kind.")
+
+            if not errors:
+                if not child_derived.sample_kind.is_extracted:
```
UlysseFG, line 36: If top ups apply to sample that are not extracted, all of this will explode.

backend/fms_core/services/library.py
```diff
@@ -83,11 +83,8 @@ def convert_library(process, platform, sample_source, container_destination, coo
         errors.append(f"Execution date is not valid.")
 
     # Retrieve library object linked to the source sample
-    library_source_obj = sample_source.derived_sample_not_pool.library
-    if not sample_source.is_library or library_source_obj is None:
-        errors.append(f"Sample {sample_source.name} is not a library.")
-    elif library_source_obj.platform == platform:
-        errors.append(f"Source library platform and destination library platform can't be the same.")
+    if not sample_source.is_library and not sample_source.is_pool_of_libraries:
```
UlysseFG, line 86: this can be tested as `not is_library`.

backend/fms_core/services/library.py
```diff
@@ -109,6 +106,13 @@ def convert_library(process, platform, sample_source, container_destination, coo
             derived_samples_destination = []
             volume_ratios = {}
             for derived_sample_source in sample_source.derived_samples.all():
+                library_source_obj = derived_sample_source.library
+                # extra validation
+                if library_source_obj is None:
+                    errors.append(f"Sample {sample_source.name} is not a library or pool of libraries. ")
```
UlysseFG, line 112: This should be reported differently. It means there is a mix of libraries and samples in the pool. I would report it as "Pool {sample_source.name} contains a sample {derived_sample_source.biosample.alias} that is not converted into a library." Since you tested for library already a pool is the only case you'll get here.

backend/fms_core/services/library.py
```diff
@@ -145,41 +149,44 @@ def convert_library(process, platform, sample_source, container_destination, coo
...
+
 def update_library(sample, **kwargs):
     errors = []
     warnings = []
 
     if sample is None:
         errors.append('Missing sample')
-    elif not sample.is_library:
-        errors.append('Sample {sample.name} is not a library')
+    elif not sample.is_library and not sample.is_pool_of_libraries:
```
UlysseFG, line 159: you just need to change the message.

backend/fms_core/services/library.py
```diff
@@ -145,41 +149,44 @@ def convert_library(process, platform, sample_source, container_destination, coo
 
     return library_destination, errors, warnings
 
+
 def update_library(sample, **kwargs):
```
UlysseFG, line 153: You don't normally want to overwrite a whole bunch of different libraries with the same values. In fact it might be wise to remove the service altogether, or replace the signature by update_library(derived_sample, **kwargs) to be used during capture or inside a controlled loop (applying the change while testing the content of the current library). At the moment the only thing using this is lib_qc to change lib_size so i would test for pool outside and call this service with the appropriate derived_sample instead. We can keep the single library action here instead of the loop. It makes more sense i think.

backend/fms_core/services/sample.py
```diff
@@ -149,6 +149,7 @@ def inherit_sample(sample_source, new_sample_data, derived_samples_destination,
                                            volume_ratio=volume_ratios[derived_sample_destination.id])
         
         # project inheritance
+        # TODO: Move project to derived?
```
UlysseFG, line 152: I actually patched that in my PR so it does not try to create multiple association for the same project. I think we'll be moving the project to Derived_sample. In that situation, this part will become obsolete and we'll remove it in the project migration PR.

backend/fms_core/services/sample.py
```diff
@@ -307,7 +308,7 @@ def extract_sample(process: Process,
 def prepare_library(process: Process,
                     sample_source: Sample,
                     container_destination: Container,
-                    library: Library,
+                    libraries_by_derived_sample,
```
UlysseFG, line 311: Add a docstring to the service to detail the dict structure (and the rest) to know what the function has to receive as input.

backend/fms_core/services/sample.py
```diff
@@ -353,12 +354,15 @@ def prepare_library(process: Process,
                 quality_flag=None
             )
 
-            new_derived_sample_data = {
-                "library_id": library.id
-            }
+
             derived_samples_destination = []
             volume_ratios = {}
+            # For pools of samples (a library for each derived sample)
```
UlysseFG, line 360: This is actually the generic way of doing it ;) should have been that way from the get-go. Let's blame the old self.

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
```diff
@@ -44,3 +44,6 @@ def process_row_inner(self, source_sample, resulting_sample, process_measurement
                                volume_destination=resulting_sample['volume'],
                                source_depleted=check_truth_like(source_sample['depleted']) if source_sample['depleted'] else None,
                                comment=process_measurement['comment'])
+
```
UlysseFG, line 47: extra line.

backend/fms_core/template_importer/row_handlers/extraction/extraction.py
```diff
@@ -31,7 +31,7 @@ def process_row_inner(self, source_sample, resulting_sample, process_measurement
         original_sample, self.errors['sample'], self.warnings['sample'] = get_sample_from_container(barcode=source_sample['container']['barcode'],
                                                                                                     coordinates=source_sample['coordinates'])
 
-        if original_sample:
+        if original_sample and not original_sample.is_pool:
```
UlysseFG, line 34: pending the answer on top up of pre-extraction kinds...

backend/fms_core/template_importer/row_handlers/library_preparation/library_preparation.py
```diff
@@ -27,7 +27,17 @@ def process_row_inner(self, library_batch_info, source_sample, volume_used,
             self.errors['volume_used'] = f"Volume used ({volume_used}) exceeds the current volume of the sample ({source_sample_obj.volume})"
 
         if source_sample_obj:
+            # Check if sample is not a library or a pool of libraries
+            if source_sample_obj.is_library or source_sample_obj.is_pool_of_libraries:
```
UlysseFG, line 31: is_library is sufficient.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -29,8 +29,8 @@ def process_row_inner(self, sample_container, measures, process, process_measure
             self.errors['library'] = f"Library sample for QC was not found in container ({barcode}) at ({coordinates})"
             return
 
-        if not source_sample_obj.is_library:
-            self.errors['sample'] = f'The sample {source_sample_obj.name} at {barcode}@{coordinates} is not a library '
+        if not source_sample_obj.is_library and not source_sample_obj.is_pool_of_libraries:
```
UlysseFG, line 32: not is_library sufficient for the test.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -73,8 +73,9 @@ def process_row_inner(self, sample_container, measures, process, process_measure
                 self.errors['library_volume'] = f'The library\'s computed final volume would be less than zero ({final_volume}). Please verify the volume currently stored for the library.'
 
         # library size
+        # TODO: I modified to optional if sample is pool. Is the assumption correct?
```
UlysseFG, line 76: yes as far as i know, lib_qc on a pool will not be setting lib_size since they should have all been QCed before being pooled. The QC will update the concentration as nM and volume. That mean we need to use the partial library_size (and molecular_weigth) to calculate each contained library concentration as ng/uL then use the volume ratio to store the concentration. This will likely require a wrapper service for pools concentration calculations.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -134,7 +138,7 @@ def process_row_inner(self, sample_container, measures, process, process_measure
 
         # set the library size on the library
         _, self.errors['library-size'], self.warnings['library-size'] = \
-           update_library(source_sample_obj, **{'library_size':library_size})
+            update_library(source_sample_obj, **{'library_size': library_size})
```
UlysseFG, line 141: We need to skip this update for pools or we will overwrite library_size with None for pools which is not the desired result (which is keeping the current value).

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -84,11 +85,14 @@ def process_row_inner(self, sample_container, measures, process, process_measure
         if measures['concentration_nm'] is not None and measures['concentration_uL'] is not None:
             self.errors['concentration'] = 'Concentration must be specified in either nM or ng/uL, not both.'
 
+        # TODO: if it is a pool, library size is not provided. Should we force concentration in uL?
```
UlysseFG, line 88: This part should be moved bellow the update_library library size for non-pools. That way we can use uniformly the stored library_size (don't forget to update the object from db) for concentration calculations.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -84,11 +85,14 @@ def process_row_inner(self, sample_container, measures, process, process_measure
...
         concentration = measures['concentration_uL']
         if concentration is None:
             concentration = measures['concentration_nm']
             if concentration is not None:
-                molecular_weight = source_sample_obj.derived_sample_not_pool.library.molecular_weight_approx
+                # TODO: for pool, we should come with a better way to computer molecular weight
+                molecular_weight = source_sample_obj.derived_samples.first().library.molecular_weight_approx
```
UlysseFG, line 94: This cannot be assumed. molecular_weight is related mainly to strandedness (it's already a bit of an approximation). If we mix 50-50 sample of different strandedness the approximation will simply be wrong. The best way to deal with this i think is to fetch for each derived_sample the library molecular weight, library_size and volume ratio. You call the conversion for each derived sample then you multiply each by the volume ratio and add each fractional concentration up to get the whole pool concentration in ng/uL. You can wrap this in a service to have a single call here (with source_sample_obj and concentration as params) since this will be true even for a single library (volume_ratio being 1)

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -46,8 +46,10 @@ def process_row_inner(self, source_sample, destination_sample, process_measureme
             # Case when nM is given
             elif destination_sample['concentration_nm']:
                 if source_sample_obj.is_library:
+                    # TODO: Adapt so we use avg. library size of a pool
                     library = source_sample_obj.derived_sample_not_pool.library
                     if library.library_size:
+                        # TODO: improve conversion service to include pool of libraries
```
UlysseFG, line 52: Use same service as lib_qc (convert_sample_concentration_from_nm_to_ngbyul(sample, concentration). We have to remember to have an comparable service for the reverse operation for the frontend.

backend/fms_core/template_importer/row_handlers/normalization/normalization.py
```diff
@@ -46,8 +46,10 @@ def process_row_inner(self, source_sample, destination_sample, process_measureme
             # Case when nM is given
             elif destination_sample['concentration_nm']:
                 if source_sample_obj.is_library:
+                    # TODO: Adapt so we use avg. library size of a pool
```
UlysseFG, line 49: We'll try to get around without avg library size. The less approximation we introduce the better. If this is too slow for the frontend display we'll have one just for the reverse conversion (just for display/endpoint).

backend/fms_core/template_importer/row_handlers/normalization_planning/normalization_planning.py
```diff
@@ -46,7 +46,8 @@ def process_row_inner(self, source_sample, destination_sample, measurements, rob
                 self.errors['concentration'] = f'A sample or library needs a known concentration to be normalized. QC sample {source_sample_obj.name} first.'
 
             # ensure that the sample source is a library if the norm choice is library
-            if robot['norm_choice'] == LIBRARY_CHOICE and not source_sample_obj.is_library:
+            # If it is a pool we have to check if it is a pool of libraries
```
UlysseFG, line 49: is_library should still cover that case ... ( a logical or between is_library and is_pool_library will always be true if is_library is true and false when is_library is false )

backend/fms_core/template_importer/row_handlers/normalization_planning/normalization_planning.py
```diff
@@ -60,13 +61,16 @@ def process_row_inner(self, source_sample, destination_sample, measurements, rob
                 na_qty = decimal.Decimal(measurements['volume']) * decimal.Decimal(concentration_nguL)
                 combined_concentration_nguL = concentration_nguL
             elif measurements['concentration_nm'] is not None:
-                if not source_sample_obj.is_library:
-                    self.errors['concentration'] = 'Concentration in nM cannot be used to normalize samples that are not libraries.'
+                if not source_sample_obj.is_library and not source_sample_obj.is_pool_of_libraries:
```
UlysseFG, line 64: is_library is enough.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -22,6 +22,10 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
         )
 
         if sample_obj:
+            # Check if sample is not a library or a pool of libraries
+            if sample_obj.is_library or sample_obj.is_pool_of_libraries:
```
UlysseFG, line 26: is_library is sufficient

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -72,7 +76,9 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
                         self.errors['instrument'] = f'Invalid instrument {type}.'
 
                 # Validate required RIN for RNA
-                if sample_obj.derived_sample_not_pool.sample_kind.name == 'RNA' and process_measurement_properties['RIN']['value'] is None:
-                    self.errors['RIN'] = 'RIN has to be specified for RNA.'
+                # For pools we check all sample kinds
+                for derived_sample in sample_obj.derived_samples.all():
+                    if derived_sample.sample_kind.name == 'RNA' and process_measurement_properties['RIN']['value'] is None:
```
UlysseFG, line 81: I think this may not be the optimal way to deal with this. Users will receive multiple errors for a single pooled sample if the sample kind is RNA. First sample pools should not contain different sample_kind. I think we should enforce that (i have to double check in the current pool service if it is the case). In this case i think we should make a sample property that return a single sample kind for sample pools (they should all be the same) and it would return pool for pooled libraries (meaning we don't care too much at this point and we are not enforcing anything about those).
Here you should be able to ask sample_obj.sample_kind == "RNA" and ...: a single time for your sample and have a single error message.

PR #373, opened 2022-09-14
Protocol pooling tests

UlysseFG:
- [x] Sample pooling Template  submission test
- [ ] pool_sample service test


backend/fms_core/tests/test_template_importers/sample_pooling_test.py
```diff
@@ -0,0 +1,191 @@
...
+        # platform
+        platform, _, _ = get_platform(name="ILLUMINA")
+
+
+        library_1, _, _ = create_library(index=self.index_1, library_type=library_type, platform=platform, strandedness=DOUBLE_STRANDED, library_size=150)
+        library_2, _, _ = create_library(index=self.index_2, library_type=library_type, platform=platform, strandedness=DOUBLE_STRANDED, library_size=150)
+        library_3, _, _ = create_library(index=self.index_3, library_type=library_type, platform=platform, strandedness=DOUBLE_STRANDED, library_size=150)
+
+        self.source_sample_1, _, _ = \
+            create_full_sample(name="SOURCESAMPLE1POOL1", alias="SOURCESAMPLE1POOL1", volume=self.source_sample_initial_volume, concentration=25,
```
ckostiw, line 57: It might be good to have some non-integer values in our tests (eg. 25.333). I see linter complaints about illegal decimal/integer conversions, but I assume that the linter is just confused. Still, it would be good to test that decimal values are not being accidentally rounded down to integer values in our code.

UlysseFG: Trouble is that comparing floats and decimal is not very safe (25.333 may not be equal Decimal('25.333'). If we use fractional values, we should use Decimal instead which in turn defeats the purpose.

UlysseFG:
Closing to create a new branch to clean up history and skip rebasing again.

PR #376, opened 2022-09-19
Project migration to derived

UlysseFG:


backend/fms_core/viewsets/sample.py
```diff
@@ -285,224 +284,20 @@ def update(self, request, *args, **kwargs):
...
-                'created_at': sample["created_at"],
-                'updated_by': sample["updated_by"],
-                'updated_at': sample["updated_at"],
-                'deleted': sample["deleted"],
-                'comment': sample["comment"],
-            }
-            serialized_data.append(data)
-
+        serialized_data = fetch_sample_data([], samples_queryset, self.request.query_params)
         return Response(serialized_data)
```
sebastianballesteros, line 301: change this line to: 

`return Response({
            'results': serialized_data,
            'count': samples_queryset.count()
  })
`

Change a few lines in the SampleListComponent in the frontend i.e. the render method of sample's project in the column list to :

`
render: (_, sample) => (sample.projects &&
        <Link to={`/projects/${sample.projects}`}> {projectsByID[sample.projects]?.name} </Link>
      ),
`


And the list will show. There's a a few details that might have to change to be able to show all the fields.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -0,0 +1,133 @@
...
+            'quantity_flag',
+            'depleted',
+            'comment',
+            'is_library',
+            'derived_count',
+            'first_derived_sample',
+            'derived_samples__project',
+            'child_of',
+            'extracted_from',
+            'process_measurement',
```
sebastianballesteros, line 70: for process_measurements you would need .values_list() and add it manually to the sample's query set (the frontend expects a list)

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -0,0 +1,133 @@
...
+            'concentration': sample["concentration"],
+            'child_of': sample["child_of"],
+            'extracted_from': sample["extracted_from"] if not is_pool else None,
+            'individual': derived_sample["biosample__individual_id"] if not is_pool else None,
+            'container': sample["container_id"],
+            'coordinates': sample["coordinates"],
+            'sample_kind': derived_sample["sample_kind_id"] if not is_pool or (is_pool and not is_library) else "POOL",
+            'is_library': is_library,
+            'is_pool': is_pool,
+            'projects': sample["derived_samples__project"],
```
sebastianballesteros, line 117: Might be better to change to 'project'

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -0,0 +1,133 @@
...
+            'process_measurements': sample["process_measurement"],
+            'tissue_source': derived_sample["tissue_source_id"] if not is_pool else None,
+            'creation_date': sample["creation_date"],
+            'collection_site': derived_sample["biosample__collection_site"] if not is_pool else None,
+            'experimental_group': json.dumps(derived_sample["experimental_group"]) if not is_pool else None,
```
sebastianballesteros, line 122: I am not sure about this one but I think the front end expects an array too.

backend/fms_core/tests/test_template_importers/project_link_samples_test.py
```diff
@@ -75,12 +75,11 @@ def test_import(self):
...
 
         #Custom tests for each template
-        self.assertEqual(len(SampleByProject.objects.all()), 3)
-        self.assertTrue(SampleByProject.objects.filter(sample=self.sample1, project=self.project1).exists())
-        self.assertTrue(SampleByProject.objects.filter(sample=self.sample2, project=self.project1).exists())
-        self.assertTrue(SampleByProject.objects.filter(sample=self.sample2, project=self.project2).exists())
-        self.assertFalse(SampleByProject.objects.filter(sample=self.sample1, project=self.project2).exists())
-        self.assertFalse(SampleByProject.objects.filter(sample=self.sample3, project=self.project3).exists())
+        self.assertEqual(len(DerivedSample.objects.filter(project__isnull=False).all()), 2)
+        self.assertTrue(DerivedSample.objects.filter(samples__in=self.sample1, project=self.project3).exists())
```
sebastianballesteros, line 79: Change samples__in look up to just sample since it is not an array. Same for the next 3 lines.

backend/fms_core/viewsets/_fetch_data.py
UlysseFG: There is still many projects by sample potentially with the pools.

backend/fms_core/tests/test_template_importers/project_link_samples_test.py
UlysseFG: samples is the reverse relation of derived_samples... its likely to hold more than 1 sample. sample does not exist on DerivedSample... I'll test for "=" instead of "__in".

backend/fms_core/viewsets/sample.py
```diff
@@ -298,7 +298,7 @@ def retrieve(self, _request, pk=None, *args, **kwargs):
     def list(self, _request, *args, **kwargs):
         samples_queryset = self.filter_queryset(self.get_queryset())
         serialized_data = fetch_sample_data([], samples_queryset, self.request.query_params)
-        return Response(serialized_data)
+        return Response({"results": serialized_data, "count": len(serialized_data)})
```
sebastianballesteros, line 301: You want the count of the sample_queryset once filtered, not the serialized_data. Otherwise user will always see the first X (depending on limit) items and won't be able to use pagination.

sebastianballesteros (approved):

PR #381, opened 2022-09-23
Ux sample details pool

ckostiw:
Display pooled samples in a Pool tab in the Sample Details page.


frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -328,6 +329,13 @@ const SampleDetailsContent = ({
         <TabPane tab={`Lineage`} key="6" style={tabStyle}>
           <SampleDetailsLineage sample={sample} />
         </TabPane>
+
+        // Only display the Pool tab if the sample is a pool
+        {sample.is_pool &&
```
UlysseFG, line 334: So far we've adopted the position to always show all tabs and mention they are not required in them inside the tab if the tab does not apply. Maybe our position could be revised but until then to keep things uniform i would move this test and add it inside the Pane to display a message in case this is not a pool.

frontend/src/components/filters/descriptions.js
```diff
@@ -572,3 +572,46 @@ export const DATASET_FILE_FILTERS = {
...
+      type: FILTER_TYPE.INPUT,
+      key: "derived_sample__library__library_type__name",
+      label: "Library Type"
+    },
+    library_size: {
+      type: FILTER_TYPE.RANGE,
+      key: "derived_sample__library__library_size",
+      label: "Library Size",
+    },
+    index: {
```
UlysseFG, line 610: There is some information that are missing from the table that could be useful. Individual name (with link to individual), experimental group, collection site would be nice and sample kind (if there is some space left).

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+from rest_framework.exceptions import APIException
+from fms_core.models import Biosample, DerivedSample, DerivedBySample, Index, Library, LibraryType, Project, Sample, SampleKind
+from fms_core.filters import PooledSampleFilter
+
+# Custom serializers
+# These serializers serialize the data we want to receive in pool samples.
+
+EXCLUDE_BOOKKEEPING = ['created_at', 'updated_at', 'created_by', 'updated_by', 'deleted']
+
+class IndexSerializer(serializers.ModelSerializer):
```
UlysseFG, line 15: Move the serializers to serializers.py, use the existing one as much as possible and add generics ones if you need. If you need  2 fields from each model you are better off making a custom serializer with just the fields you want those mini serializers will not see much use.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+from rest_framework.response import Response
+from rest_framework.pagination import LimitOffsetPagination
+from rest_framework.exceptions import APIException
+from fms_core.models import Biosample, DerivedSample, DerivedBySample, Index, Library, LibraryType, Project, Sample, SampleKind
+from fms_core.filters import PooledSampleFilter
+
+# Custom serializers
+# These serializers serialize the data we want to receive in pool samples.
+
+EXCLUDE_BOOKKEEPING = ['created_at', 'updated_at', 'created_by', 'updated_by', 'deleted']
```
UlysseFG, line 13: don't exclude those fields from serializers unless you do a serializer that is meant solely for the end user (model export for users)

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+    ''' Serializes a DerivedBySample object, representing a pooled sample. 
+        The result is a nested data structure containing all of the information about the
+        pooled sample needed by the frontend for display.
+    '''
+    volume_ratio = serializers.DecimalField(max_digits=20, decimal_places=3, read_only=True)
+    derived_sample = DerivedSampleSerializer()
+    # Since DerivedBySample doesn't have its own id field, we use the derived_sample id
+    # as a top level id in the returned data structure. The UX needs this for 'objectsById' stuff.
+    id = serializers.IntegerField(read_only = True, source='derived_sample.id')
+    parent_sample_name=serializers.SerializerMethodField()
```
UlysseFG, line 69: you can remove those parent infos if you don't use them.

frontend/src/modules/pooledSamples/reducers.js
```diff
@@ -0,0 +1,177 @@
...
+            return state;
+    }
+}
+
+function preprocess(pooledSample) {
+    pooledSample.isFetching = false
+    pooledSample.isLoaded = true
+
+    // Flatten the data structure for the UX
+   return flattenSample(pooledSample)
```
UlysseFG, line 102: Why not return a flat object ? you build a nested structure in the backend and you remove it in the frontend. We do not use much nesting in our structures you can do the same.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+        model = Library
+        exclude = EXCLUDE_BOOKKEEPING
+        depth = 1
+
+class ProjectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Project
+        fields = ['name', 'id']
+
+class DerivedSampleSerializer(serializers.ModelSerializer):
```
UlysseFG, line 46: You would be better off to list all fields directly than making those tiny serializers that won't be reused. It would also save you the effort of flattening your structure on the frontend side.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+        sample = Sample.objects.get(parent_of=obj.sample.id, derived_samples=obj.derived_sample.id)
+        return sample.name if sample is not None else ''
+
+
+class MissingPoolIDException(APIException):
+    status_code = 400
+    default_code = 'bad_request'
+    default_detail = 'No pool ID: query must include sample__id parameter containing the pool ID.'
+
+class PooledSamplesViewSet(viewsets.ModelViewSet):
```
UlysseFG, line 92: Build a generic viewset that can be used for other purpose than the pool detail. We may have a request down the line to list pools and export them as a csv. The current viewset would not work... The get_queryset need to be usable by list and retrieve both. You should add this filtering to an overloaded retrieve instead. Also this bit of code is likely not useful since you could just add the sample_id or sample__id to the filterset_fields and have the filtering take care of this.

frontend/package.json
```diff
@@ -65,6 +65,8 @@
     "cypress": "^10.3.0",
     "cypress-fail-fast": "^5.0.0",
     "cypress-file-upload": "^5.0.2",
+    "eslint": "^8.24.0",
```
UlysseFG, line 68: do we really need a linter ?

frontend/.eslintrc.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
UlysseFG, line 1: might want to add your linter to gitignore.

ckostiw: Normally linter configuration is part of the project as a team usually shares the same linter rules.

frontend/package.json
ckostiw: Yes, we should be using a linter in the frontend code.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,113 @@
+from rest_framework import viewsets, serializers
+from rest_framework.exceptions import APIException
+from fms_core.models import DerivedBySample, Sample
+from ._constants import (
+    _pooled_sample_filterset_fields,
+)
+from ._utils import _list_keys
+
+        
+class PooledSampleSerializer(serializers.Serializer):
```
UlysseFG, line 10: Move this to serializers.py

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,113 @@
...
+        parent_sample = Sample.objects.get(parent_of=obj.sample.id, derived_samples=obj.derived_sample.id)
+        return parent_sample.id if parent_sample is not None else ''
+
+    # Finds the id of the parent sample from which this pooled sample was derived.
+    def get_parent_sample_name(self, obj):
+        sample = Sample.objects.get(parent_of=obj.sample.id, derived_samples=obj.derived_sample.id)
+        return sample.name if sample is not None else ''
+
+
+# class MissingPoolIDException(APIException):
```
UlysseFG, line 80: cleanup commented out code.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,113 @@
...
+        pool sample. If not, MissingPoolIDException is raised.
+    '''
+    queryset = DerivedBySample.objects.all()
+    serializer_class = PooledSampleSerializer
+    filterset_fields = _pooled_sample_filterset_fields
+    ordering_fields = {
+        *_list_keys(_pooled_sample_filterset_fields),
+    }
+
+    # def get_queryset(self):
```
UlysseFG, line 101: clean up.

frontend/src/components/filters/descriptions.js
```diff
@@ -572,3 +572,71 @@ export const DATASET_FILE_FILTERS = {
...
+      type: FILTER_TYPE.INPUT,
+      key: "derived_sample__biosample__individual__name",
+      label: "Individual"
+    },
+    collection_site: {
+      type: FILTER_TYPE.INPUT,
+      key: "derived_sample__biosample__collection_site",
+      label: "Collection Site"
+    },
+    // TODO: The backend doesn't support filtering on a JSONField yet. Once that is fixed,
```
UlysseFG, line 633: Clean up.

frontend/src/components/samples/details/SampleDetailsPool.js
```diff
@@ -0,0 +1,198 @@
...
+        },
+        {
+            title: "Alias",
+            dataIndex: "alias",
+            sorter: true,
+            render: (_, pooledSample) => {
+                return (
+                    // The link points to the parent sample of the pooled derived sample, eg. the library
+                    // that was pooled.
+                    <Link to={`/samples/${pooledSample.parent_sample_id}`}>{pooledSample.alias}</Link>
```
UlysseFG, line 33: Move link to a parent sample ID column.

ckostiw: https://206.12.92.46/issues/1295

frontend/src/components/samples/details/SampleDetailsContent.js
ckostiw: Done

frontend/src/components/filters/descriptions.js
ckostiw: Done

backend/fms_core/viewsets/sample_pooled.py
ckostiw: Keeping them for now, will be replaced with a better recursive lookup in a separate ticket

frontend/src/components/filters/descriptions.js
```diff
@@ -572,3 +572,46 @@ export const DATASET_FILE_FILTERS = {
...
+
+export const POOLED_SAMPLES_FILTERS = {
+    sample__id: {
+        // Note: The sample__id filter adds the pool id to the query.
+        // It's not mapped to any column, but is required for the pool id to be sent
+        // to the endpoint properly.
+      type: FILTER_TYPE.INPUT_OBJECT_ID,
+      key: "sample__id",
+    },
+    project_name: {
```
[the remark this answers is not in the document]
ckostiw: I fixed that with a new serializer that maps my column key to a filter key.

backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -0,0 +1,117 @@
...
+
+        The request must include a 'sampled__id__in` query parameter specifying the id of the
+        pool sample. If not, MissingPoolIDException is raised.
+    '''
+    queryset = DerivedBySample.objects.all()
+    serializer_class = PooledSampleSerializer
+    filter_class = PooledSampleFilter
+
+    def get_queryset(self):
+        # Ensure that the pool id is specified to avoid trying to return all of the derived
```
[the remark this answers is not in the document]
ckostiw: I think that to retrieve the list of samples that are pools it would make more sense to add that capability to the Sample viewset (eg. add an 'is_pool' filter option to return samples that are pools). If the user wants to get the set of samples contained by a pool then they would use this endpoint.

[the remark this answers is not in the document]
UlysseFG: feel free to add a default filter on "is_pool" in get_queryset since this endpoint is intended to return pool information. I would not set it as an option though as this is really not useful for anything that is not a pool (it will force us to support code that is not useful).

frontend/src/components/samples/details/SampleDetailsPool.js
```diff
@@ -0,0 +1,217 @@
...
+        sample.is_pool ?
+            <PooledSamples sample={sample}/>
+        :
+            <NoPooledSamples/>
+    )
+}
+
+export const NoPooledSamples = () => {
+    return (
+        <div>This sample does not contain any pooled samples.</div>
```
UlysseFG, line 116: The message is a bit strange. I would go with "This is not a pooled sample." instead.

PR #384, opened 2022-10-03
Pools for sample submission

sebastianballesteros:


sebastianballesteros:
TODO: 

- Include pool submission in the template submission test.
- Once we have the final version of the template make the corrections and the right modifications.

backend/fms_core/services/sample.py
```diff
@@ -429,6 +429,120 @@ def pool_samples(process: Process,
...
+                       "experimental_group",
+                       "library",
+                       "project",
+                       "volume"}
+
+        pool_name: the name given to the pool by the user.
+        container_destination: a container object that will receive the pool.
+        coordinates_destination: the coordinate on the container where the pool is stored.
+        reception_date: The date which the pool was received .
+
```
UlysseFG, line 464: comment not listed in docstring args.

backend/fms_core/services/sample.py
```diff
@@ -429,6 +429,120 @@ def pool_samples(process: Process,
...
+        if pool_sample_obj:
+            for sample in samples_info:
+                volume_ratio = sample["volume_ratio"]
+
+                # Create Derived Samples and Biosamples for each sample
+                biosample_data = dict(
+                    collection_site=sample['collection_site'],
+                    **(dict(individual=sample['individual']) if sample['individual'] is not None else dict()),
+                    **(dict(alias=sample['alias']) if sample['alias'] is not None else dict()),
```
UlysseFG, line 513: Alias is no longer optional ... Make sure the Alias is set to the sample name,

backend/fms_core/services/sample.py
```diff
@@ -429,6 +429,120 @@ def pool_samples(process: Process,
...
+                )
+
+                try:
+                    biosample = Biosample.objects.create(**biosample_data)
+
+                    derived_sample_data = dict(
+                        biosample_id=biosample.id,
+                        sample_kind=sample['sample_kind'],
+                        **(dict(library=sample['library']) if sample['library'] is not None else dict()),
```
UlysseFG, line 522: Do not allow empty library to be part of a pool. This would be an issue since we do not have an integrity test for that. Make sure to test and set an error if library is None­.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -128,35 +109,75 @@ def process_row_inner(self, sample, library, container, project, parent_containe
...
+        project_obj = None
+        if project['name']:
+            project_obj, self.errors['project'], self.warnings['project'] = get_project(project['name'])
+
+        # Continue creating the sample objects if this sample is not associated with a pool
+        if library['pool_name'] is None:
+            # Check concentration fields given sample_kind (moved from sample and DerivedBySample because information unavailable until multiple relations created)
+            if sample['concentration'] is None and sample_kind_obj.concentration_required:
+                self.errors['concentration'] = [f"Concentration must be specified for a submitted sample if the sample_kind is DNA."]
+
```
UlysseFG, line 125: too many empty line.

backend/fms_core/services/sample.py
```diff
@@ -429,6 +429,124 @@ def pool_samples(process: Process,
...
+            sample_volume = sample["volume"]
+            # Calculate the volume ratio of each sample in the pool
+            sample["volume_ratio"] = decimal_rounded_to_precision(sample_volume / pool_volume)
+        # Create a a new sample - Concentration value is not set (need a QC to set it)
+        pool_data = dict(
+            name=pool_name,
+            volume=pool_volume,
+            creation_date=reception_date,
+            container=container_destination,
+            **(dict(comment=comment) if comment is not None else f"Automatically generated by pooling submission on {datetime.utcnow().isoformat()}Z"),
```
ckostiw, line 498:
```suggestion
            comment=comment if comment is not None else f"Automatically generated by pooling submission on {datetime.utcnow().isoformat()}Z"),
```
I don't understand why you would create a dict for the comment value and then unpack it. My linter complains that an unpacked dictionary is not a valid type of parameter for the dict() constructor.
Same thing for coordinates_destination.

ckostiw: Is this a python thing that I don't get?

ckostiw: Actually, only the comment line causes a problem, because if comment is none then you are trying to unpack a string, rather than an empty dict().

ckostiw: It's not just the linter that complains...

Screen Shot 2022-11-03 at 4 40 13 PM

UlysseFG: The dict unpacking is an approach that comes from kwargs argument unpacking i think in order to not set some values that were not to be updated by leaving the value empty. We usually have empty dict for case where the parameter is None in order to skip the attribute. This is an inline solution that is more readable that an if else tree, but this is overzealous unpacking ;). we should not do a dict just to break it down.

ckostiw: Ah, I see. It allows you to avoid adding keys with values of None to the dict. Instead, the keys are simply left out of the dict. I guess that helps if you want to call a function with the expanded dict as parameters, and the function defines default values for missing parameters. If a key is set to None then the function will use None for the value of the corresponding parameter, but if the key isn't present at all then the function will use any default value defined for the parameter.

ckostiw (approved):

UlysseFG:
Container Kind Data Validation (drop down) is not working correctly on the pool sheet for me.

UlysseFG:
When you save the template, expand the Location Barcode cell so the see Location not just Barcode.

UlysseFG:
Library Size Drop down has Standedness Validation values.

UlysseFG:
Library Size column format is Text instead of Numbers.

UlysseFG:
If you replace the sample submission destination container and Reception date, add conditional formating to darken the sampleSubmission page corresponding columns. 

Also PoolSubmission Reception date column is missing a Date in the name.

backend/fms_core/template_importer/row_handlers/sample_submission/pool.py
```diff
@@ -15,6 +15,11 @@ def process_row_inner(self, samples_info, pool, reception_date, comment):
                                             f"Make sure SampleSubmission sheet Pool Name column values "
                                             f"match a value in PoolSubmission sheet Pool Name column.")
         else:
+            # Validate that all libraries have the same platform
+            set_platform = set(sample['library'].platform_id for sample in samples_info)
+            if len(set_platform) > 1:  # len(set_kind) > 1 => not all same platform
+                self.errors["source_sample"] = (f"Source samples in pool {pool['name']} must be have the same platform .")
```
UlysseFG, line 21: Since the pool is already made i would say "Libraries in pool ..." and there is an extra "be" and an extra space before the ".".

UlysseFG (approved):

PR #385, opened 2022-10-03
Fixed some validations for pooling pools,

UlysseFG:


backend/fms_core/models/sample.py
```diff
@@ -70,18 +70,6 @@ def is_pool(self) -> bool:
     def is_library(self) -> bool:
         return True if any([derived_sample.library is not None for derived_sample in self.derived_samples.all()]) else False
 
-    @property
-    def sample_kind_name(self) -> str:
-        if self.is_library:
-            if self.is_pool:
-                sample_kind_name = POOL_KIND_NAME
```
sebastianballesteros, line 77: Get rid of unnecessary constant.

PR #387, opened 2022-10-03
For sample pool (top ups) allow showing individual and project on top of sample_kind.

UlysseFG:


sebastianballesteros:
This is assuming we will only allow top-ups with the same sample kind, project and individual, right?

UlysseFG:
> This is assuming we will only allow top-ups with the same sample kind, project and individual, right?

Yeah sample_kind and individual are already enforced by validation during pooling. Project ... Maybe we could automatically add the same project to all derived sample for top ups. It does not really make sense to do a top up and assign a different project to one of the derived_sample.

PR #395, opened 2022-10-06
Removed page preloading from PaginatedTable component.

UlysseFG:


frontend/src/components/PaginatedTable.js
```diff
@@ -37,31 +37,23 @@ function PaginatedTable ({
...
-  const shouldLoadNextChunk =
-    !loading && (isCurrentPageUnloaded || doesNextPageContainUnloaded);
+  const shouldLoadNextChunk = !loading && isCurrentPageUnloaded;
 
   if (shouldLoadNextChunk) {
     let offset
 
     if (isCurrentPageUnloaded)
       offset = Math.floor(startIndex / pageSize) * pageSize;
-    else if (doesNextPageContainUnloaded)
```
ckostiw, line 63: The value of `offset` should not be left undefined if `isCurrentPageUnloaded` is false.

ckostiw (changes requested):

frontend/src/components/PaginatedTable.js
```diff
@@ -37,31 +37,27 @@ function PaginatedTable ({
...
-  const doesNextPageContainUnloaded = !isLastPage && nextPageEndIndex > items.length && items.length < totalCount;
-  const shouldLoadNextChunk =
-    !loading && (isCurrentPageUnloaded || doesNextPageContainUnloaded);
+  const shouldLoadNextChunk = !loading && isCurrentPageUnloaded;
 
   if (shouldLoadNextChunk) {
     let offset
 
-    if (isCurrentPageUnloaded)
+    if (isCurrentPageUnloaded){
```
ckostiw, line 55: I think that we can just get rid of the `isCurrentPageUnloaded` test here and simply compute the offset. We want the same offset if `hasUnloadedItems`  or '`isCurrentPageUnloaded` is true. If we are on page six and `hasUnloadedItems` is true, but we ask for the items on page 0, we will never load the page 6 items and go into an infinite loop.

ckostiw (changes requested):

ckostiw (approved):

PR #397, opened 2022-10-07
Fixed library sorters

sebastianballesteros:


ckostiw (commented):
I tested this and it solved the issue of the Libraries table going into an infinite loop whenever I would sort by Index.

UlysseFG:
We'll use another solution to fix the issue. Closing this PR.

PR #399, opened 2022-10-07
Fixes lineage graph navigation

ckostiw:
The new hash url feature broke lineage graph navigation. When the user clicks a node in the graph we navigate to the clicked sample. The url generated by the graph did not include the `#lineage` key, so the url would navigate the user to the Overview tab of the sample details page, rather than the lineage graph.

Now the sample url will include the `#lineage` key and the user will remain in the lineage graph.

To avoid problems in the future, I made the navigation functions props of SampleDetailsLineage to avoid having hard-coded links in the graph itself.

I also fixed the key for the Pool tab, which was still a number rather than "pool".

I cleaned up some dead code and useless imports I found in the code I touched.

In the urlHash code, I stopped setting the default hash key automatically because it was messing something up in the history which made it impossible to navigate from a process details page back to the lineage graph (clicking the back button didn't do anything - you would be stuck on the process details page).

Removed useCallback from useHaskKey and this was causing the tab view to route to stale url's unexpectedly.

### DIFF
--- a/frontend/src/components/samples/details/SampleDetailsContent.js
+++ b/frontend/src/components/samples/details/SampleDetailsContent.js
@@ -131,6 +131,22 @@ const SampleDetailsContent = ({
   const [sampleMetadata, setSampleMetadata] = useState([])
   const [activeKey, setActiveKey] = useHashURL('overview')
 
+
+  // Navigate to a sample when the sample's node is clicked in the lineage graph.
+  const navigateToSample = (sample_id) => {
+    if (sample_id) {
+      history(`/samples/${sample_id}#lineage`)
+    }
+  }
+
+  // Navigate to a process measurement when the user clicks an edge
+  // in the lineage graph.
+  const navigateToProcess = (process_id) => {
+    if (process_id) {
+      history(`/process-measurements/${process_id}`)
+    }
+  }
+
   // TODO: This spams API requests
   if (!samplesByID[id])
     getSample(id);
@@ -329,22 +345,8 @@ const SampleDetailsContent = ({
 
         <TabPane tab={`Lineage`} key="lineage" style={tabStyle}>
           <SampleDetailsLineage sample={sample} 
-            handleSampleClick={
-              // Navigate to another sample in the lineage graph
-              (sample_id) => {
-                if (sample_id) {
-                  history(`/samples/${sample_id}#lineage`)
-                }
-              }
-            }
-            handleProcessClick={
-              // Navigate to the process measurement details page
-              (process_id) => {
-                if (process_id) {
-                  history(`/process-measurements/${process_id}`)
-                }
-              }
-            }
+            handleSampleClick={navigateToSample}
+            handleProcessClick={navigateToProcess}
           />
         </TabPane>
 
--- a/frontend/src/components/samples/details/SampleDetailsContent.js
+++ b/frontend/src/components/samples/details/SampleDetailsContent.js
@@ -328,7 +328,24 @@ const SampleDetailsContent = ({
         </TabPane>
 
         <TabPane tab={`Lineage`} key="lineage" style={tabStyle}>
-          <SampleDetailsLineage sample={sample} tabPaneKey="lineage"/>
+          <SampleDetailsLineage sample={sample} 
+            handleSampleClick={
+              // Navigate to another sample in the lineage graph
+              (sample_id) => {
+                if (sample_id) {
+                  history(`/samples/${sample_id}#lineage`)
+                }
+              }
+            }
+            handleProcessClick={
+              // Navigate to the process measurement details page
+              (process_id) => {
+                if (process_id) {
+                  history(`/process-measurements/${process_id}`)
+                }
+              }
+            }
+          />
         </TabPane>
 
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/useHashURL.js
```diff
@@ -22,12 +22,6 @@ export const useHashURL = (defaultKey) => {
     const location = useLocation()
     const currentKey = location.hash.slice(1)
     const setCurrentKey = useCallback((key) => { history(`#${key}`) }, [])
-    // If no key is currently in the url then add the default key automatically.
```
ckostiw, line 25: This was messing up the history and also generating a React warning in the console, so I got rid of it.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -328,10 +328,27 @@ const SampleDetailsContent = ({
         </TabPane>
 
         <TabPane tab={`Lineage`} key="lineage" style={tabStyle}>
-          <SampleDetailsLineage sample={sample}/>
+          <SampleDetailsLineage sample={sample} 
+            handleSampleClick={
+              // Navigate to another sample in the lineage graph
+              (sample_id) => {
```
UlysseFG, line 334: I would prefer not to have those function inline. Assign them to const to make it more readable. to reduce brackets you can use the form sample_id && history() i think.

frontend/src/components/samples/details/SampleDetailsLineage.js
```diff
@@ -202,10 +201,19 @@ const SampleDetailsLineage = ({
                     id="graph-id"
                     data={adjustedGraphData}
                     config={graphConfig}
-                    onClickNode={(id, _) => history(`/samples/${id}`)}
+                    // tabPaneKey is used to generate a hash url for the lineage tab.
+                    // Without the hash url, clicking a node navigates the user to the Overview tab
+                    // rather than staying in the graph.
+                    onClickNode={(id, _) => {
+                      if (id && handleSampleClick) {
```
UlysseFG, line 208: is handleSampleClick ever false ? maybe skip this test and remove the if using id && handleSampleClick(id)

ckostiw:
The lineage graph crashes when I navigate away from the lineage table to a sample's Overview tab, so I'll have to fix that too.

Screen Shot 2022-10-11 at 10 15 37 AM

sebastianballesteros:
mmm weird, I tried replicating the crash but didn't work.

frontend/src/components/samples/details/SampleDetailsLineage.js
UlysseFG: I am not sure we want failures to be silent (it might be partially desirable in some case but here if we forget to set a callback i would expect a failure) if we want to remove the functionality we just remove the hook altogether. At the moment, it is not optional. Since we provide both the hook and the callback i would take a missing callback as a programming error which we typically don't test for. If you want to keep the test then add an else so it actually fails when a callback is not provided (or at least a console message is present to make debugging easier).

ckostiw: It's a standard pattern in react for a component to take callback functions as props, and to make those props optional. The caller can choose to pass a function, or leave it undefined. If the component receives a callback it will call it. If not, it ignores it gracefully, without exploding. This is the expected behaviour for callback props. There's nothing in the lineage graph that requires either of these two callbacks to be defined, so they are optional. 
I don't agree with the idea of purposefully causing an error that results in a blank page for the user as a way of guarding against a programming error. I'd much rather have the user report that nothing is happening when they click a node than have the app crash.

UlysseFG: We made that decision already we don't want those optional ... we want them to link to sample and processMeasurement. The optional test is purposeful testing of programming error since the only way this could be true is if we forget to add the callbacks ... making the choice of making something optional is not driven by the component, but by us. If we start leaving all those "options" open this will cause much more harm than good. Easiest way to see it is each time you can remove an if do it ...

frontend/src/hooks/useHashURL.js
```diff
@@ -17,17 +17,12 @@ import { useCallback } from "react"
  * @param {*} defaultKey 
  * @returns 
  */
+
 export const useHashURL = (defaultKey) => {
     const history = useNavigate()
     const location = useLocation()
     const currentKey = location.hash.slice(1)
-    const setCurrentKey = useCallback((key) => { history(`#${key}`) }, [])
```
ckostiw, line 24: useCallback was causing a routing issue where clicking on a node would properly update the current url with the new sample id, but clicking on a tab in sample details would "reset" the sample id to a previous sample, and this would occasionally cause the lineage graph to crash if it couldn't find the sample in samplesById.
The problem is that I used useCallback without specifying any dependencies. I thought that `history` was just a function, but in fact it is a closure and a location is stored in the closure scope. My callback had a closure on a stale version of history that contained a stale url, and calling `history` would reset the url to a previously visited sample.
I don't think useCallback adds any real value here, so I just got rid of it.

sebastianballesteros:
It works smoothly for me compared to before.

UlysseFG (approved):

sebastianballesteros (approved):

PR #400, opened 2022-10-11
Addition of new taxon to sample submission

UlysseFG:


ckostiw (changes requested):
I checked out the branch and tried to open the template. Excel complains that the file needs repairs when I try to open it.
Screen Shot 2022-10-11 at 9 25 30 AM

ckostiw:
I checked out the original template from master and I added the new taxon id to the index. Here's the updated file. Does it work for you?

[Sample_submission_v3_12_0.xlsx](https://github.com/c3g/freezeman/files/9756504/Sample_submission_v3_12_0.xlsx)

sebastianballesteros:
I have the same problem as Chris when trying to open the new template. However, if I "recover" the file then it works. 
@ckostiw your's works fine.

UlysseFG:
There was an filter field maybe that was the cause of the issue for you guys. I removed it... If it does not work, i will keep looking ... Your file does not have the new taxon in the comment. The first line of the Taxons column refer to the named range the 499 next lines refer to the range directly. I fixed that in my file. Try again if it does not work, i will get rid of the named ranged altogether.

sebastianballesteros:
I am still having the same problem.

UlysseFG:
I don't know if it's still your excel that makes troubles for nothing on my side it works ok but i guess i have no way to know... @sebastianballesteros can you convert my file and resubmit it so that it works for you guys ? @ckostiw file had a couple things to fix so i would prefer you taking mine to make sure it is the right version ... Make sure it still works with prefiling ...

sebastianballesteros:
Done. Can you make sure it works @ckostiw  @UlysseFG? Thank you.

ckostiw (approved):

ckostiw:
The template now loads for me without any problems.

PR #401, opened 2022-10-11
Alias migration

sebastianballesteros:


backend/fms_core/migrations/0044_v3_13_0.py
```diff
@@ -0,0 +1,52 @@
...
+    Biosample = apps.get_model("fms_core", "Biosample")
+
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+
+        reversion.set_comment(f"Populate biosamples' alias with the name of the first submitted sample.")
+        reversion.set_user(admin_user)
+
+        for biosample in Biosample.objects.all():
+            first_sample = biosample.derived_samples.order_by('created_at').first().samples.order_by('created_at').first()
```
UlysseFG, line 27: Don't rely on date for that please... it should technically be true but it really is a risky approach. Use samples without parent instead, i think it is more reliable than dates. It is likely equivalent in the current state of the DB but i don't want to have to test every single entry to make sure.

backend/fms_core/services/sample.py
```diff
@@ -37,7 +37,7 @@ def create_full_sample(name, volume, collection_site, creation_date,
         biosample_data = dict(
             collection_site=collection_site,
             **(dict(individual=individual) if individual is not None else dict()),
-            **(dict(alias=alias) if alias is not None else dict()),
+            **(dict(alias=alias) if alias else dict(alias=name)),
```
UlysseFG, line 40: you should not get rid of the is not None i think. If someone submit a falsy value (e.g. numeric 0) we'll get mixups.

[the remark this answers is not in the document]
sebastianballesteros: I do not understand what you are trying to say.

UlysseFG (approved):

backend/fms_core/tests/test_template_importers/sample_submission_test.py
```diff
@@ -50,6 +50,7 @@ def test_import(self):
             sample = Sample.objects.get(name=sample_name)
             derived_sample_id = DerivedBySample.objects.filter(sample_id=sample.id).first().derived_sample_id
```
ckostiw, line 51: My linter is complaining that `derived_sample_id` is not a property of DerivedBySample. Maybe this should be `first().derived_sample.id`? I don't know how the test could be passing though, if this is a real error...

ckostiw: I traced the test in the debugger and derived_sample_id does seem to be returning id values, so the test is okay.

UlysseFG: This is one of Django magic trick. Each FK is in fact "attribute_name"_id in the DB and can be referred as such in the code. The advantage over derived_sample.id is that you do not need to join the derived_sample table to get the id, you have it on the derived_by_sample table.

ckostiw (approved):

PR #404, opened 2022-10-12
prevent crash for empty result set. Hide pool fields for sample export.

UlysseFG:


ckostiw (approved):
Tested sample export with pools and with zero samples and this fixed the 500 errors that would happen on master.

PR #405, opened 2022-10-12
Add pooling to normalization planning

UlysseFG:


ckostiw (commented):
I find it confusing that in the Sample_pooling_v3... template the 'Robot Destination Container' and 'Robot Destination Coord' columns show up in between the "Destination Container xxx" columns. I would group the 4 Destination Container columns together, and put the robot columns after them.

UlysseFG:
> I find it confusing that in the Sample_pooling_v3... template the 'Robot Destination Container' and 'Robot Destination Coord' columns show up in between the "Destination Container xxx" columns. I would group the 4 Destination Container columns together, and put the robot columns after them.

This was meant to make it easier for he user to check between the container coord and robot coord without having to skip over the name and kind columns. The robot coord are meant for the user to have a reference for the robot file. It is not super cute when you don't care what is there, but i think it will make it easier to check for each container coordinate their robot equivalent. I agree that it is ugly but i would prefer not to change it just to have to change it back later.

ckostiw:
Just checking - normalization only supports 96 and 384 well plates? I can't pool normalized samples in a tube? The template only allows plate container kinds.

ckostiw:
Normalization planning template: no drop-downs for container coords

UlysseFG:
> Just checking - normalization only supports 96 and 384 well plates? Those seem to be the only container kinds allowed for normalized samples. You can use a tube for the pooled result, but not for normalized samples.

yep the only formats accepted for normalization output are plates (verified with the lab). The pooling from the normalization output can then only take plates as input. If someone need to pool tubes they will have to use the pooling template and edit their robot file manually.

The combination of all possibility is and the amount of code for each is far from negligible we'll see down the line if this is something that is needed but it too much work for doing it just in case. We could ask that in a future meeting with the lab.

UlysseFG:
> Normalization planning template: no drop-downs for container coords

From what i recall someone from the lab mentioned they preferred not to have drop down for Coordinates. On a 386-well plate, i don't think any one in their right mind would use the drop down. Maybe we can mention it when checking with the lab.

ckostiw (changes requested):

ckostiw (approved):

PR #408, opened 2022-10-13
Add pool toggle sample list

sebastianballesteros:


frontend/src/components/samples/SamplesListContent.js
```diff
@@ -196,17 +212,43 @@ const SamplesListContent = ({
...
+
+  // Listen to the changes in toggle to get new columns and set filters accordingly
+  useEffect(() => {
+    // Get the new columns depending on the new option
+    setColumns(getTableColumns(containersByID, individualsByID, projectsByID, sampleKinds, toggleOption))
+
+    // Only apply filters if the new selected option is different from the current one
+    if (getCurrentToggleOption() !== toggleOption){
+      // Need to reset the sortBy when there's a new toggle option to avoid problems
+      setSortByWithoutList('', '')
```
UlysseFG, line 225: Maybe you should call this clearSortBy. The List would otherwise be required ... From Antd, null are used instead of empty string ... I would prefer having null instead of empty strings if it works. If  as it seem this is a clearSortBy you can remove the params (i don't think we will ever use this for something else)

frontend/src/modules/samples/actions.js
```diff
@@ -91,6 +91,13 @@ export const setSortBy = thenList((key, order) => {
     }
 });
 
+export const setSortByWithoutList = (key, order) => async dispatch => {
```
UlysseFG, line 94: Rename and remove params,

frontend/src/components/samples/SamplesListContent.js
```diff
@@ -184,8 +185,23 @@ const SamplesListContent = ({
   setFilterOption,
   clearFilters,
   setSortBy,
+  clearSortBy,
 }) => {
 
+  const isPooledFilterKey = SAMPLE_FILTERS.is_pooled.key
+
+  function getCurrentToggleOption(){
+    if (filters[isPooledFilterKey]?.value === "true") return TOGGLE_OPTIONS.POOLS
```
UlysseFG, line 194: switch case here too

UlysseFG (approved):

ckostiw (changes requested):
Just minor stuff.

ckostiw (approved):

PR #411, opened 2022-10-19
Fix to project display issue.

UlysseFG:


sebastianballesteros (approved):

ckostiw:
I tested this PR, and I still get missing projects. I checked the api response and the project field for several of my pools are null. It looks like, at least in my case, the problem is in the backend.

Screen Shot 2022-10-21 at 8 42 38 AM

ckostiw:
Ah, I just noticed that the project is displayed for pooled samples but not for pooled libraries- that's the difference. I guess this is the expected behaviour.

PR #413, opened 2022-10-21
Fix attempt #1

UlysseFG:


ckostiw (approved):
The template is fixed and sample submission works properly.

PR #416, opened 2022-11-01
Sample name parent column

UlysseFG:


backend/fms_core/viewsets/sample_pooled.py
```diff
@@ -33,10 +32,20 @@ class PooledSamplesViewSet(viewsets.ModelViewSet):
         output_field=BooleanField()
     )).distinct()
     queryset = queryset.filter(is_pooled=True)
-    
+
+    queryset = queryset.annotate(
```
ckostiw, line 36: I'm a bit confused because PooledSampleSerializer already contains a getter to get the parent sample name. It also has a getter to get the parent sample id. I don't understand why this annotate is necessary. I'm also concerned that we could end up matching the wrong parent sample name to the parent sample id from the serializer.

ckostiw (approved):

PR #423, opened 2022-11-08
Added platform validation

sebastianballesteros:


UlysseFG:
This looks good but we should add it to the normalization planning too to prevent someone of planning a bad pool.

UlysseFG (approved):

PR #425, opened 2022-11-08
Addition of serial_id to the instrument model.

UlysseFG:


sebastianballesteros:
You will have to modify the tests to include a serial ID when creating test instruments.

sebastianballesteros:
You will have to modify the tests to include a serial ID when creating test instruments.

PR #426, opened 2022-11-08
Experiment Run Illumina

sebastianballesteros:


UlysseFG:
Other than the migration conflict to fix ... 
the template does not keep the dropdowns when prefilled ... this is the usual issue. Maybe we want to merge it then fix it in order to streamline the PR merging process.
Also not directly related to the PR, the sample display in containers shows sample kind (e.g. (DNA)) but for pools its shows (). It could be fixed in another PR too.

UlysseFG (approved):

sebastianballesteros:
Really? I purposely exported it from Google Excel sheets so dropdown would work when prefilling, I will check it again :(

PR #429, opened 2022-11-09
Capture protocol

UlysseFG:


UlysseFG:
Reverted to draft. I realized that i need to add the capture and ChIP-Seq libraries to the sample submission template.

backend/fms_core/services/library.py
```diff
@@ -58,16 +78,86 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+                                                                            volume_destination=volume_destination,
+                                                                            execution_date=execution_date,
+                                                                            comment=comment)
+    errors.extend(errors_inherit)
+    warnings.extend(warnings_inherit)
+
+    return sample_destination, errors, warnings
+
+
+def capture_library(process, library_selection, sample_source, container_destination, coordinates_destination, volume_used,
```
sebastianballesteros, line 115: The python doc comments for the services are missing.

backend/fms_core/services/library.py
```diff
@@ -58,16 +78,86 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+                                                                            execution_date=execution_date,
+                                                                            comment=comment)
+
+    errors.extend(errors_inherit)
+    warnings.extend(warnings_inherit)
+
+    return sample_destination, errors, warnings
+
+
+def _inherit_library(process, new_library_info, sample_source, container_destination, coordinates_destination, volume_used,
```
sebastianballesteros, line 153: Same for this service.

backend/fms_core/template_importer/row_handlers/library_capture/library_capture.py
```diff
@@ -0,0 +1,89 @@
...
+        if not capture_batch_info:
+            self.errors['library_capture'] = 'No batch is associated with this capture.'
+
+        # Calling the service creator for Samples in LibraryPreparation
+        source_sample_obj, self.errors['container'], self.warnings['container'] = \
+            get_sample_from_container(barcode=source_sample['barcode'], coordinates=source_sample['coordinates'])
+
+        if not volume_used:
+            self.errors['volume_used'] = f"Volume used must be entered"
+        elif source_sample_obj and volume_used > source_sample_obj.volume:
```
sebastianballesteros, line 26: Accessing `source_sample.obj.volume` before validating if the source sample is not None will result in problems.

backend/fms_core/template_importer/row_handlers/library_capture/library_capture.py
```diff
@@ -0,0 +1,89 @@
...
+            self.errors['volume_used'] = f"Volume used ({volume_used}) exceeds the current volume of the sample ({source_sample_obj.volume})"
+
+        if source_sample_obj:
+            # Check if sample is not a library or a pool of libraries
+            if not source_sample_obj.is_library:
```
sebastianballesteros, line 31: We may want to add the failed qc flag warning to capture too.

sebastianballesteros:
Another small comment,  the icon in the action dropdown for Capture is not really representative (it is the default download one).

backend/fms_core/services/library.py
```diff
@@ -58,16 +78,86 @@ def create_library(library_type, index, platform, strandedness, library_size=Non
...
+    if not library_selection:
+        errors.append(f"Capture type is required.")
+
+    if not sample_source:
+        errors.append(f"Source sample is required.")
+    else:
+        for derived_sample_source in sample_source.derived_samples.all():
+            library_source_obj = derived_sample_source.library
+              
+            if library_source_obj.library_selection is not None:
```
sebastianballesteros, line 130: This will crash because even if you check that the source sample is a library you go on and call this service with the source sample object.

backend/fms_core/template_importer/importers/library_capture.py
```diff
@@ -0,0 +1,102 @@
...
+            capture_batch_info = dict(
+                # Library attributes data dictionary and related objects
+                capture_date=input_to_date_and_none(capture_batch_dict['Capture Date (YYYY-MM-DD)']),
+                # library selection
+                **batch_objects,
+                # Additional data for this row
+                protocol=self.preloaded_data['protocol'],
+            )
+
+            capture_batch_rows_data[capture_batch_dict['Capture Batch ID']] = capture_batch_info
```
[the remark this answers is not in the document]
UlysseFG: I'll add a check to make sure the missing Capture Batch ID does not crash the processing. The row handler will return the missing ID error.

backend/fms_core/template_importer/row_handlers/library_capture/library_capture.py
UlysseFG: isn't that exactly what is being done here ? if source_sample_obj is None python will never evaluate source_sample_obj.volume.

sebastianballesteros: Yeap, sorry. misread it

backend/fms_core/template_importer/row_handlers/library_capture/library_capture.py
```diff
@@ -27,9 +27,14 @@ def process_row_inner(self, capture_batch_info, source_sample, volume_used,
...
+            if any([source_sample_obj.quality_flag is False, source_sample_obj.quantity_flag is False]):
+                self.warnings["qc_flags"] = (f"Source library {source_sample_obj.name} has failed QC.")
+
             # Check if sample is not a library or a pool of libraries
             if not source_sample_obj.is_library:
                 self.errors['source_sample'] = f"Source sample {source_sample_obj.name} must be a library to be captured."
+
             # Make sure no Capture were previously done on the library
             error = []
             for derived_sample in source_sample_obj.derived_samples.all():
```
sebastianballesteros, line 40: Same problem here. If sample is not a library next line will check `derived_sample.None.library_selection` and will crash.

sebastianballesteros:
Test seems to fail on my side too.

backend/fms_core/template_importer/row_handlers/library_capture/library_capture.py
UlysseFG: There won't be non-library coming in here. The issue i have is that we can test every single step of the way if we have a library. With a layered architecture like ours it adds overhead and lines of codes but do not really provide much value. The service layers should have all user input validated beforehand if they can be. even if i add a check here it will not get back to the user ... In the end it's up to us not to call this service with something that is not a library. Its not a valid option to let this not crash by returning an error. Its safer if it crashes. Maybe we should start using some assert statements.

sebastianballesteros (approved):

PR #431, opened 2022-11-10
Fixed library counts in library summary

sebastianballesteros:


frontend/src/components/DashboardPage.js
```diff
@@ -142,8 +142,9 @@ const DashboardPage = ({
               </Col>
               <Col {...STATS_COL_PROPS}>
                 {((librariesSummary.library_type_counts && Object.keys(librariesSummary.library_type_counts)) || []).map((library_type) =>
-                  <Statistic title={libraryTypesByID[library_type]?.name} value={librariesSummary.library_type_counts[library_type] || "—"} />
+                  <Statistic title={libraryTypesByID[library_type]?.name || "-"} value={librariesSummary.library_type_counts[library_type] || "—"} />
```
UlysseFG, line 145: Why the additional dashes ? If you have libraries you should have library_types ... this is meant while it is loading ?

sebastianballesteros: Yes

backend/fms_core/viewsets/library.py
```diff
@@ -183,12 +184,12 @@ def summary(self, _request):
...
+        # Creates a dictionary of libraries of the form {library_id : library_type_id} which removes duplicates (pools)
+        libraries = self.queryset.values('id', 'is_pooled', 'derived_samples__library__library_type')
+        libraries_type = {l['id']: l['derived_samples__library__library_type'] for l in libraries if not l['is_pooled']}
         return Response({
-            "total_count": self.queryset.all().count(),
-            "library_type_counts": {
-                 c["derived_samples__library__library_type"]: c["derived_samples__library__library_type__count"]
-                 for c in self.queryset.values("derived_samples__library__library_type").annotate(Count("derived_samples__library__library_type"))
-            },
+            "total_count": len(libraries),
```
UlysseFG, line 191: The total count doesn't work as far as i can see ...  the pooled libraries with more than a single type appear multiple time.

sebastianballesteros: Libraries_type doesn't count the libraries that are pools.

sebastianballesteros: The new changes should remove the duplicates.

sebastianballesteros: We need to think about all the cases cause I think this need to be tested thoroughly

UlysseFG: Eheh i think building a dictionnary with all libraries will likely become very slow with time (with libraries). You are making it more complicated that it needs to be. Maybe try ".exclude(is_pooled=True)" to the library_type_count and ".filter(is_pooled=True).values("id") for the pooled libraries. You get the the total by doing the sum... or you could get the total by removing the filter. The distinct is likely already included in the endpoint queryset already.

PR #433, opened 2022-11-14
Experiment run processing launch

ckostiw:
UX support for launching run processing, with a stub service in the backend where the launch will be implemented.


backend/fms_core/services/experiment_run.py
```diff
@@ -86,3 +87,14 @@ def create_experiment_run(experiment_run_name,
       experiment_run = None
 
     return (experiment_run, errors, warnings)
+
+def launch_experiment_run(pk):
+    errors = []
+    warnings = []
+
+    # For now, just set the launch timestamp on the experiment run...
+    experiment_run = ExperimentRun.objects.get(id=pk)
```
sebastianballesteros, line 96: Might be worth to wrap this in a Try/Except to catch DoesNotExist exceptions.

backend/fms_core/viewsets/experiment_run.py
```diff
@@ -41,4 +42,22 @@ def get_renderer_context(self):
     @action(detail=False, methods=["get"])
     def list_export(self, _request):
         serializer = self.serializer_export_class(self.filter_queryset(self.get_queryset()), many=True)
-        return Response(serializer.data)
\ No newline at end of file
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["get"])   # TODO should this be post? update?
```
sebastianballesteros, line 47: In any case it would be patch (because we are partially updating an entry). I think we mostly use PATCH throughout FreezeMan. Refer to `set_release_status` in the dataset viewset for an example.

frontend/src/utils/api.js
```diff
@@ -58,6 +58,7 @@ const api = {
       check:  (action, template) => post(`/experiment-runs/template_check/`, form({ action, template })),
       submit: (action, template) => post(`/experiment-runs/template_submit/`, form({ action, template })),
     },
+    launchRunProcessing: experimentRunId => get(`/experiment-runs/${experimentRunId}/launch_run_processing/`)
```
sebastianballesteros, line 61: If you decided to use PATCH in the backend we need patch() here.

frontend/src/components/experimentRuns/ExperimentRunsListContent.js
```diff
@@ -80,14 +73,24 @@ const getTableColumns = (containersByID, runTypes, instruments) => [
...
     width: 180,
   },
-
+  {
+    title: "Launch",
+    dataIndex: "run_processing_launch_date",
+    sorter: true,
+    render: (_, experimentRun) => (
+      <div style={{minWidth: "12rem"}}>
+        <ExperimentRunLaunchCard experimentRun={experimentRun} experimentRunLaunch={launchesById[experimentRun.id]} launchExperimentRun={launchExperimentRun}/>
```
sebastianballesteros, line 82: isn't the parameter `launchExperimentRun` superfluous? I might be wrong though.

frontend/src/components/experimentRuns/ExperimentRunLaunchCard.js
```diff
@@ -0,0 +1,119 @@
...
+        )
+      }
+    }
+
+    function getContents() {
+      if (panelIsOpen) {
+        return getLaunchPanelContents()
+      } else {
+        const launchDate = experimentRun.run_processing_launch_date ?
+          moment(experimentRun.run_processing_launch_date).format("YYYY-MM-DD LT")
```
sebastianballesteros, line 97: Might be my setup but my timezone is not set up correctly. moment().tz().format()?

frontend/src/components/experimentRuns/ExperimentRunsListContent.js
```diff
@@ -80,14 +73,24 @@ const getTableColumns = (containersByID, runTypes, instruments) => [
     sorter: true,
     width: 180,
   },
-
+  {
+    title: "Launch",
+    dataIndex: "run_processing_launch_date",
+    sorter: true,
```
sebastianballesteros, line 79: Sorter doesn't work (maybe it does but I have to launches, each with same dates but different time and it does not reorder them)

frontend/src/components/experimentRuns/ExperimentRunLaunchCard.js
ckostiw: I think that's because I was setting the timestamp with a naive date, using datetime.now(). I changed that in the generate-event-file to use timezone from django_utils. The problem should be fixed in that branch.

frontend/src/components/experimentRuns/ExperimentRunsListContent.js
ckostiw: Fixed - I was missing the filter key for run_processing_launch_date

frontend/src/components/experimentRuns/ExperimentRunsListContent.js
ckostiw: Yup. Originally I was passing the function as a prop, but now I just import the function in the card component.

sebastianballesteros (approved):

PR #435, opened 2022-11-15
Added initial concentration column and formula to template

sebastianballesteros:


backend/fms_core/templates.py
```diff
@@ -174,11 +174,11 @@
 }
 
 NORMALIZATION_TEMPLATE = {
-  "identity": {"description": "Template to perform normalization", "file": static("submission_templates/Normalization_v3_10_0.xlsx")},
+  "identity": {"description": "Template to perform normalization", "file": static("submission_templates/Normalization_v3_14_0.xlsx")},
   "sheets info": [
       {
         'name': 'Normalization',
-        'headers': ['Sample Name', 'Source Container Barcode', 'Source Container Coord', 'Robot Source Container', 'Robot Source Coord',
+        'headers': ['Sample Name', 'Source Container Barcode', 'Source Container Coord', 'Sample Initial Concentration', 'Robot Source Container', 'Robot Source Coord',
```
UlysseFG, line 181: Placement of the field is not great in between the source container and its robot equivalent. Rename (Initial Conc. (ng/uL)) and move the field after Source Depleted. Do not forget to update "Conc. (ng/uL)" formula.

UlysseFG:
In the normalization template, the comment on the column Conc. (ng/uL) contains an id and a mention of Microsoft Office ... those are not useful.

PR #439, opened 2022-11-16
Moved concentration validation before calculations

sebastianballesteros:


backend/fms_core/template_importer/row_handlers/normalization_planning/normalization_planning.py
```diff
@@ -42,10 +42,10 @@ def process_row_inner(self, source_sample, destination_sample, measurements, rob
             barcode=source_sample['container']['barcode'],
             coordinates=source_sample['coordinates'])
 
-        if source_sample_obj is not None and "concentration" not in self.errors.keys():
-            if source_sample_obj.concentration is None:
-                self.errors['concentration'] = f'A sample or library needs a known concentration to be normalized. QC sample {source_sample_obj.name} first.'
+        if source_sample_obj.concentration is None:
+            self.errors['concentration'] = f'A sample or library needs a known concentration to be normalized. QC sample {source_sample_obj.name} first.'
 
+        if source_sample_obj is not None and "concentration" not in self.errors.keys():
```
UlysseFG, line 48: maybe use self.has_errors() instead. Less likely that some new error will get past unnoticed.

PR #442, opened 2022-11-18
Generate event file

ckostiw:
A service that generates an event file for run processing.


backend/fms_core/services/event_file.py
```diff
@@ -0,0 +1,233 @@
...
+    fms_run_start_date: Union[str, None]
+
+    # Flowcell / container for experiment
+    fms_container_id: int
+    fms_container_barcode: str
+
+    instrument_id: str
+    fms_instrument_type: str
+
+    samples: List[EventSample]
```
UlysseFG, line 75: You'll have to discuss any change in format with TechDev ... I would recommend something as close to the flat original TSV as possible to make the changes on their side as small as possible for day 1. This is still a very small change (and for the better) but you should still validate with them.

backend/fms_core/services/event_file.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+EVENT_FILE_VERSION = "1.0.0"
```
UlysseFG, line 78: please get rid of any event file references. Let's give it a more useful name. Run info file. fms run export file.

backend/fms_core/services/event_file.py
```diff
@@ -0,0 +1,233 @@
...
+    
+
+@dataclass
+class EventManifest:
+    # Manifest version (1.0.0 to start) 
+    version: str
+
+    # Experiment run name and id
+    run_name: str
+    fms_run_id: int
```
UlysseFG, line 65: you were right about the too many "fms_".

backend/fms_core/services/experiment_run.py
```diff
@@ -89,12 +92,31 @@ def create_experiment_run(experiment_run_name,
     return (experiment_run, errors, warnings)
 
 def launch_experiment_run(pk):
```
UlysseFG, line 94: We should rename to something closer to what it actually does. generate_run_info_file(experimentrun_id). I would recommend this service also return the file instead of the experiment_run in order to send it back upon request.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,267 @@
...
+        row.project_obj_id = project.id
+        row.project_name = project.name
+        row.hercules_project_id = project.external_id
+        row.hercules_project_name = project.external_name
+
+    row.container_coordinates = sample.coordinates
+
+    # Convert coordinates from A01 format to lane number.
+    row.lane = {
+        'A01': 1,
```
UlysseFG, line 187: ugh... you can do better ;) what if we get a 3000 coordinates flowcell ? ; )

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,267 @@
...
+    Protocol,
+    Sample, 
+    SampleLineage,
+)
+
+Obj_Id = Union[int, None]
+
+@dataclass
+class RunInfoSample:
+    sample_name: Union[str, None]
```
UlysseFG, line 26: The direct pooled sample name is not very informative for techdev. I would store this specific as "pool_name" which is true and leave us space to use sample_name to store the alias.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,353 @@
...
+        # Find the library preparation process measurement
+        protocol: Protocol = Protocol.objects.get(name="Library Preparation")
+        lineage: SampleLineage = SampleLineage.objects.get(
+            process_measurement__process__protocol_id=protocol.pk,
+            child__derived_samples__library__id=library_to_find.pk)
+        library_prep = lineage.process_measurement
+    except SampleLineage.DoesNotExist:
+        pass
+    except SampleLineage.MultipleObjectsReturned:
+        pass
```
UlysseFG, line 301: please don't pass on exception handling ... even if it is a bit of a performance hit test with .exist() first then if it does do the get(). This is one more hit on the DB but it is much less risky. i also doubt it is a normal occurrence that we get a MultipleObjectReturned... we would not get this to pass by silently (although i think it would crash anyway but then why try to catch it). Given the peculiar get() in django it could make sense for performance to use the except pass, but not much for the multipleobjectreturned.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,353 @@
...
+            except:
+                pass
+
+        # Find the library preparation process measurement
+        protocol: Protocol = Protocol.objects.get(name="Library Preparation")
+        lineage: SampleLineage = SampleLineage.objects.get(
+            process_measurement__process__protocol_id=protocol.pk,
+            child__derived_samples__library__id=library_to_find.pk)
+        library_prep = lineage.process_measurement
+    except SampleLineage.DoesNotExist:
```
UlysseFG, line 298: do not shadow other "try...except" unless it is needed. Here the only risky calls are the lineage get() which should be dealt with individually and separately (at the same level, not one within the other.)

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,353 @@
...
+    except SampleLineage.MultipleObjectsReturned:
+        pass
+
+    return library_prep
+
+
+def _find_library_capture_process(derived_sample: DerivedSample) -> Union[Process, None]:
+    '''If this is a captured library, find the process measurement from the capture step.'''
+    if derived_sample.library is None:
+        return None
```
UlysseFG, line 309: I think we need to give a thought about the error handling structure here and other sub-functions. For exemple at the moment it is not an option for us to deliver a run info file for a sample without a library, but it is possible for a library not to have a capture (library_selection). Both will see a return value of None. At this level it is not possible to make a decision about what to do about these value, so i would recommend this function just make sure there is or not a capture and pass to the parent the task of making sure if this make sense that there is a library. Some future run (Axiom or other type of processing) are likely to involve samples without library we may want to reuse these functions between different platforms. The capture is not quite reusable without library but the idea is to have a error return value that can be interpreted as either an unexpected error OR an empty value not both.

backend/fms_core/coordinates.py
```diff
@@ -58,6 +58,70 @@ def ints(end: int, pad_to: int = 0) -> CoordinateAxis:
...
+    # Spec must be for "A01" style coordinates.
+    if not is_alpha_digit_spec(spec):
+        raise CoordinateError(f'Cannot convert coord {coord} to ordinal - CoordinateSpec does not support coordinate style.')
+    spec_letters = spec[0]
+    spec_digits = spec[1]
+    
+    # Collect the characters at the start of the coord.
+    letters = ''
+    digits = ''
+    for index in range(0, len(coord)):
```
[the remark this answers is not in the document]
ckostiw: Yes, it's overly complex, but I wanted to build in support for coordinates like "AB001" in case we need to support flowcells with thousands of coordinates.

backend/fms_core/services/experiment_run_info.py
ckostiw: Yeah, I wasn't sure what to do with MultipleObjectsReturned - I'll raise an exception instead.
I need to let DoesNotExist pass, though, because not all libraries have a library prep step. For one of my experiments I submitted libraries in a pool, so they did not go through Prep or QC, and thus there is no Library Preparation process for them (and thus, no Library Kit value for the run info). I'll refactor the function to make this clearer.

backend/README.md
```diff
@@ -44,6 +44,10 @@ Python package dependencies are listed in `requirements.txt`.
      export PG_HOST=127.0.0.1
      export PG_PORT=5432
      ```
+
+    Set a `FMS_RUN_INFO_PATH` environment variable to point to a directory where
```
UlysseFG, line 48: Adjust numbering to allow this entry to have a number too.

backend/README.md
```diff
@@ -44,6 +44,10 @@ Python package dependencies are listed in `requirements.txt`.
      export PG_HOST=127.0.0.1
      export PG_PORT=5432
      ```
+
+    Set a `FMS_RUN_INFO_PATH` environment variable to point to a directory where 
+    Freezeman will output run info files to trigger run processing (defaults to
+    `lims-run-info` in the project folder).
```
UlysseFG, line 50: Add an example of the command to add the variable to the environment.

backend/example.uwsgi.ini
```diff
@@ -25,6 +25,9 @@ env = FMS_EMAIL_USER=romain.gregoire@mcgill.ca
 env = FMS_EMAIL_PASSWORD=secret
 env = FMS_EMAIL_TLS=False
 
+# Run processing spool directory for run info files.
+env = FMS_RUN_INFO_PATH=/data/freezeman-lims-run-info
```
UlysseFG, line 29: I would use the default path name here.

backend/fms_core/services/experiment_run.py
```diff
@@ -88,13 +91,68 @@ def create_experiment_run(experiment_run_name,
...
+            generate_run_info_file(experiment_run)
+
+            experiment_run.run_processing_launch_date = timezone.now()
+            experiment_run.save()
+        except Exception as e:
+            errors.append(str(e))
+               
     return (experiment_run, errors, warnings)
+
+def get_run_info_for_experiment(pk):
```
UlysseFG, line 128: This service hardly does anything ... it is a near empty wrapper to generate_run_info. I would get rid of it.

backend/fms_core/utils.py
```diff
@@ -163,3 +165,15 @@ def make_generator(obj: Union[Any, None, Iterable[Any]]) -> Generator[Any, None,
...
         except TypeError:
             yield obj
+
+def make_timestamped_filename(file_name: str, extension: str) -> str:
+    """
+    Creates a file name composed of a base file name followed by a timestamp, followed
+    by an extension, eg "MyFile_2022-11-25_08-13-45.json".
+
+    The file name and timestamp are separated by an underscore.
+    The extension parameter should not include a '.', as one is inserted automatically.
```
UlysseFG, line 175: You might be better to align with os.path.splitext which return the "." as part of the extension. If we ever need to add a timestamp to an existing filename + extension this will be easier.

backend/fms_core/viewsets/experiment_run.py
```diff
@@ -46,7 +47,21 @@ def list_export(self, _request):
...
+            The experiment ID.
+
+        Returns:
+            On success:
+            {'ok': True}
+
+            On error:
+            {'ok': False, message: <error message>}
+        '''
+        experiment_run, errors, warnings = start_experiment_run_processing(pk)
```
UlysseFG, line 64: The 2 services created in experiment_run are not really part of experiment_run. I would move the code from those 2 services to the viewsets and get rid of those 2 experiment_run services. They are just a redirection of call to experiment_run_info. As a more general comment, a function need to accomplish a function. There is too many layers of paper wrapper here.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+RUN_INFO_FILE_VERSION = "1.0.0"
+
+def generate_run_info_file(experiment_run: ExperimentRun):
+    '''
+    Generate the run info for an experiment, to be submitted to tech dev for run processing.
+
+    This function writes the data to the file it is given. The file must be text-based, open,
+    and writable.
+
+    The output is a RunInfo object in the JSON format.
```
UlysseFG, line 105: There is no return value for this function. If i want the file i must got fetch the file at a pre-established place ... this is likely more risk and more work. There is also no error return for a function that has IO which is not very safe. I would encourage to rework the signature. Maybe take the file location as additional input and return error messages on failure. Another way could be to return the file object to be saved to disk or sent as a response (this could be done in addition to saving it to disk too) failure would return None. Either way it would be more safe and clear.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+    samples: List[RunInfoSample]
+
+
+RUN_INFO_FILE_VERSION = "1.0.0"
+
+def generate_run_info_file(experiment_run: ExperimentRun):
+    '''
+    Generate the run info for an experiment, to be submitted to tech dev for run processing.
+
+    This function writes the data to the file it is given. The file must be text-based, open,
```
UlysseFG, line 102: Try to follow the Args Return format...

backend/fms_core/coordinates.py
```diff
@@ -58,6 +58,70 @@ def ints(end: int, pad_to: int = 0) -> CoordinateAxis:
 
     return tuple(str(i).zfill(pad_to) for i in range(1, end + 1))
 
+def is_alpha_digit_spec(spec: CoordinateSpec) -> bool:
+    '''Determines if a CoordinateSpec is for the alpha/digit style, eg "A01, B12, etc..."'''
```
UlysseFG, line 62: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+
+    path = os.path.join(settings.RUN_INFO_OUTPUT_PATH, file_name)
+
+    with open(path, "w", encoding="utf-8") as file:
+        json.dump(run_info, file, indent=4)
+
+def generate_run_info(experiment_run: ExperimentRun) -> RunInfo:
+    ''' 
+    Generates the run info for the experiment, including all of the derived
+    samples in the experiment.
```
UlysseFG, line 120: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+        samples=[]
+    )
+
+    run_info.samples = _generate_run_info_samples(experiment_run)
+
+    return asdict(run_info)
+
+
+def _generate_run_info_samples(experiment_run: ExperimentRun) -> List[RunInfoSample]:
+    ''' Generates the run info for every derived sample in the experiment. '''
```
UlysseFG, line 149: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+        if (sample.is_pool):
+            generated_rows += _generate_pooled_samples(experiment_run, sample)
+        else:
+            run_info_sample = _generate_sample(experiment_run, sample, sample.derived_sample_not_pool)
+            generated_rows += [run_info_sample]
+
+    return generated_rows
+
+def _generate_pooled_samples(experiment_run: ExperimentRun, pool: Sample) -> List[RunInfoSample]:
+    ''' Generates the run info for all of the derived samples in a pool. '''
```
UlysseFG, line 166: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+        derived_by_sample: DerivedBySample = DerivedBySample.objects.get(derived_sample=derived_sample, sample=pool)
+        run_info_sample.pool_volume_ratio = float(derived_by_sample.volume_ratio)
+
+        run_info_samples.append(run_info_sample)
+
+    return run_info_samples    
+
+def _generate_sample(experiment_run: ExperimentRun, sample: Sample, derived_sample: DerivedSample) -> RunInfoSample:
+    '''
+    Generates the data for one derived sample in the experiment.
```
UlysseFG, line 183: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+                capture_details = _get_capture_details(library)
+                row.capture_kit = capture_details['capture_kit']
+                row.capture_baits = capture_details['capture_baits']
+            elif library.library_selection.name == "ChIP-Seq":
+                row.chip_seq_mark = library.library_selection.target
+
+    return row
+
+def _convert_flowcell_coordinate_to_lane_number(coord: str, container: Container) -> int:
+    '''Converts a freezeman coordinate (eg. A01) to an integer flowcell lane number.'''
```
UlysseFG, line 260: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+    if container_spec is None:
+        raise Exception(f'Cannot convert coord {coord} to lane number. No ContainerSpec found for container kind "{container.kind}".')
+
+    lane = convert_alpha_digit_coord_to_ordinal(coord, container_spec.coordinate_spec)
+    return lane
+
+def _find_library_prep(library: Library) -> Union[ProcessMeasurement, None]:
+    ''' 
+    Given a library, find the Library Preparation process that was used to create the
+    library, if any, to extract the library preparation properties (eg. Library Kit).
```
UlysseFG, line 272: missing args and return.

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+    except SampleLineage.DoesNotExist:
+        # A Library Preparation step does not necessarily exist. Samples submitted as libraries
+        # will not have gone through the prep step in freezeman.
+        pass
+
+    return library_prep
+
+
+def _find_library_capture_process(captured_library: Library) -> Union[Process, None]:
+    '''Given a captured library, find the Library Capture process that produced the library.'''
```
UlysseFG, line 317: missing args and return.

backend/fms_core/utils.py
```diff
@@ -163,3 +165,15 @@ def make_generator(obj: Union[Any, None, Iterable[Any]]) -> Generator[Any, None,
                 yield x
         except TypeError:
             yield obj
+
+def make_timestamped_filename(file_name: str, extension: str) -> str:
+    """
+    Creates a file name composed of a base file name followed by a timestamp, followed
+    by an extension, eg "MyFile_2022-11-25_08-13-45.json".
```
UlysseFG, line 172: missing args and return.

UlysseFG:
We also need some rudimentary tests for the services.

backend/fms_core/services/experiment_run.py
ckostiw: Personally, I think it's cleaner if the viewset only uses services from the `experiment_run` services file, rather than calling functions directly in `experiment_run_info`. This function also does exception handling to provide a function that returns errors and warnings, to conform with all of the other service calls. I think it's worth keeping this "service" function, even if it doesn't do a lot.

backend/fms_core/utils.py
ckostiw: Ah yes. I had forgotten about `splitext`.

backend/fms_core/services/experiment_run.py
UlysseFG: It does have some structural and functional potential. It is however the single call of another placeholder function. The real function is in the experiment_run_info services. if the placement of code is the only reason to create a function you can move experiment_run_info services into experiment_run services we'll save a few dozen lines of code ;). The experiment_run_info is not really a service. It is because we don't have a better place for it but it does not wrap any model (best approximation is experiment_run). You can leave it if you really want but it really lessen code readability. you have 4 functions with nearly the same name that do pretty much the same thing (which is next to nothing). All this error checking can be moved along. I could see a functional use for start_experiment_run_processing but this one is pretty much just a redirection.

backend/fms_core/viewsets/experiment_run.py
ckostiw: I can change that, but I don't really agree. To me, the responsibility of a viewset should be to provide an end point and handle the request parameters and the response. I prefer a cleaner separation of concerns - http mechanics in the viewset, business logic in the service. Otherwise you can end up with business logic locked up in a viewset, where the code cannot be reused, or which gets copy-pasted later.

backend/fms_core/services/experiment_run.py
```diff
@@ -88,13 +91,68 @@ def create_experiment_run(experiment_run_name,
...
+    except ExperimentRun.DoesNotExist:
+        errors.append(f'Experiment run with id {pk} not found.')
+
+    if experiment_run is not None:
+        try:
+            run_info = generate_run_info(experiment_run)
+        except Exception as e:
+            errors.append(str(e))
+    
+    return (run_info, errors, [])
```
UlysseFG, line 157: use warnings instead of [] we'll end up setting up a warning that never get returned ...

backend/fms_core/services/experiment_run_info.py
```diff
@@ -0,0 +1,347 @@
...
+        container_barcode=experiment_run.container.barcode,
+        platform=instrument.type.platform.name,
+        instrument_serial_number=experiment_run.instrument.serial_id,
+        instrument_type=instrument.type.type,
+        samples=[]
+    )
+
+    run_info.samples = _generate_run_info_samples(experiment_run)
+
+    return asdict(run_info)
```
UlysseFG, line 145: We should return errors and warnings and propagate them between sub-functions. We expect services to provide information on failures. Here if requesting to start the run processing we expect to have some form of error message if there is failure (the user can forward the message to us in that case)

backend/fms_core/services/experiment_run_info.py
ckostiw: Yeah, I wasn't too sure about this function either. I trashed it and moved all the code into the service, where I think it makes more sense.

backend/fms_core/services/experiment_run_info.py
ckostiw: My plan was to have the service catch any exceptions that occur and populate the errors[] with the exception strings.

UlysseFG: Yeah maybe we should move this code out of services... we can let it here for now, but it is exactly the kind of code i talked about during the summer (neither a service nor anything else...)

backend/fms_core/viewsets/experiment_run.py
UlysseFG: you are right but that would mean we have already a lot of refactoring to do to reach that point for the other viewsets. The idea is not that i want that code in viewset or any specific place it's more that it takes 3 files changes to get to any actual code ... The distinction between viewset and the services i agree with. But i don't see why you have all those cascading function where you basically have a single functionality. The way i see it you treat experiment_run_info as an object class without making one of it. Maybe you can define it more officially and we can make a home for it (outside of services, maybe a new reporting package). That way you could use the constructor to build the run_info then you can have .tojson, .tofile, .torunprocessingspool methods to call in the service.

ckostiw:
I added a templates folder for services tests, instead of adding my stuff to existing templates, because it was so complicated just getting my single set of templates set up properly.

sebastianballesteros (approved):

PR #444, opened 2022-11-25
Metadata search

sebastianballesteros:


backend/fms_core/filters.py
```diff
@@ -75,6 +76,16 @@ def is_pooled_filter(self, queryset, name, values):
         bool_value = (values == 'true')
         return queryset.filter(is_pooled=bool_value)
 
+    def metadata_filter(self, queryset, name, values):
+        condition = Q()
+        # Metadata is of form: 'name1__value1, name2__value2, name2__Value3'
+        for metadatum in values.split(','):
+            if metadatum:
+                name_value = metadatum.split('__')
```
UlysseFG, line 84: There may be more than an instance of "__" since the value part is not controlled. We should add a validation to the metadata name attribute in order to prevent "__" to appear in the name. That way you can split on the first occurrence of "__" in the querystring value (not on all or you'll get an array greater than 2 possibly).

backend/fms_core/viewsets/sample_metadata.py
```diff
@@ -15,4 +17,14 @@ class SampleMetadataViewSet(viewsets.ModelViewSet):
...
     filterset_fields = {
         **_sample_metadata_filterset_fields
-    }
\ No newline at end of file
+    }
+
+    @action(detail=False, methods=["get"])
+    def search(self, _request):
+        """
+        Searches for metadata that match the given query
```
UlysseFG, line 25: might be good to specify they match the name.

UlysseFG:
Very nice... One thing that is a bit "irregular" is that when you apply a metadata search then go to the sample (to check on the values) come back to the filtered list you still have all the filters listed but the list of metadata search is not set anymore. That part is not a big surprise. But when i apply another metadata_filter at that point without clearing the filters, the metadata_filter is replaced by the new entries instead (my expectation) of being added to the existing ones (the other filters do not get replaced).

backend/fms_core/filters.py
```diff
@@ -75,6 +76,23 @@ def is_pooled_filter(self, queryset, name, values):
...
+                # Check that there's at least 1 '__' that separates name and value
+                if name_value and len(name_value) > 1:
+                    # Meaning value was left empty
+                    if not name_value[1]:
+                        condition |= Q(name=name_value[0])
+                    # Use the string after the first '__' as the value (even if it contained '__')
+                    else:
+                        condition |= Q(name=name_value[0], value='__'.join(name_value[1:]))
```
UlysseFG, line 92: you can use the maxsplit=1 param of split() to save the need to join after.

frontend/src/components/filters/filterGroup/FilterMetadata.js
```diff
@@ -0,0 +1,123 @@
...
+                     allowClear
+                     filterOption={false}
+                     options={metadataOptions}
+                     onSearch={onSearchMetadata}
+                     onFocus={onFocusMetadata}
+                     onChange={onNameChange}
+                     value={metadataName}
+                   />
+                 </Form.Item>
+                 <Tooltip title="Leave empty for any value.">
```
UlysseFG, line 90: you can more precise in your tooltip."Leave the value field empty to search for any sample with the given metadata field."

frontend/src/components/filters/FiltersInfos.js
```diff
@@ -31,6 +31,10 @@ const FiltersInfos = ({
         if (value !== "")
           return value
         throw new Error('MIN and MAX values not defined for Filter range')
+      case FILTER_TYPE.METADATA:
+        return valuesArray.reduce((metadataString, metadata) => {
+          return metadataString + metadata.name + ": " + (metadata.value ? metadata.value : "any") + ', '
```
UlysseFG, line 36: Maybe use "*" instead of "any" for the case where the actual content of the field would be any. * is a bit less likely. You could also use an icon to make it clear it is not looking for an user input.

backend/fms_core/filters.py
```diff
@@ -75,6 +76,23 @@ def is_pooled_filter(self, queryset, name, values):
         bool_value = (values == 'true')
         return queryset.filter(is_pooled=bool_value)
 
+    def metadata_filter(self, queryset, name, values):
+        condition = Q()
+        # Metadata is of form: 'name1__value1, name2__value2, name2__Value3'
+        for metadatum in values.split(','):
+            if metadatum:
+                name_value = metadatum.split('__')
```
UlysseFG, line 84: There is still no validation in sample_metadata to prevent "__" in the name. You can add a new metadata_name_validator to prevent specifically that sequence of characters.

backend/fms_core/filters.py
sebastianballesteros: Great suggestion. Didn't know about this capability.

backend/fms_core/models/_validators.py
```diff
@@ -14,3 +15,5 @@
 container_barcode_validator = RegexValidator(re.compile("^[\S]{1,200}$"))
 email_validator = EmailValidator()
 sequence_validator = RegexValidator(re.compile("^[ATCGU]{0,500}$"))
+metadata_name_validator = RegexValidator(regex=re.compile(r"^(?!.*__)[a-zA-Z0-9.\-_]{1,200}$"),
+                                         message="Double underscore i.e '__' are not allowed when naming metadata fields.")
```
UlysseFG, line 19: This message is displayed even when the invalid characters are not the '__', you should point out the valid characters too to make sure the person knows what was wrong with it's name (make it a general message).

backend/fms_core/filters.py
```diff
@@ -75,6 +76,36 @@ def is_pooled_filter(self, queryset, name, values):
...
+        # Build dictionary of form {name1: [value1, value2], name2: [value1] }
+        metadata_dict = {}
+        for metadatum in values.split(','):
+            if metadatum:
+                # Use the string after the first '__' as the value (even if it contains '__')
+                name_value = metadatum.split('__', 1)
+                # Check that there's at least 1 '__' that separates name and value
+                if name_value and len(name_value) > 1:
+                    # Metadata name exists in the dict, append value
+                    if name_value[0] in metadata_dict.keys():
```
UlysseFG, line 89: check if [] is in the list, if it is ignore any other filters for that name.

UlysseFG (approved):

PR #453, opened 2022-12-14
Workflow models

UlysseFG:


backend/fms_core/models/study.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def clean(self):
+        super().clean()
+        errors = {}
+
+        def add_error(field: str, error: str):
+            _add_error(errors, field, ValidationError(error))
+
+        if self.start and self.end and self.start > self.end:
+            add_error("workflow", f"The start step of the workflow need to have a lower or equal order than the end step.")
```
sebastianballesteros, line 42: needs*

backend/fms_core/migrations/0049_v4_0_0.py
```diff
@@ -0,0 +1,348 @@
...
+    STEPS = [
+        # {name, protocol_name, specifications}
+        {"name": "Extraction (DNA)", "protocol_name": "Extraction",
+         "specifications": [{"display_name": "Extraction Type", "sheet_name": "ExtractionTemplate", "column_name": "Extraction Type", "value": "DNA"}]},
+        {"name": "Extraction (RNA)", "protocol_name": "Extraction",
+         "specifications": [{"display_name": "Extraction Type", "sheet_name": "ExtractionTemplate", "column_name": "Extraction Type", "value": "RNA"}]},
+        {"name": "Sample QC", "protocol_name": "Sample Quality Control",
+         "specifications": []},
+        {"name": "Normalization (Sample)", "protocol_name": "Normalization",
+         "specifications": [{"display_name": "Normalization Type", "sheet_name": "Normalization", "column_name": "Robot Norm Choice", "value": "Sample Janus"}]},
```
sebastianballesteros, line 64: We are not supporting Sample Biomek?

UlysseFG: I don't want to have a workflow dedicated to a robot in a normalization step. The idea here was to prefill with a valid value (either Janus or Biomek) i think the default is Janus for Ariane so i used this one. Ideally we would have just value Sample and we could move the janus and biomek to another template column but i'm not sure it's desirable for the lab so i did this little hack.

sebastianballesteros (approved):

PR #455, opened 2022-12-19
Workflow viewset

UlysseFG:
Rebased on the Workflow model PR to get the basic models.

### DIFF
--- a/backend/fms_core/serializers.py
+++ b/backend/fms_core/serializers.py
@@ -32,7 +32,8 @@ from .models import (
     ImportedFile,
     Workflow,
     Step,
-    ReferenceGenome
+    ReferenceGenome,
+    Study
 )
 
 from .models._constants import ReleaseStatus
@@ -80,7 +81,8 @@ __all__ = [
     "PooledSampleSerializer",
     "PooledSampleExportSerializer",
     "WorkflowSerializer",
-    "ReferenceGenomeSerializer"
+    "ReferenceGenomeSerializer",
+    "StudySerializer"
 ]
 
 
@@ -685,3 +687,8 @@ class ReferenceGenomeSerializer(serializers.ModelSerializer):
     class Meta:
         model = ReferenceGenome
         fields = ("id", "assembly_name", "synonym", "genbank_id", "refseq_id", "taxon_id", "size")
+
+class StudySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Study
+        fields = ("id", "letter", "project_id", "workflow_id", "start", "end", "reference_genome_id")
--- a/backend/fms_core/serializers.py
+++ b/backend/fms_core/serializers.py
@@ -670,7 +670,7 @@ class PooledSampleExportSerializer(serializers.Serializer):
 class StepSerializer(serializers.ModelSerializer):
     class Meta:
         model = Step
-        fields = ["name"]
+        fields = ["id", "name", "protocol_id"]
 
 class WorkflowSerializer(serializers.ModelSerializer):
     steps = serializers.SerializerMethodField()
--- a/backend/fms_core/serializers.py
+++ b/backend/fms_core/serializers.py
@@ -29,7 +29,9 @@ from .models import (
     SampleMetadata,
     Sequence,
     Taxon,
-    ImportedFile
+    ImportedFile,
+    Workflow,
+    Step
 )
 
 from .models._constants import ReleaseStatus
@@ -75,7 +77,8 @@ __all__ = [
     "TaxonSerializer",
     "ImportedFileSerializer",
     "PooledSampleSerializer",
-    "PooledSampleExportSerializer"
+    "PooledSampleExportSerializer",
+    "WorkflowSerializer"
 ]
 
 
@@ -660,3 +663,18 @@ class PooledSampleExportSerializer(serializers.Serializer):
             'pedigree',
             ]
 
+class StepSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Step
+        fields = ["name"]
+
+class WorkflowSerializer(serializers.ModelSerializer):
+    steps = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Workflow
+        fields = ("id", "name", "structure", "steps")
+
+    def get_steps(self, instance):
+        steps = instance.steps.all().order_by("StepsOrder__order")
+        return StepSerializer(steps, many=True).data
--- a/backend/fms_core/serializers.py
+++ b/backend/fms_core/serializers.py
@@ -31,7 +31,8 @@ from .models import (
     Taxon,
     ImportedFile,
     Workflow,
-    Step
+    Step,
+    ReferenceGenome
 )
 
 from .models._constants import ReleaseStatus
@@ -78,7 +79,8 @@ __all__ = [
     "ImportedFileSerializer",
     "PooledSampleSerializer",
     "PooledSampleExportSerializer",
-    "WorkflowSerializer"
+    "WorkflowSerializer",
+    "ReferenceGenomeSerializer"
 ]
 
 
@@ -678,3 +680,8 @@ class WorkflowSerializer(serializers.ModelSerializer):
     def get_steps(self, instance):
         steps = instance.steps.all().order_by("StepsOrder__order")
         return StepSerializer(steps, many=True).data
+
+class ReferenceGenomeSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = ReferenceGenome
+        fields = ("id", "assembly_name", "synonym", "genbank_id", "refseq_id", "taxon_id", "size")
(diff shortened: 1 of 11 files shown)


backend/fms_core/viewsets/workflow.py
```diff
@@ -0,0 +1,19 @@
...
+
+from fms_core.models import Workflow, StepOrder
+from fms_core.serializers import WorkflowSerializer
+
+from ._constants import _workflow_filterset_fields
+
+class WorkflowViewSet(viewsets.ModelViewSet):
+    queryset = Workflow.objects.prefetch_related(Prefetch("StepsOrder", queryset=StepOrder.objects.select_related("step"))).all()
+    serializer_class = WorkflowSerializer
+    pagination_class = None
```
ckostiw, line 14: Setting the pagination_class to None changes the response received by the frontend. In 99% of the api calls, the data is returned as an object with a `results` property, and in the frontend we get that data like this: `action.data.results`. 
If the pagination_class is None, then we just receive an array of workflow objects (in `action.data`).
Was this intentional?

ckostiw (changes requested):

backend/fms_core/serializers.py
```diff
@@ -660,3 +663,18 @@ class Meta:
             'pedigree',
             ]
 
+class StepSerializer(serializers.ModelSerializer):
+    class Meta:
```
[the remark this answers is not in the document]
UlysseFG: What about the protocol_id? you can match it in the frontend the the right protocol ...

backend/fms_core/viewsets/workflow.py
UlysseFG: yeah and no i did not expect the format to change for the frontend but i thought this was a call we would not want to page. But we can just send a high limit and circumvent the issue. I'll remove it if this make things simpler.

ckostiw: I think it's simpler (for me) if all the api's return a consistent style of response.

ckostiw (approved):

backend/fms_core/viewsets/workflow.py
UlysseFG: So far we've tried to send back id for foreign keys for the user interface API. The exports generally have more human readable formats.

backend/fms_core/viewsets/study.py
```diff
@@ -13,4 +18,26 @@ class StudyViewSet(viewsets.ModelViewSet):
...
         **_study_filterset_fields,
-    }
\ No newline at end of file
+    }
+
+    def create(self, request, *args, **kwargs):
+        study_data = request.data
+
+        # Generate a sequential letter by counting the number of existing studies tied to the provided project
+        study_count = Study.objects.filter(project=study_data['project']).count()
```
UlysseFG, line 27: I would move these inside the create_study service unless we plan to someday allow non sequential studies (i doubt we'll ever do that).

backend/fms_core/services/study.py
```diff
@@ -0,0 +1,39 @@
+from django.core.exceptions import ValidationError
+from fms_core.models import Study
+
+def create_study(letter, project, worfklow, start, end, reference_genome=None):
```
UlysseFG, line 4: Along with my comment about moving letter generation i would remove the parameter altogether. As long as we cannot delete studies this will be ok. Once we can delete we'll have to change to an algorithm that pick the highest letter then add one (maybe we should do that right away). In any case the choice of letter should never be given to the caller i think.

backend/fms_core/viewsets/study.py
```diff
@@ -13,4 +18,26 @@ class StudyViewSet(viewsets.ModelViewSet):
...
+        # Call create study service
+        study, errors, warnings = create_study(letter=letter,
+                                               project=study_data['project'],
+                                               worfklow=study_data['worfklow'],
+                                               start=study_data['start'],
+                                               end=study_data['end'],
+                                               reference_genome=study_data['reference_genome'])
+
+        if errors:
+            raise ValidationError(err)
```
UlysseFG, line 39: err is not defined.

backend/fms_core/services/study.py
```diff
@@ -0,0 +1,39 @@
...
+        errors['workflow'] = 'Missing workflow.'
+        return study, errors, warnings
+
+    if not start:
+        errors['start'] = 'Missing start.'
+        return study, errors, warnings
+    
+    if not end:
+        errors['end'] = 'Missing end.'
+        return study, errors, warnings
```
UlysseFG, line 27: what about a single return after all the initial tests in case one fails ?

backend/fms_core/services/study.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if not project:
+        errors['project'] = 'Missing project.'
+        return study, errors, warnings
+
+    if not worfklow:
+        errors['workflow'] = 'Missing workflow.'
+        return study, errors, warnings
+
+    if not start:
```
UlysseFG, line 21: I was wondering if we could accept default values as input and fill then internally at insertion time so when we want to test a sample against it's study workflow the information is calculated, but without forcing default entries (the start and end of the full workflow)

backend/fms_core/services/study.py
ckostiw, line 4:
```suggestion
def create_study(letter, project, workflow, start, end, reference_genome=None):
```

backend/fms_core/services/study.py
```diff
@@ -0,0 +1,39 @@
...
+
+    if not letter:
+        errors['letter'] = 'Missing letter.'
+        return study, errors, warnings
+
+    if not project:
+        errors['project'] = 'Missing project.'
+        return study, errors, warnings
+
+    if not worfklow:
```
ckostiw, line 17:
```suggestion
    if not workflow:
```

ckostiw (changes requested):

backend/fms_core/services/study.py
ckostiw: This ('worfklow') causes a Key Error when I try to create a study with the api.

backend/fms_core/viewsets/study.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def create(self, request, *args, **kwargs):
+        study_data = request.data
+
+        # Generate a sequential letter by counting the number of existing studies tied to the provided project
+        study_count = Study.objects.filter(project=study_data['project']).count()
+        letter = string.ascii_uppercase[study_count]
+
+        # Call create study service
+        study, errors, warnings = create_study(letter=letter,
```
ckostiw, line 31: The frontend sends the project ID in the `project` field (and a workflow ID in the `workflow` field). When those are passed to `create_study` I get an exception because the service is expecting Project and Workflow objects, not ID's.

UlysseFG: Forget what i told you earlier i thought you were hitting the default django create endpoint. If you are hitting the create added yesterday ... it's a work in progress. we'll have to fetch to either fetch the objects using the ids or change the service to take directly ids. Do not change your call on behalf of this. I'll tweak the code some so it works.

backend/fms_core/viewsets/study.py
```diff
@@ -0,0 +1,67 @@
...
+        errors = defaultdict(list)
+        study_data = request.data
+
+        try:
+            if study_data['project']:
+                project = Project.objects.get(id=study_data['project'])
+            else:
+                project = None
+        except Project.DoesNotExist:
+            errors['project'].append(f"Project with id {study_data['project']} does not exists.")
```
ckostiw, line 33:
```suggestion
            errors['project'].append(f"Project with id {study_data['project']} does not exist.")
```

backend/fms_core/viewsets/study.py
```diff
@@ -0,0 +1,67 @@
...
+
+        try:
+            if study_data['workflow']:
+                workflow = Workflow.objects.get(id=study_data['workflow'])
+            else:
+                workflow = None
+        except Workflow.DoesNotExist:
+            errors['workflow'].append(f"Workflow with id {study_data['workflow']} does not exists.")
```
ckostiw, line 41:
```suggestion
            errors['workflow'].append(f"Workflow with id {study_data['workflow']} does not exist.")
```

backend/fms_core/viewsets/study.py
```diff
@@ -0,0 +1,67 @@
...
+
+        try:
+            if study_data['reference_genome']:
+                reference_genome = ReferenceGenome.objects.get(id=study_data['reference_genome'])
+            else:
+                reference_genome = None
+        except ReferenceGenome.DoesNotExist:
+            errors['reference_genome'].append(f"Reference genome with id {study_data['reference_genome']} does not exists.")
```
ckostiw, line 49:
```suggestion
            errors['reference_genome'].append(f"Reference genome with id {study_data['reference_genome']} does not exist.")
```

ckostiw (changes requested):

sebastianballesteros (approved):

ckostiw (approved):

PR #457, opened 2022-12-22
Wf add study tab

ckostiw:
Project studies UX

- REQUIRES NPM INSTALL (UPDATED ANTD AND ADDED REDUX TOOLKIT).
- Upgraded antd to 4.22.7 to be able to use setFieldValue() in the Form component. I tried updating to 4.24 but it broke the lineage graph layout. Maybe that can be fixed when we revisit the graph.
- Added typescript support. Ported the redux store setup code to TS, and the networkAction code to make it work with TS components.
- Added new components for the studies edit form in a `studies` folder.
- Started defining interfaces for fms object models


frontend/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+	"printWidth": 140,
+	"semi": false,
+	"singleQuote": true,
+	"tabWidth": 4,
+	"useTabs": true
```
UlysseFG, line 6: Personally i would prefer we don't get tabs mixed in there ...

frontend/src/components/studies/StructuredWorkflows.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * StructuredWorkflows is a simply key based object that uses a workflow structure
+ * name as a key and maps the key to an array of workflows with that structure.
+ */
+export interface StructuredWorkflows {
+	[key: string]: Workflow[]
+}
+
+export function createStructuredWorkflows(workflows: Workflow[]) {
```
UlysseFG, line 11: This would make a pretty reduce one liner ...

frontend/src/models/fms_api_models.ts
```diff
@@ -0,0 +1,106 @@
+/**
```
UlysseFG, line 1: If you want to introduce a structure for models in the frontend maybe you should move the models.js file here. Maybe rename it to backend_models... The idea is not to introduce new varying way to deal with models without integrating the existing code base. In general try to separate effort to overhaul the code from feature creation. That way we don't get many small departure from the way that things were done all over the code (even if your part is nice and shiny it makes the code that much harder to track and understand if the overhaul is not complete).

frontend/src/modules/studies/actions.ts
```diff
@@ -0,0 +1,60 @@
...
+export default {
+	GET,
+	ADD,
+	UPDATE,
+	get,
+	add,
+	update,
+}
+
+// Helper to call list() after another action
```
UlysseFG, line 54: you can remove these commented out lines.

UlysseFG:
In the add study form the workflow detail appears a bit like a repetition of the workflow field. If there is not a finite number of levels for the workflow details maybe we could add another level above the structures to add a "Alternative workflow selection using detailed browser" first level to tell people they can optionally use that way. It is really not that much of a deal. I just find the form atypical.

UlysseFG:
You can rebase on master to include changes we merged.

UlysseFG:
As far as i can see this seem to be breaking a few pages in the app. The list in container and protocols (just a quick check not an extensive list) fail to render the objects that are loaded "on the fly" (the second page for exemple are not preloaded) This could be a result of my older installed packages (pg, node, ... but production has even older node version ... Double check on your side if you have those issues and we'll see what can be done (maybe we can push an upgrade of packages for 4.0.0.

frontend/src/store.ts
```diff
@@ -0,0 +1,28 @@
...
+import { createLogger } from 'redux-logger'
+import rootReducer from './reducers'
+
+const logger = createLogger({
+	level: 'info'
+})
+
+const store = configureStore({
+	reducer: rootReducer,
+	// Disable immutable check because it throws an exception if withItem() tries to modify
```
UlysseFG, line 12: I would recommend creating tickets (in the backlog) for those or we'll forget to fix these issues.

UlysseFG (approved):

PR #460, opened 2023-01-09
First iteration. WIP

sebastianballesteros:


backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,96 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
                                                              study=study_obj)
         except Exception as err:
             errors.append(err)
-    return sample_next_step, errors, warnings
\ No newline at end of file
+    return sample_next_step, errors, warnings
+
+def dequeue_sample_from_study_workflow(sample_obj: Sample, study_obj: Study, order: int=None) -> Tuple[Union[int, None], List[str], List[str]]:
```
UlysseFG, line 53: I don't like the way it could do either a single dequeue or a multiple one. I would create a different service for each. Also we have to keep in mind that finalized sample will have a next_step_step = None. So if you want to remove just that one you would end up removing all sample_next_step. This may change (we may want to store the workflow history of a sample in a more permanent way.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +35,75 @@ def process_row_inner(self, sample, project, action):
...
+                        else:
+                            self.errors['study'] = f"Sample [{sample_obj.name}] is already queued in study [{project['study_letter']}] \
+                                of project [{project['name']}] at step [{project['study_start'] if project['study_start'] else '1'}]"
+                    else:
+                        self.errors['study'] = f"Specified study [{project['study_letter']}] doesn't exist for project [{project['name']}]"
+
+            # Perform a remove project action
+            elif action['name'] == REMOVE_PROJECT_ACTION:
                 # Remove link object if no errors
                 link_removed, self.errors['link'], self.warnings['link'] = remove_link(sample=sample_obj,
```
UlysseFG, line 64: Move this in the else below since we don't want to execute that if there is an error.

backend/fms_core/template_importer/importers/project_link_sample.py
```diff
@@ -16,6 +16,8 @@ def import_template_inner(self):
         for row_id, row_data in enumerate(project_link_samples_sheet.rows):
             project = {
                 'name': str_cast_and_normalize(row_data['Project Name']),
+                'study_letter': str_cast_and_normalize(row_data['Study']),
+                'study_start': str_cast_and_normalize(row_data['Workflow Step Start']) if row_data['Workflow Step Start'] else None,
```
UlysseFG, line 20: Since this is used both to insert and remove maybe we should rename that column and variables using "step order" instead of "step/study start".

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +35,75 @@ def process_row_inner(self, sample, project, action):
...
-            elif action['name'] == REMOVE_ACTION:
+                # Queue sample to study if specified
+                if project['study_letter']:
+                    study_obj, self.errors['study'], self.warnings['study'] = get_study(project_obj, project['study_letter'])
+
+                    if study_obj:
+                        study_start = project['study_start']
+
+                        # Check if sample is already queued 
+                        is_sample_queued, self.errors['sample_queued'], self.warnings['sample_queued'] = is_sample_queued_in_study(sample_obj, study_obj, study_start)
```
UlysseFG, line 52: If you add a sample to a project there is no real possibility to have it already queued to a study associated to it.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +35,75 @@ def process_row_inner(self, sample, project, action):
...
+                        if not is_sample_queued:
+                            _, self.errors['add_to_study'], self.warnings['queue_to_study'] = queue_sample_to_study_workflow(sample_obj, study_obj, study_start)
+                        else:
+                            self.errors['study'] = f"Sample [{sample_obj.name}] is already queued in study [{project['study_letter']}] \
```
UlysseFG, line 56: you can have a sample queued to multiple steps of the study. This is not an error unless you are adding the same sample at the same step order of the study, If the study_start is empty you'll get an error which should not be the case unless the sample is queued on the first step of the study workflow.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +35,75 @@ def process_row_inner(self, sample, project, action):
...
+                # If the study is not specified, dequeue the sample from all the studies of the selected project
+                else:
+                    num_dequeued = 0
+                    for study in project_obj.studies.all():
+                         # Check if sample is queued 
+                        is_sample_queued, self.errors['sample_queued'], self.warnings['sample_queued'] = is_sample_queued_in_study(sample_obj, study)
+                        
+                        if is_sample_queued:
+                            # Don't specify order to be able remove all instances of sample in the study, regardless of step
+                            num_dequeued, self.errors['remove_from_study'], self.warnings['queue_to_study'] = dequeue_sample_from_study_workflow(sample_obj, study_obj)
```
UlysseFG, line 80: Use a different service that remove all sample_next_step for that sample.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
UlysseFG: I think we will be adding a validation on sample_next_step that verifies that the sample_next_step study has the same project as the sample. This would mean that we need to add project before adding studies and remove studies before removing project in order not to trigger this validation.

UlysseFG:
The Project Link name is a bit off (in code and template). Not sure it's worth changing.

backend/fms_core/services/sample_next_step.py
sebastianballesteros: I don't understand why would you end up removing all sample_next_step instances when finalized samples get to the last step ?

UlysseFG: If you specifically want to remove that sample_next_step ... it's step_order is None but providing an empty order to the service will be understood as remove all sample_next_step...

sebastianballesteros:
Project_study_link?

sebastianballesteros:
TODO: Change Project_link_samples to Project_study_link_samples all over the application

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
                                                              study=study_obj)
         except Exception as err:
             errors.append(err)
-    return sample_next_step, errors, warnings
\ No newline at end of file
+    return sample_next_step, errors, warnings
+
+def dequeue_sample_from_specifc_step_study_workflow(sample_obj: Sample, study_obj: Study, order: int) -> Tuple[Union[int, None], List[str], List[str]]:
```
UlysseFG, line 53: specifc => specific

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+        errors.append(f"A valid sample instance must be provided.")
+
+    if not isinstance(study_obj, Study):
+        study_obj = None
+        errors.append(f"A valid study instance must be provided.")
+    
+    # Dequeuing from study workflow by deleting the SampleNextStep instance
+    if sample_obj and study_obj and not errors:
+        try:
+            num_deleted, _ = SampleNextStep.objects.filter(sample=sample_obj, study=study_obj).delete()
```
UlysseFG, line 124: This is a batch delete() this will not trigger our revision update. To use the model delete (instead of the manager delete) you would have to iterate on instances.

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+        sample_obj = None
+        errors.append(f"A valid sample instance must be provided.")
+
+    if not isinstance(study_obj, Study):
+        study_obj = None
+        errors.append(f"A valid study instance must be provided.")
+    
+    if not order:
+        order = None
+        errors.append(f"A step order must be provided.  ")
```
UlysseFG, line 79: Extra spaces in error message.

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+    step_order=None
+    # Dequeuing from a specific step
+    if order is not None:
+        try:
+            step_order = StepOrder.objects.get(order=order, workflow=study_obj.workflow)
+        except StepOrder.DoesNotExist:
+            errors.append(f"No step found for the given order.")
+    
+    # Dequeuing from study workflow by deleting the SampleNextStep instance
+    if sample_obj and study_obj and step_order and not errors:
```
UlysseFG, line 90: It might be good to allow this service to remove the samples from completed workflow (just in case). For that, we have to allow step_order is None.

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+    Args:
+        `sample_obj`: Sample instance to be dequeued from the study workflow.
+        `study_obj`: Study instance currently associated to the sample.
+        `order`: Positive integer indicating the current position of the sample in the workflow.
+                 
+    Returns:
+        Tuple containing the number of SampleNextStep instances deleted, the error messages and the warning messages. 
+
+    """
+    dequeued = None
```
UlysseFG, line 65: this is not used ...

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+    if order is not None:
+        try:
+            step_order = StepOrder.objects.get(order=order, workflow=study_obj.workflow)
+        except StepOrder.DoesNotExist:
+            errors.append(f"No step found for the given order.")
+    
+    # Dequeuing from study workflow by deleting the SampleNextStep instance
+    if sample_obj and study_obj and step_order and not errors:
+        try:
+            num_deleted, _ = SampleNextStep.objects.filter(sample=sample_obj, study=study_obj, step_order=step_order).delete()
```
UlysseFG, line 92: This uses the manager delete... use first() to get the model delete and ensure only a single object is deleted (the unique constraint does prevent this for all step_order different from None, but cannot enforce unique on None values).

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+        except Exception as err:
+            errors.append(err)
+    return num_deleted, errors, warnings
```
UlysseFG, line 95: cast num_deleted to a boolean ...

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+def dequeue_sample_from_specifc_step_study_workflow(sample_obj: Sample, study_obj: Study, order: int) -> Tuple[Union[int, None], List[str], List[str]]:
+    """
+    Deletes a SampleNextStep instance to indicate the removal of a sample from a workflow at a specific step. 
+    Args:
+        `sample_obj`: Sample instance to be dequeued from the study workflow.
+        `study_obj`: Study instance currently associated to the sample.
+        `order`: Positive integer indicating the current position of the sample in the workflow.
+                 
+    Returns:
+        Tuple containing the number of SampleNextStep instances deleted, the error messages and the warning messages.
```
UlysseFG, line 62: Not much sense in having a number when you can either have one or zero entries removed.

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+    Meaning that if the sample is queued in 2 different steps in the provided study, both instances will be deleted. 
+    Args:
+        `sample_obj`: Sample instance to be dequeued from the study workflow.
+        `study_obj`: Study instance currently associated to the sample.
+                 
+    Returns:
+        Tuple containing the number of SampleNextStep instances deleted, the error messages and the warning messages. 
+
+    """
+    dequeued = None
```
UlysseFG, line 109: Not used ...

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+    if sample_obj and study_obj and not errors:
+        try:
+            num_deleted, _ = SampleNextStep.objects.filter(sample=sample_obj, study=study_obj).delete()
+        except Exception as err:
+            errors.append(err)
+    return num_deleted, errors, warnings
+
+def is_sample_queued_in_study(sample_obj: Sample, study_obj: Study, order: int=None) -> Tuple[Union[bool, None], List[str], List[str]]:
+    """
+
```
UlysseFG, line 131: Missing description / too many empty lines

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +48,125 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+        try:
+            step_order = StepOrder.objects.get(order=order, workflow=study_obj.workflow)
+        except StepOrder.DoesNotExist:
+            errors.append(f"No step found for the given order.")
+    
+    # Specify step_order in filter if it is provided
+    filters = dict(
+        sample=sample_obj,
+        study=study_obj,
+        **(dict(step_order=step_order) if step_order is not None else dict())
```
UlysseFG, line 165: what if i want to know if a sample completed a study workflow ... the expected step order would be none. The function would assume i look for any order... might be a good idea to either split the function or to add another boolean param to change the behaviour in a predictable way.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,94 @@ def process_row_inner(self, sample, project, action):
                               f"is named {sample_obj.name} not {sample['sample_name']}."
                 self.warnings['sample'].append(warning_msg)
 
-            # Check if link exists to ensure there's not a duplicated association
-            if action['name'] == ADD_ACTION:
+            # Perform an add project and/or study action
+            if action['name'] == ADD_PROJECT_ACTION:
```
UlysseFG, line 36: This action should allow both project and study add at once. a missing study would add just the project if both are present you add both starting with project. This is a less orthogonal organization, but it will save user the extra entry that is required by separating add project and study. It also better mirror the remove project (which removes the studies).

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,94 @@ def process_row_inner(self, sample, project, action):
...
+                if project['study_letter']:
+                    self.errors['add_to_project'] = f"A study was provided. If you want to add the sample to a study use the [Add to study] action."
+
+            
+            # Perform an add study action
+            elif action['name'] == ADD_STUDY_ACTION:
+                # Make sure the sample is already associated to the project of the given study 
+                for derived_sample in sample_obj.derived_samples.all():
+                    if derived_sample.project is not None and derived_sample.project.id != project_obj.id:
+                        self.errors['add_to_study'] = (f"[Sample {sample_obj.name}] is not associated to project [{project_obj.name}].")
```
UlysseFG, line 51: Not true. The pools are a complex case ... 

The way they will be dealt with as far as i see it they will be queued by study as long as a single sample is part of that study. So if we receive a ready to sequence pool. All samples in the pool will be part of a single study (or different study in the same project). We will queue de pool sample once for the whole study. For this case your test would work but most pools will be made "in house" and will contain a mix of different project samples). 

Naturally when the samples in a pool will pass the pool step the next step will use the same sample (pool) in each study they belong to. This will lead to multiple instance of the same sample_next_step being scheduled, we will ignore duplicates a single pool by study will support each sample queued to the workflow for that study. That sample derived sample will not all belong to the same project. 

So if we want to manually add a pool to the study we have to support this too. I would include is_pool to check if everything is ok. If you have a pool you can have sample that do not belong to the project but at least one sample as to belong to the project. Also if you have a pool and there is already an identical sample_next_step you may want to return a simple warning. 

Lastly these tests should be included in the queue_sample_to_study_workflow service instead since they likely apply to all cases we will add sample to studies.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,94 @@ def process_row_inner(self, sample, project, action):
...
+                    for study in project_obj.studies.all():
+                         # Check if sample is queued 
+                        is_sample_queued, self.errors['sample_queued'], self.warnings['sample_queued'] = is_sample_queued_in_study(sample_obj, study)
+                        
+                        if is_sample_queued:
+                            # Don't specify order to be able remove all instances of sample in the study, regardless of step
+                            num_dequeued, self.errors['remove_from_study'], self.warnings['queue_to_study'] = dequeue_sample_from_all_steps_study_workflow(sample_obj, study_obj)
+                            num_dequeued += num_dequeued
+                    if num_dequeued > 0:
+                        self.warnings['remove_project'] = f"Sample [{sample_obj.name}] was removed from all the studies belonging to project [{project_obj.name}]"
```
UlysseFG, line 92: Maybe use a conditional form to make this warning less final. "Removing the link between the project X and the sample Y will also remove Sample Y from Z studies."

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,94 @@ def process_row_inner(self, sample, project, action):
...
+                
+                if not any(self.errors.values()):
+                    # Remove link object if no errors
+                    link_removed, self.errors['link'], self.warnings['link'] = remove_link(sample=sample_obj,
+                                                                                        project=project_obj)
+                
+
```
UlysseFG, line 99: too much empty space.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,94 @@ def process_row_inner(self, sample, project, action):
...
+            # Perform a remove study action
+            elif action['name'] == REMOVE_STUDY_ACTION:
+                # Dequeue sample from a specified study 
+                if project['study_letter']:
+                    study_obj, self.errors['study'], self.warnings['study'] = get_study(project_obj, project['study_letter'])
+
+                    if study_obj:
+                        # To avoid empty step orders
+                        step_order = project['step_order'] if project['step_order'] else 1
```
UlysseFG, line 108: ick ... no ... i would prefer we make it impossible for users to remove sample from completed study than assuming that an empty field means 1 ;). I would raise an error for remove_study if the order is empty. Removing a sample that completed a study will be possible from the direct service call for us and we can do curations otherwise (i don't see any use case to do that for the lab).

backend/fms_core/tests/test_services/sample_next_step_test.py
```diff
@@ -32,22 +32,71 @@ def setUp(self):
...
     def test_queue_sample_to_valid_step(self):
         sample_next_step, errors, warnings = queue_sample_to_study_workflow(self.sample, self.study, 3)
-        self.assertEqual(errors, [])
-        self.assertEqual(warnings, [])
         self.assertTrue(isinstance(sample_next_step, SampleNextStep))
         self.assertEqual(sample_next_step.step_order.order, 3)
         self.assertEqual(sample_next_step.sample, self.sample)
 
+        #Test sample queued
+        is_queued, errors, warnings = is_sample_queued_in_study(self.sample, self.study, 3)
```
UlysseFG, line 52: This should be in another test.

backend/fms_core/tests/test_services/sample_next_step_test.py
```diff
@@ -32,22 +32,71 @@ def setUp(self):
...
         self.assertIsNone(sample_next_step)
+
+        #Test sample queued
+        is_queued, errors, warnings = is_sample_queued_in_study(self.sample, self.study)
+        self.assertFalse(is_queued)
+
+    def test_dequeue_sample_from_valid_step(self):
+        # Queue sample first and test it
+        sample_next_step, errors, warnings = queue_sample_to_study_workflow(self.sample, self.study)
+        is_queued, errors, warnings = is_sample_queued_in_study(self.sample, self.study)
```
UlysseFG, line 70: you should not test using another service as much as possible. This will just lead to cascade errors of the testing service as an error.

backend/fms_core/utils.py
```diff
@@ -97,6 +97,11 @@ def str_cast_and_normalize_lower(s) -> Union[str, None]:
     result = str_cast_and_normalize(s)
     return result.lower() if result is not None else result
 
+def str_cast_to_int_or_none(s) -> Union[int, None]:
```
UlysseFG, line 100: input_to_integer_and_none already exists in ../template_importer/_utils.py

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
sebastianballesteros: `Also if you have a pool and there is already an identical sample_next_step you may want to return a simple warning.`

Identical meaning (sample, study)?

UlysseFG: sample, study and order

backend/fms_core/services/sample_next_step.py
```diff
@@ -48,4 +49,160 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order:
...
+        try:
+            instances_to_delete = SampleNextStep.objects.filter(sample=sample_obj, study=study_obj)
+            for instance in instances_to_delete:
+                instance.delete()
+                num_deleted += 1
+        except Exception as err:
+            errors.append(err)
+    return num_deleted, errors, warnings
+
+def is_sample_queued_in_study(sample_obj: Sample, study_obj: Study, order: int=None) -> Tuple[Union[bool, None], List[str], List[str]]:
```
UlysseFG, line 139: The function till never return None as the first element of the Tuple.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,111 @@ def process_row_inner(self, sample, project, action):
...
+
+                        # Check if sample is already queued 
+                        is_sample_queued, self.errors['sample_queued'], self.warnings['sample_queued'] = is_sample_queued_in_study(sample_obj, study_obj, step_order)
+                        if not is_sample_queued:
+                            _, self.errors['add_to_study'], self.warnings['add_to_study'] = queue_sample_to_study_workflow(sample_obj, study_obj, step_order)
+                        else:
+                            self.errors['study'] = f"Sample [{sample_obj.name}] is already queued in study [{project['study_letter']}] \
+                                of project [{project['name']}] at step [{project['step_order']}]"
```
UlysseFG, line 55: missing "." at the end of the error message.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,111 @@ def process_row_inner(self, sample, project, action):
...
+                    else:
+                        self.errors['study'] = f"Specified study [{project['study_letter']}] doesn't exist for project [{project['name']}]"
```
UlysseFG, line 57: missing "." at the end of the error message.

backend/fms_core/utils.py
UlysseFG: you can remove it if it's not used.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
sebastianballesteros: I can return a warning but wouldn't it be caught by the unique constraint [sample, study, order]?

UlysseFG: Sample1, Study1, None is considered different from Sample1, Study1, None but Sample1, Study1, Order1 is considered identical to Sample1, Study1, Order1. this is part of why it would be good to send a warning. I don't think this really apply to the row handler function since we would not want people to add sample at the end of the workflow, but these checks should be done in the service and at that level we would want that.
If i assign a sample to the end of the workflow and the sample already completed the workflow once we may not want a second copy of the sample_next_step identical (to us not to the DB) to the one already there. Also when pooling all samples will become a single sample for the next step so there will likely be duplicate sample_next_step after completing the pooling step, one way to do the next step would be to make a set to ensure we don't have duplicates but we could also call a service  and have it ignore duplicates with a warning. The ignore option is not really a good practice, but it might be faster and simpler.
So the tldr version is if you do the test in the row_handler the unique constraint validation should be sufficient and an error is possibly a good way to tell the user the sample is already queued there.

UlysseFG:
Template only contain text all colors and borders are gone.

UlysseFG:
Actions and template in the app should mention study ... Link Projects and Studies to Samples ... Project and Study link samples Template ... atm there is no mention anywhere visible to the user that this can be done that way. If you don't want to rename all code identifier, it's fine but we need to provide some form of visual cue to the user so they can find that functionality.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -5,9 +5,13 @@
 from fms_core.services.sample import get_sample_from_container
 from fms_core.services.project import get_project
 from fms_core.services.project_link_samples import create_link, remove_link
+from fms_core.services.study import get_study
+from fms_core.services.sample_next_step import queue_sample_to_study_workflow, is_sample_queued_in_study, dequeue_sample_from_specific_step_study_workflow, dequeue_sample_from_all_steps_study_workflow
 
-ADD_ACTION = "ADD"
-REMOVE_ACTION = "REMOVE"
+ADD_PROJECT_ACTION = "ADD TO PROJECT"
```
UlysseFG, line 11: Should mention optional addition to study too. "AND STUDY"

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,110 @@ def process_row_inner(self, sample, project, action):
...
-                # Remove link object if no errors
-                link_removed, self.errors['link'], self.warnings['link'] = remove_link(sample=sample_obj,
-                                                                                       project=project_obj)
+                # Queue sample to study if provided
+                if project['study_letter']:
+                    study_obj, self.errors['study'], self.warnings['study'] = get_study(project_obj, project['study_letter'])
+
+                    if study_obj:
+                        # To avoid empty step orders
+                        step_order = project['step_order'] if project['step_order'] else 1
```
UlysseFG, line 47: 1 is not the default of all studies ... queue_sample_to_study_workflow would take care of it if you left it empty.

backend/fms_core/template_importer/row_handlers/project_link_samples/project_link_samples.py
```diff
@@ -28,18 +32,110 @@ def process_row_inner(self, sample, project, action):
...
+                # If the study is not specified, dequeue the sample from all the studies of the selected project
+                else:
+                    num_dequeued = 0
+                    for study in project_obj.studies.all():
+                         # Check if sample is queued 
+                        is_sample_queued, self.errors['sample_queued'], self.warnings['sample_queued'] = is_sample_queued_in_study(sample_obj, study)
+                        
+                        if is_sample_queued:
+                            # Don't specify order to be able remove all instances of sample in the study, regardless of step
+                            num_dequeued, self.errors['remove_from_study'], self.warnings['queue_to_study'] = dequeue_sample_from_all_steps_study_workflow(sample_obj, study_obj)
```
UlysseFG, line 104: study_obj used before being initialized.You need to use study instead ...

UlysseFG (approved):

PR #462, opened 2023-01-09
Change Version, DB schema and added Node installation to the deployment file.

UlysseFG:


DEPLOYMENT.md
```diff
@@ -86,6 +86,11 @@ On this page we list the various steps needed for deployments. The first section
     mkdir -p /usr/lib64/llvm5.0/bin
     ln -s /opt/rh/llvm-toolset-7/root/usr/bin/llvm-lto /usr/lib64/llvm5.0/bin/
     ```
+  * Install Node.js
```
sebastianballesteros, line 89: Maybe we need to specify the version?

UlysseFG: When checking with Chris i could not find a limitation on version ... Chris told me you had 16.0.0. but i have lower and chris higher and both are working fine. I am not convinced this was a node version issue (16.0.0 on macOS seem to have issues with disconnect... maybe this led to running out of available connections to the backend). If you have an idea on what are the limitation i'll add them but i am not really sure.

sebastianballesteros (approved):

PR #470, opened 2023-01-17
Sample Next Step viewset

sebastianballesteros:


backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -0,0 +1,90 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.decorators import action
+
+from fms_core.models import SampleNextStep, StepSpecification
+from fms_core.serializers import SampleNextStepSerializer, StepSpecificationSerializer
+
+class SampleNextStepViewSet(viewsets.ModelViewSet):
+    queryset = SampleNextStep.objects.all()
+    serializer_class = SampleNextStepSerializer
```
UlysseFG, line 11: The serializer may not contain enough information for the frontend to build the lab work filtered list (2nd page in the lab work menu hierarchy). You would need to include the step and step specification in each serialized entry. We could also build a step/step_order endpoint with these details and stitch them in the frontend but we need them somehow.

backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -0,0 +1,90 @@
...
+                step_specs = StepSpecificationSerializer(step_specs, many=True).data
+
+                # The protocol of the step is already in the dictionary
+                if step_obj.protocol.id in sample_next_step_by_protocol["protocols"]:
+                    sample_next_step_by_protocol["protocols"][step_obj.protocol.id]["count"] += 1
+                    # Look for the specific step within the protocol
+                    step = next((step for step in sample_next_step_by_protocol["protocols"][step_obj.protocol.id]["steps"] if step["name"] == step_obj.name), None)
+                    # If the step is found add to the count
+                    if step:
+                        step["count"] += 1
```
UlysseFG, line 66: Not sure why you have to iterate instead of adding a count to a queryset.

backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -0,0 +1,90 @@
...
+                    protocols:
+                        {
+                            name
+                            count
+                            steps: {
+                                name 
+                                count
+                                step_specifications
+                            }
+
```
UlysseFG, line 39: empty space ...

backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -0,0 +1,90 @@
...
+        As well as their step specifications.
+
+        Args:
+            `request`: The request object received then whe API call was made.
+            `*args`: Arguments to set on the view.
+            `**kwargs`: Additional properties to set on the view.
+                    
+        Returns:
+          An object of the form:
+          {
```
UlysseFG, line 27: Might be useful to indicate where are the arrays in the structure.

backend/fms_core/serializers.py
```diff
@@ -667,21 +673,39 @@ class Meta:
...
     class Meta:
         model = Workflow
         fields = ("id", "name", "structure", "steps")
-
+    
     def get_steps(self, instance):
-        steps = instance.steps.all().order_by("StepsOrder__order")
-        return StepSerializer(steps, many=True).data
+        steps = instance.steps.all().annotate(order=F("steps_order__order")).order_by("order")
+        serialized_data =  StepWithOrderSerializer(steps, many=True)
```
UlysseFG, line 707: extra space before serializer name

backend/fms_core/serializers.py
```diff
@@ -667,21 +673,39 @@ class Meta:
...
+    order = serializers.SerializerMethodField(read_only=True)
+    class Meta:
+        model = Step
+        fields = ["id", "name", "protocol_id", "order"]
+    
+    def get_order(self, instance):
+        if hasattr(instance, 'order'):
+            return instance.order
+        else:
+            return ""
```
UlysseFG, line 697: use None instead of ""

backend/fms_core/serializers.py
```diff
@@ -667,21 +673,39 @@ class Meta:
...
 class StepSerializer(serializers.ModelSerializer):
+    step_specifications = StepSpecificationSerializer(read_only=True, many=True)
     class Meta:
         model = Step
-        fields = ["id", "name", "protocol_id"]
+        fields = ["id", "name", "protocol_id", "step_specifications"]
 
-class WorkflowSerializer(serializers.ModelSerializer):
-    steps = serializers.SerializerMethodField()
+class StepWithOrderSerializer(serializers.ModelSerializer):
```
UlysseFG, line 687: Not sure i like the StepWithOrder serializer name and content. I think we should review this to be a StepOrder serializer that also include a step name and protocol_id for convenience but at the moment it has a unique step id but the order is not related uniquely to that id but instead to the step order instance. This could cause issues. I would use the StepOrderSerializer (step_order_id, step_id, step_name, protocol_id, worflow_id, order) for the workflow. It is not a big change i think but it might make it more usefull and less confusing.

backend/fms_core/serializers.py
```diff
@@ -696,4 +715,11 @@ class Meta:
 class StudySerializer(serializers.ModelSerializer):
     class Meta:
         model = Study
-        fields = ("id", "letter", "project_id", "workflow_id", "start", "end")
+        fields = ("id", "letter", "project_id", "workflow_id", "start", "end", "reference_genome_id")
```
UlysseFG, line 718: nope ...you must have merged an old version of master in ...

sebastianballesteros: True.

sebastianballesteros: It was not an old version. There was a conflict and I chose the wrong line.

backend/fms_core/serializers.py
```diff
@@ -672,21 +678,34 @@ class Meta:
...
     class Meta:
         model = Workflow
         fields = ("id", "name", "structure", "steps")
-
+    
     def get_steps(self, instance):
-        steps = instance.steps.all().order_by("StepsOrder__order")
-        return StepSerializer(steps, many=True).data
+        steps = instance.steps.all().order_by("steps_order__order")
+        serialized_data = StepSerializer(steps, many=True)
```
UlysseFG, line 707: Workflow should use the step_order serializer if possible ...

sebastianballesteros: Apparently Chris doesn't need the step order for the Workflow endpoint.

UlysseFG: It would be cleaner for the workflow to reference the step_order than the step (the order of the step being directly related to the step_order anyway). By definition the workflow is a sequence (or an ordering) of steps. If we ever have to add information to the workflow it's likely to be in the workflow or step_order models. The step model is shared between workflows and if we include it we may come to have fields that conflict.

PR #475, opened 2023-01-24
Added the sample metadata export

sebastianballesteros:


backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+                metadata_per_biosample[biosample] = [metadata] 
+
+        serialized_data = []
+        if not samples:
+            serialized_data.append({}) # Allow the returned csv file to be named instead of random name.
+        for sample in samples.values():
+            derived_sample = derived_samples[sample["first_derived_sample"]]
+            is_library = derived_sample["is_library"]
+            is_pool = sample["count_derived_samples"] > 1
+            metadata = metadata_per_biosample[derived_sample["biosample__id"]]
```
UlysseFG, line 398: This will be wrong for pools ....

backend/fms_core/viewsets/sample.py
```diff
@@ -324,14 +324,38 @@ def list_export(self, _request):
         serialized_data = self.fetch_export_data()
         return Response(serialized_data)
 
+    @action(detail=False, methods=["get"])
+    def list_export_metadata(self, _request):
+        self.queryset = self.filter_queryset(self.get_queryset())
+        serialized_data = self.fetch_export_metadata()
```
UlysseFG, line 330: We need to think about the use cases ... Will people want them by sample, will they want the pools metadata broken down, should we prevent pools to be included ... I would think people will want to extract the meta data for any sample even if they are pooled ... so we would need to return them broken down by biosample/derived sample...

backend/fms_core/viewsets/_fetch_data.py
UlysseFG: Since there is nothing to be build around this we can "break down" the sample into derived samples ... each line would be a derived sample/ biosample... with their matching metadata.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            if biosample in metadata_per_biosample.keys():
+                metadata_per_biosample[biosample].append(metadata)
+            else:
+                metadata_per_biosample[biosample] = [metadata] 
+
+        serialized_data = []
+        if not samples:
+            serialized_data.append({}) # Allow the returned csv file to be named instead of random name.
+        for sample in samples.values():
+            derived_sample = derived_samples[sample["first_derived_sample"]]
```
UlysseFG, line 395: iterated on derived_samples too do not use the first_derived_sample ...

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            is_library = derived_sample["is_library"]
+            is_pool = sample["count_derived_samples"] > 1
+            metadata = metadata_per_biosample[derived_sample["biosample__id"]]
+            
+            data = {
+                'sample_name': sample["name"],
```
UlysseFG, line 401: Sample name repeated for multiple samples in the same pool.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+                'biosample_id': derived_sample["biosample__id"] if not is_pool else None,
+                'alias': derived_sample["biosample__alias"] if not is_pool or not is_library else None,
```
UlysseFG, line 403: alias is the better identifier here ... always display.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            serialized_data.append({}) # Allow the returned csv file to be named instead of random name.
+        for sample in samples.values():
+            derived_sample = derived_samples[sample["first_derived_sample"]]
+            is_library = derived_sample["is_library"]
+            is_pool = sample["count_derived_samples"] > 1
+            metadata = metadata_per_biosample[derived_sample["biosample__id"]]
+            
+            data = {
+                'sample_name': sample["name"],
+                'biosample_id': derived_sample["biosample__id"] if not is_pool else None,
```
UlysseFG, line 402: do not mask for pools.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+                'alias': derived_sample["biosample__alias"] if not is_pool or not is_library else None,
+                'container_kind': sample["container__kind"],
+                'container_name': sample["container__name"],
+                'container_barcode': sample["container__barcode"],
+                'coordinates': sample["coordinates"],
+                'location_barcode': sample["container__location__barcode"] or "",
+                'location_coord': sample["container__coordinates"] or "",
+                'project': derived_sample["project__name"] if not is_pool else None,
```
UlysseFG, line 410: Project is not really relevant ... you can keep it but do not mask for pools ...

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            
+            data = {
+                'sample_name': sample["name"],
+                'biosample_id': derived_sample["biosample__id"] if not is_pool else None,
+                'alias': derived_sample["biosample__alias"] if not is_pool or not is_library else None,
+                'container_kind': sample["container__kind"],
+                'container_name': sample["container__name"],
+                'container_barcode': sample["container__barcode"],
+                'coordinates': sample["coordinates"],
+                'location_barcode': sample["container__location__barcode"] or "",
```
UlysseFG, line 408: location barcode and coord might be too much ...

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            derived_sample = derived_samples[sample["first_derived_sample"]]
+            is_library = derived_sample["is_library"]
+            is_pool = sample["count_derived_samples"] > 1
+            metadata = metadata_per_biosample[derived_sample["biosample__id"]]
+            
+            data = {
+                'sample_name': sample["name"],
+                'biosample_id': derived_sample["biosample__id"] if not is_pool else None,
+                'alias': derived_sample["biosample__alias"] if not is_pool or not is_library else None,
+                'container_kind': sample["container__kind"],
```
UlysseFG, line 404: Not sure we need container_kind ...

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            'container__coordinates',
+            'count_derived_samples',
+            'first_derived_sample',
+        )
+        samples = {s["id"]: s for s in self.queryset}
+
+        first_derived_sample_ids = self.queryset.values_list("first_derived_sample", flat=True)
+        derived_sample_values_queryset = (
+            DerivedSample.objects
+            .filter(id__in=first_derived_sample_ids)
```
UlysseFG, line 367: get all derived samples ...

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,87 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+            .annotate(is_library=ExpressionWrapper(Q(library__isnull=False), output_field=BooleanField()))
+            .values(
+                'id',
+                'biosample__id',
+                'biosample__alias',
+                'project__name',
+                'is_library',
```
UlysseFG, line 374: library status not really relevant.

backend/fms_core/viewsets/sample.py
```diff
@@ -324,14 +324,38 @@ def list_export(self, _request):
...
             fields = SampleExportSerializer.Meta.fields
             context['header'] = fields
             context['labels'] = {i: i.replace('_', ' ').capitalize() for i in fields}
+        elif self.action == 'list_export_metadata':
+            fields = self.get_metadata_fields()
+            context['header'] = fields
+            context['labels'] = {i: i.replace('_', ' ').capitalize() for i in fields}
         return context
 
+    def get_metadata_fields(self):
```
UlysseFG, line 345: This is a likely source of bugs ... we should store the columns after returning from the fetch into a self variable and pick it from there in the get_renderer_context ... otherwise we risk having synchronization issues between fetch and this function.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -333,6 +333,78 @@ def fetch_export_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
+                'biosample_id': derived_sample["biosample__id"],     
+                'container_name': sample["container__name"],
+                'container_barcode': sample["container__barcode"],
+                'coordinates': sample["coordinates"], 
+                'project': derived_sample["project__name"],
+                **dict((item["name"], item["value"]) for item in metadata)
+            }
+            serialized_data.append(data)
+
+        return (metadata_names, serialized_data)
```
UlysseFG, line 406: metadata_names => data.keys() better to provide all fields and make sure we don't have mismatch between the fetch and the get_renderer_context.

UlysseFG (approved):

PR #476, opened 2023-01-24
Wf list study samples

ckostiw:
List samples in the study view, grouped by step.


frontend/src/components/projects/ProjectsAssociatedSamples.js
```diff
@@ -45,7 +46,8 @@ const getTableColumns = (sampleKinds, individualsByID) => [
         const individual = sample.individual
         return (individual &&
           <Link to={`/individuals/${individual}`}>
-            {withIndividual(individualsByID, individual, individual => individual.cohort, "loading...")}
+            <WithIndividualRenderComponent objectID={individual} render={individual => <>{individual.cohort}</>} placeholder={"Loading..."}/>
+            {/* {withIndividual(individualsByID, individual, individual => individual.cohort, "loading...")} */}
```
UlysseFG, line 50: remove old code

frontend/src/components/shared/WithItemRenderComponent.tsx
```diff
@@ -0,0 +1,85 @@
...
+export const WithContainerRenderComponent = WithItemRenderComponentFactory<typeof withContainer, Container>(withContainer, selectContainersByID)
+export const WithIndexRenderComponent = WithItemRenderComponentFactory<typeof withIndex, Index>(withIndex, selectIndicesByID)
+export const WithIndividualRenderComponent = WithItemRenderComponentFactory<typeof withIndividual, Individual>(withIndividual, selectIndividualsByID)
+export const WithProjectRenderComponent = WithItemRenderComponentFactory<typeof withProject, Project>(withProject, selectProjectsByID)
+export const WithSampleRenderComponent = WithItemRenderComponentFactory<typeof withSample, Sample>(withSample, selectSamplesByID)
+export const WithUserRenderComponent = WithItemRenderComponentFactory<typeof withUser, User>(withUser, selectUsersByID)
+
+
+/*
+import { Menu, Typography } from 'antd'
```
UlysseFG, line 68: If we really need an example maybe we can improve the function docstring instead.

frontend/src/components/studySamples/TODO.md
```diff
@@ -0,0 +1,17 @@
+# Study Samples TODO's
```
UlysseFG, line 1: Move the Todo to redmine...

frontend/src/models/protocols.ts
```diff
@@ -0,0 +1,26 @@
+
+/* Defines enum values for the protocol names. */
+export enum ProtocolNames {
```
UlysseFG, line 3: don't copy the DB in the frontend ... I doubt you need all those names mapped in the frontend. Keep to what is used. We don't want to keep updating those if they don't have a use.

frontend/src/components/studySamples/LibraryTableColumns.tsx
```diff
@@ -0,0 +1,222 @@
+import { TableColumnType } from 'antd'
```
UlysseFG, line 1: Unless you intend to use these definition outside they should be nested within components/studies/ unless it is an original structure. This appear to be a whole lot of definition. Keep only what is used. I'm not sure this will be useful. We'll have to check with the PM and the lab see if there is need for that level of specificity.

UlysseFG:
Steps that are not part of the study are listed. You should use study start and end to filter out the steps that should not appear.

frontend/src/components/shared/WithItemComponent.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const createWithItemComponent = <T extends FetchedObject>(withItem: WithItemFunc) => {
+	return (objectsByID: ItemsByID<T>, objectID: string, fn: ItemMappingFunc<T> = (item: T) => item, defaultValue: any = undefined) => {
+		return <WithItemComponent withItem={withItem} objectsByID={objectsByID} objectID={objectID} fn={fn} defaultValue={defaultValue} />
+	}
+}
+
+export const withSampleComponent = createWithItemComponent<Sample>(withSample)
+
+const TryWithSample = ({}) => {
```
UlysseFG, line 47: This looks like test code ... in the code.

frontend/src/components/studySamples/StudySamples.tsx
```diff
@@ -0,0 +1,51 @@
...
+	if (hideEmptySteps) {
+		renderedSteps = renderedSteps.filter(step => step.samples.length > 0)
+	}
+	return (
+		<>
+			<Collapse>
+				{renderedSteps.map(step => {
+					const hasSamples = step.samples.length > 0
+					return (
+						<Collapse.Panel key={step.stepID} header={step.stepName} extra={<Title level={5}>{step.samples.length}</Title>}>
```
UlysseFG, line 40: The sample count could be larger ... There is space... enlarge it a bit so it stands out more.

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
```diff
@@ -0,0 +1,172 @@
...
+import LIBRARY_COLUMNS, { ObjectWithLibrary } from './LibraryTableColumns'
+
+
+
+
+interface SampleAndLibrary extends ObjectWithSample, ObjectWithLibrary {}
+
+function getColumnsForStep(step: StudySampleStep, protocol: Protocol): ColumnsType<SampleAndLibrary> {
+
+	const DEFAULT_SAMPLE_COLUMNS = [
```
UlysseFG, line 16: The idea of default sample and library is not bad ... There is a problem with QC though since the listed sample will not have gone through QC yet... This is the columns i would add.
1. default sample
id
kind
name
container_barcode
coordinates
volume
conc.
qc_flag
creation_date
depleted

2. default library
id
kind
name
container_barcode
coordinates
index
volume
conc.
qc_flag
creation_date
depleted

3. experiment
id
name
container_barcode
coordinates
volume
# samples => nice to have but not required.
qc_flag
creation_date
depleted

Not sure we need anything else. Even then we could strip a couple columns most likely, the lab work page is where people will check for these informations.

frontend/src/models/protocols.ts
ckostiw: The protocol name is the only way for the frontend to know which protocol is which, so I created some constants for the names, in case the names change in the future and to avoid silly typo bugs.

UlysseFG: This is not ideal but yeah i see ...

frontend/src/components/shared/WithItemRenderComponent.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * WithItemRenderComponent is used where we need to load an object from the backend if it
+ * is not already in the redux store. It handles loading the object and when it is ready
+ * it passes the object to a render function that you provide. 
+ * 
+ * There specific WithXXXRenderComponent components for samples, containers, etc..
+ * 
+ * The component takes these props:
+ * 
+ * objectID: The ID of the object you want to render
+ * render: A function that takes the object as a parameter and returns a React.Element.=
```
UlysseFG, line 19: Something i wonder if you are trying to trigger my OCD ... = after . here. Missing . last sentence. etc ... a bit earlier ...

frontend/src/models/labwork_summary.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export interface LabworkSummaryStep {
+	name: string
+	count: number
+	specifications: LabworkSummaryStepSpecification[]
+}
+
+export interface LabworkSummaryStepSpecification extends FMSLabworkStepSpecification {}
+
+
```
UlysseFG, line 33: Let's keep max 2 blank lines.

frontend/src/components/studies/StudyDetails.tsx
```diff
@@ -58,6 +84,18 @@ const StudyDetails = ({studyId} : StudyDetailsProps) => {
                 <Descriptions.Item label="Project" span={2}>{project?.name ?? ''}</Descriptions.Item>
```
UlysseFG, line 84: Not very useful to display the project. Instead add the start and end steps of the study.

frontend/src/components/shared/WithItemRenderComponent.tsx
ckostiw: Sorry, I let ChatGPT write that comment for me.

UlysseFG (approved):

PR #478, opened 2023-01-26
Global hideEmptySteps flag for study details

ckostiw:
The study details page sample list has a switch to show or hide empty steps. Originally, this flag was a property of the study detail page. Now, the flag is stored in redux and shared by all study details view. It was irritating to have to keep flicking the switch separately for each study.


UlysseFG (approved):

ckostiw:
This is part of the labwork steps PR and doesn't need to be merged here.

PR #481, opened 2023-01-30
change normalization specifications

UlysseFG:


UlysseFG (commented):
Pushing this to save prevent to have to that as a more complex migration in next release,

PR #488, opened 2023-02-15
Wf labwork step page

ckostiw:
Labwork Step page - display samples at a specific step, allow users to select samples and then generate a prefilled template for the corresponding protocol.

### DIFF
--- a/frontend/src/components/App.js
+++ b/frontend/src/components/App.js
@@ -43,6 +43,8 @@ import {fetchInitialData, fetchSummariesData} from "../modules/shared/actions";
 import {logOut} from "../modules/auth/actions";
 import {get} from "../modules/users/actions";
 import DatasetsPage from "./datasets/DatasetsPage";
+import LabworkPage from "./labwork/LabworkPage";
+ 
 
 const { Title } = Typography;
 
@@ -109,12 +111,11 @@ const MENU_ITEMS = [
     icon: <ProjectOutlined />,
     text: "Projects",
   },
-  // DISABLED UNTIL THIS FEATURE IS READY TO SHIP
-  // {
-  //   url: "/lab-work",
-  //   icon: <ExperimentOutlined />,
-  //   text: "Lab Work"
-  // },
+  {
+    url: "/lab-work",
+    icon: <ExperimentOutlined />,
+    text: "Lab Work"
+  },
   {
     url: "/indices",
     icon: <BarcodeOutlined />,
@@ -263,12 +264,11 @@ const App = ({userID, usersByID, logOut, fetchInitialData, fetchSummariesData, g
                 <ProjectsPage />
               </PrivateNavigate>
             } />
-            {/* DISABLED UNTIL THIS FEATURE IS READY TO SHIP
-              <Route path="/lab-work/*" element={
+            <Route path="/lab-work/*" element={
               <PrivateNavigate>
                 <LabworkPage />
               </PrivateNavigate>
-            } /> */}
+            } />
             <Route path="/indices/*" element={
               <PrivateNavigate>
                 <IndicesPage />
--- a/frontend/src/selectors.ts
+++ b/frontend/src/selectors.ts
@@ -1,6 +1,6 @@
 import { Index, ItemsByID, Library, Project, Protocol, ReferenceGenome, Sample, SampleKind, Study, Taxon, User, Workflow } from "./models/frontend_models"
 import { LabworkSummaryState } from "./modules/labwork/reducers"
-import { StudySamplesState } from "./modules/studySamples/reducers"
+import { StudySamplesByID, StudySamplesState } from "./modules/studySamples/reducers"
 import { RootState } from "./store"
 
 /*
@@ -29,7 +29,8 @@ export const selectProtocolsByID = (state: RootState) => state.protocols.itemsBy
 export const selectReferenceGenomesByID = (state: RootState) => state.referenceGenomes.itemsByID as ItemsByID<ReferenceGenome>
 export const selectSamplesByID = (state: RootState) => state.samples.itemsByID as ItemsByID<Sample>
 export const selectSampleKindsByID = (state: RootState) => state.sampleKinds.itemsByID as ItemsByID<SampleKind>
-export const selectStudySamples = (state: RootState) => state.studySamples as StudySamplesState
+export const selectStudySamplesByID = (state: RootState) => state.studySamples.studySamplesById as StudySamplesByID
+export const selectHideEmptySteps = (state: RootState) => state.studySamples.hideEmptySteps
 export const selectStudiesByID = (state: RootState) => state.studies.itemsByID as ItemsByID<Study>
 export const selectTaxonsByID = (state: RootState) => state.taxons.itemsByID as ItemsByID<Taxon>
 export const selectUsersByID = (state: RootState) => state.users.itemsByID as ItemsByID<User>
--- a/frontend/src/components/labwork/overview/LabworkOverview.tsx
+++ b/frontend/src/components/labwork/overview/LabworkOverview.tsx
@@ -11,6 +11,7 @@ const { Title } = Typography
 
 const LabworkOverview = () => {
 	const [loading, setLoading] = useState(false)
+	const [showEmpty, setShowEmpty] = useState(true)
 	const labworkSummaryState = useAppSelector(selectLabworkSummaryState)
 	const dispatch = useAppDispatch()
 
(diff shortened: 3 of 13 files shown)


ckostiw:
I took a crack at forcing a default order on the samples in a prefilled template - sort by plate barcode first, then by coordinate. I send an `ordering` query parameter when requesting a prefilled template: `export const PREFILLED_TEMPLATE_DEFAULT_ORDERING = 'sample__container__barcode,sample__coordinates'`. 

It works, but sorting by coordinate is still alphabetic, so there is no way to sort by column rather than by row.

frontend/src/components/App.js
```diff
@@ -138,7 +143,7 @@ const colorStyle = {
 
 const titleStyle = {
   ...colorStyle,
-  width: "100%",
+  // width: "100%",
```
UlysseFG, line 146: Remove commented out stuff unless it has a documentation value.

frontend/src/components/labwork/overview/LabworkOverviewProtocolPanel.tsx
```diff
@@ -1,16 +1,31 @@
...
 
 export interface LabworkProtocolPanelProps {
 	readonly protocol: LabworkSummaryProtocol
+	readonly hideEmptySteps: boolean
 }
 
-const LabworkOverviewProtocolPanel = ({ protocol }: LabworkProtocolPanelProps) => {
+function doesGroupHaveSamples(group: LabworkStepGroup) {
+	// count the number of samples across the group's steps
+	const numSamples = group.steps.reduce((acc, step) => {
```
UlysseFG, line 12: Not the most efficient way to do that. You can bail once you find a single sample or deal with them as booleans.

frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+	function handlePageSize(pageSize: number) {
+		dispatch(setPageSize(pageSize))
+		dispatch(loadSamplesAtStep(step.id, stepSamples.pagedItems.page?.pageNumber ?? 1))
+	}
+
+	const pagination: PaginationParameters = {
+		pageNumber: stepSamples.pagedItems.page?.pageNumber ?? 1,
+		totalCount: stepSamples.pagedItems.totalCount,
+		pageSize: stepSamples.pagedItems.page?.limit ?? 10,
```
UlysseFG, line 161: Why not use DEFAULT_PAGINATION_LIMIT ?

frontend/src/components/shared/WorkflowSamplesTable/SampleTableColumns.tsx
```diff
@@ -0,0 +1,302 @@
...
+	INDIVIDUAL = 'INDIVIDUAL',
+	CONTAINER_NAME = 'CONTAINER_NAME',
+	CONTAINER_BARCODE = 'CONTAINER_BARCODE',
+	COORDINATES = 'COORDINATES',
+	VOLUME = 'VOLUME',
+	CONCENTRATION = 'CONCENTRATION',
+	QC_FLAG = 'QC_FLAG',
+	CREATION_DATE = 'CREATION_DATE',
+	DEPLETED = 'DEPLETED',
+	// IS_POOLED = 'IS_POOLED',
```
UlysseFG, line 45: Remove

frontend/src/components/shared/WorkflowSamplesTable/SampleTableColumns.tsx
```diff
@@ -0,0 +1,302 @@
...
+ * is being used to retrieve samples.
+ */
+
+export const UNDEFINED_FILTER_KEY = 'UNDEFINED_FILTER_KEY'
+
+export const SAMPLE_COLUMN_FILTERS: { [key in SampleColumnID]: FilterDescription } = {
+	// Object keys map to column "columnID" properties, to match columns to filters.
+	[SampleColumnID.ID]: {
+		type: FILTER_TYPE.INPUT_OBJECT_ID,
+		key: UNDEFINED_FILTER_KEY,
```
UlysseFG, line 207: Why not just undefined or null ?

frontend/src/modules/labwork/services.ts
```diff
@@ -0,0 +1,124 @@
...
+			const defaultGroup : LabworkStepGroup = {
+				defaultGroup: true,
+				steps
+			}
+			addGroupIfNotEmpty(protocol, defaultGroup)
+		}		
+
+		result.protocols.push(protocol)
+
+		// TODO : sort the protocols? By which criteria? Ideally it would match the basic
```
UlysseFG, line 89: Use redmine instead of the code for TODOs. These are likely to be there and ignored for posterity...

frontend/src/modules/labwork/services.ts
```diff
@@ -0,0 +1,124 @@
...
+		// order of steps in the workflows somehow.
+	}
+
+	return result
+}
+
+function getSpecifiedValue(step: LabworkSummaryStep, displayName: string) {
+	const spec = step.specifications.find(spec => spec.display_name === displayName)
```
UlysseFG, line 97: The idea behind the display name of the specifications was to add columns to the table on the fly using this as column name. I saw you decided to predefine columns in the frontend by protocol. Are these specification information really useful in that context ?

frontend/src/selectors.ts
```diff
@@ -19,19 +20,36 @@ import { RootState } from "./store"
...
+export const selectStudySamplesByID = (state: RootState) => state.studySamples.studySamplesById as StudySamplesByID
+export const selectHideEmptySteps = (state: RootState) => state.studySamples.hideEmptySteps
+export const selectStepsByID = (state: RootState) => state.steps.itemsByID
 export const selectStudiesByID = (state: RootState) => state.studies.itemsByID as ItemsByID<Study>
 export const selectTaxonsByID = (state: RootState) => state.taxons.itemsByID as ItemsByID<Taxon>
+export const selectToken = (state: RootState) => state.auth.tokens.access
 export const selectUsersByID = (state: RootState) => state.users.itemsByID as ItemsByID<User>
 export const selectWorkflowsByID = (state: RootState) => state.workflows.itemsByID as ItemsByID<Workflow>
 
+
```
UlysseFG, line 49: You can remove the extra empty lines ...

frontend/src/components/App.js
```diff
@@ -109,12 +115,11 @@ const MENU_ITEMS = [
...
-  // DISABLED UNTIL THIS FEATURE IS READY TO SHIP
-  // {
-  //   url: "/lab-work",
-  //   icon: <ExperimentOutlined />,
-  //   text: "Lab Work"
-  // },
+  {
+    url: "/lab-work",
+    icon: <ExperimentOutlined />,
+    text: "Lab Work"
```
UlysseFG, line 121: Given the new paradigm we entered, i would move Lab Work to the top of the menu and move Project just below.

UlysseFG:
The lack of refresh on the summary page when you submit a template is very noticeable. If you have time before we send to QC try to fix that. We'll have a ton of complaints for that it's clear.

UlysseFG:
The template dropdown for multiple choice prefilling feel a bit out of place. it would be better either next to the prefill template button (on its left) or bellow it (right aligned). Ideally there would be a small card around both the button and the dropdown to let people know this applies to that button.

frontend/src/components/shared/WorkflowSamplesTable/SampleTableColumns.tsx
ckostiw: I didn't want to allow undefined keys, and I decided to use that UNDEFINED_FILTER_KEY string in there to help diagnose missing key bugs (for instance, if that key shows up in redux as a filter property then it provides a good clue about what is going wrong).

UlysseFG: So TS made you do it ? That is a weak defense ;). I don't like the "coding for failure" strategy. It might help if your code does not work well, but your code should work well ... on its own. Coding for debugging tend to make bulkier code and it's more error prone since it does not have the fail fast quality of the undefined / null value. You can keep it if you like it, but you did not convince me ;). If undefined shows in redux you are still having an issue ;).

frontend/src/modules/labwork/services.ts
ckostiw: I would rather have some kind of identifier key in the step specification (eg. "LIBRARY_PLATFORM") than having to use the display_name field ("Library Platform") to determine the platform for a step, so that we can change the display names in the future without breaking any code (especially if we need to translate the app to another language).

UlysseFG: GeleHomme would indeed be nice ...

PR #491, opened 2023-02-16
Apply changes made on QC to workflows.

UlysseFG:


UlysseFG:
Need a last check for panel capture workflows.

ckostiw (approved):
That was quite a cleanup! :) Migrations ran fine and the frontend didn't have any problems after the changes.

PR #494, opened 2023-02-22
Step History endpoint.

UlysseFG:


ckostiw (approved):
I only have a question about this. Will there always be a process measurement associated with a Step? I was wondering if, for example, at some point the lab will want to include a Transfer step for some reason. Is there a process measurement associated with Transfer? If not then there would be no way to get the sample associated with the step history.

PR #495, opened 2023-02-22
Replace coordinates by model instances.

UlysseFG:


ckostiw (approved):

ckostiw:
Sorting by coordinate in the Libraries table is not working - switching between ascending and descending order has no effect. This is the URL sent to the libraries viewset:

`http://localhost:9000/api/libraries/?limit=20&offset=0&ordering=coordinate__name`

ckostiw:
sample-next-step is broken. For the selected samples, the frontend always asks for the samples to be returned sorted by container and coordinate.


image

ckostiw:
The killer feature I most want is the ability to ask sample-next-step to return samples sorted by row/column or column/row...

UlysseFG:
> Sorting by coordinate in the Libraries table is not working - switching between ascending and descending order has no effect. This is the URL sent to the libraries viewset:
> 
> `http://localhost:9000/api/libraries/?limit=20&offset=0&ordering=coordinate__name`

Can't seem to be able to reproduce ... the ordering seem to be working as intended ...

UlysseFG:
> The killer feature I most want is the ability to ask sample-next-step to return samples sorted by row/column or column/row. I'd like to add a switch in the selected samples table to switch between those two modes.

I'll set it by column for now ... we can add more "finesse" to it in a future update...

ckostiw:
> 

Doh, you're right. I'm not sure what happened.

ckostiw:
> > Sorting by coordinate in the Libraries table is not working - switching between ascending and descending order has no effect. This is the URL sent to the libraries viewset:
> > `http://localhost:9000/api/libraries/?limit=20&offset=0&ordering=coordinate__name`
> 
> Can't seem to be able to reproduce ... the ordering seem to be working as intended ...

Yeah, now it's working for me too. I don't know what happened.

frontend/src/selectors.ts
```diff
@@ -47,3 +47,4 @@ export const selectTaxonsByID = (state: RootState) => state.taxons.itemsByID as
 export const selectToken = (state: RootState) => state.auth.tokens.access
 export const selectUsersByID = (state: RootState) => state.users.itemsByID as ItemsByID<User>
 export const selectWorkflowsByID = (state: RootState) => state.workflows.itemsByID as ItemsByID<Workflow>
+export const selectCoordinatesByID = (state: RootState) => state.coordinates.itemsByID as ItemsByID<Coordinate>
```
ckostiw, line 50: I'm trying to keep these in alphabetical order (although I see I missed one :) )

frontend/src/models/fms_api_models.ts
```diff
@@ -309,8 +309,13 @@ export interface FMSWorkflow extends FMSTrackedModel {
 export interface WorkflowStep {         // Not a tracked model - just a simple serialized object
     id: FMSId                           // Step Order ID
     order: number                       // Step order value
-    step_id : FMSId                     // Step ID
+    step_id: FMSId                     // Step ID
     step_name: string                   // Step name
     protocol_id:    FMSId               // ID of protocol associated with step
 }
 
+export interface FMSCoordinate extends FMSTrackedModel {
```
ckostiw, line 317: I'd like the models to stay in alphabetical order

frontend/src/components/containers/ContainerHierarchy.js
```diff
@@ -86,8 +86,8 @@ const renderSample = (sample, sampleKind) => {
          }
          {' '}
         <b>{sample.name}</b> sample ({sampleKind}){' '}
-        {sample.coordinates &&
-          `@ ${sample.coordinates}`
+        {sample.coordinate &&
+          `@ ${withCoordinate(coordinatesByID, sample.coordinate, coordinate => coordinate.name, "Loading...")}`
```
ckostiw, line 90: In the future, please use WithXXXRenderComponent. I'm planning on refactoring all of the old-style `withXXX` calls in components to use the render component.

frontend/src/components/shared/WithItemComponent.tsx
```diff
@@ -78,6 +78,7 @@ const createWithItemComponent = <T extends FMSTrackedModel>(withItem: WithItemFu
 	return WithItemComponentParent
 }
 
+export const withCoordinateComponent = createWithItemComponent<Coordinate>(withCoordinate)
```
ckostiw, line 81: Just FYI, I will probably remove this file from the project. The idea was to create a component with props that match the 'withItem' functions to make refactoring easier, but I think I'll just use WithItemRenderComponent everywhere.

ckostiw (changes requested):
Just minor stuff

ckostiw:
> > > Sorting by coordinate in the Libraries table is not working - switching between ascending and descending order has no effect. This is the URL sent to the libraries viewset:
> > > `http://localhost:9000/api/libraries/?limit=20&offset=0&ordering=coordinate__name`
> > 
> > 
> > Can't seem to be able to reproduce ... the ordering seem to be working as intended ...
> 
> Yeah, now it's working for me too. I don't know what happened.

I think we need to add an `ordering_fields` property to `ContainerViewSet` for coordinate sorting to work:
```
ordering_fields = {
        *_list_keys(_container_filterset_fields),
    }
```

frontend/src/components/samples/SampleEditContent.js
```diff
@@ -29,6 +29,9 @@ import { requiredRules, nameRules } from "../../constants";
 const searchSamples = (token, input) =>
   withToken(token, api.samples.search)(input).then(res => res.data.results)
 
+const searchCoordinates = (token, input, options) =>
```
ckostiw, line 32: There are a couple of issues with this.
1. If you edit an existing sample, the Select element shows the coordinate ID when the form is first displayed. It only switches to showing a coordinate name if the user clicks on the field to initialize the coordinate options.
2. `options` is undefined and the api defaults to returning 100 coordinates, from A00 to D22. This doesn't seem to match the coordinate choices for a 96 well plate or a 384 well plate.

* I think the select options should be based on the container type.
* We should initialize the coordinate options when the component is first rendered, rather than on each `onSearch` call. 
* I would disable the coordinates field if the container type hasn't been selected yet.

ckostiw (approved):

PR #496, opened 2023-02-23
Additional info sample submission

UlysseFG:


ckostiw (commented):
I suggest something like this: 
```


This template is used both internally by the lab and by external clients. Certain fields are only used by the lab.

Sample Submission

Samples are submitted by entering sample information in the SampleSubmission sheet. External clients can submit basic information about the samples. For submissions made by the lab, all required columns marked by an asterisk must be filled.

Library Submission

Columns related to libraries are greyed out by default. To submit a library you must choose a library type in the Library Type column. The remaining library columns will then become available. All columns are required, except for Selection and Selection Target.

Individual Information

You are not required to provide individual information. 

If you do need to submit information about individuals then the Individual ID, NCBI Taxon ID# and Sex are required.
The individual ID can contain any value. If the same ID is already used in the system then the lab will move the ID to the Individual Alias column and will generate a unique ID for the individual, for internal use.


Pool Submission

To submit pooled samples you must enter the pool information in the PoolSubmission sheet, one row per pool submitted. You then enter information about the pooled samples in the SampleSubmission sheet, and link the samples to the pool by selecting the pool in the Pool Name column of the SampleSubmission sheet.

The container and coordinate information of the pooled sample is specified in the PoolSubmission sheet. It is not necessary to specify the container and coordinate information for the individual samples contained in the pool.

The pool sample volume is set to sum of the invidual sample volumes contained in the pool.
```

ckostiw:
Also, I think if some columns are optional then we need to specify which columns the client must fill out and which they can leave blank.

ckostiw (changes requested):
In the sample submission form, the Index dropdown is broken. The Index Set drop down works, but not Index.
Same for Selection Target - the dropdown is not opening.

ckostiw:
Tested biosample submission, library submission and pool submission, with and without concentration - looks good.

ckostiw:
All tests pass.

ckostiw (approved):

PR #498, opened 2023-02-23
Force depleted when volume of a sample is set to 0.

UlysseFG:


ckostiw (approved):
I wonder if the lab would prefer us to set the sample as depleted if the volume is less that a certain amount, such as 1uL? I suspect that they will ask for that eventually...

UlysseFG:
> I wonder if the lab would prefer us to set the sample as depleted if the volume is less that a certain amount, such as 1uL? I suspect that they will ask for that eventually...

Yeah i am pretty sure this change will not help the lab despite what they told us. Any threshold will cause issues too they sill expect to hit the threshold and be off by 0.001 and they will have to check all their submission to make sure they hit 0. Just so they do not have to click that one time ;).

PR #499, opened 2023-02-23
Add multiple study from sample submission.

UlysseFG:


ckostiw (changes requested):
The sample submission template needs to be updated - data validation on the column blocks the user from entering multiple studies, eg. "A-B-C"

ckostiw:
Strange. If I run this backend with my current frontend, the frontend shows the samples at random steps (extraction, library qc, experiment run, etc..). I'll need to merge master into my branch, I think, and solve some issues. :)

UlysseFG:
Ok added some validation to make sure we don't go queuing sample to library steps and removed the Study dropdown.

ckostiw:
To test, I tried submitting a library to two studies - once that starts on step Library QC, and the other that starts with Extraction DNA. The error was caught and I got this message: 
`Row #9: queue_to_study - Step Extraction (DNA) expected Unextracted sample but Sample LIB1 does not match that sample type.`

I think it would be useful to include the study letter in the message, to indicate which study could not accept the sample.

PR #501, opened 2023-03-14
Sort protocols by workflow order

ckostiw:
Display protocols in the labwork overview page in a sensible order - the order in which they normally appear in workflows.

The idea is to look at each protocol and see which protocols it should follow and which it should precede based on where it appears in the workflows.

This gets done once, the first time the labwork is fetched from the backend, after the app is initialized and all the workflow definitions have been loaded.

I also tweaked the labwork-info endpoint in sample_next_step to avoid returning info about steps that are not used in any workflow - I was getting an Infinium protocol that is not used in any workflow and always had zero samples.

### DIFF
--- a/backend/fms_core/viewsets/sample_next_step.py
+++ b/backend/fms_core/viewsets/sample_next_step.py
@@ -9,7 +9,7 @@ from fms_core.filters import SampleNextStepFilter
 
 from ._utils import TemplateActionsMixin, TemplatePrefillsLabWorkMixin, _list_keys
 from ._constants import _sample_next_step_filterset_fields
-from fms_core.models import SampleNextStep, StepSpecification, Protocol, Step
+from fms_core.models import SampleNextStep, StepSpecification, Protocol, Step, Workflow
 from fms_core.serializers import SampleNextStepSerializer, StepSpecificationSerializer
 from fms_core.templates import (SAMPLE_EXTRACTION_TEMPLATE, SAMPLE_QC_TEMPLATE, NORMALIZATION_PLANNING_TEMPLATE, NORMALIZATION_TEMPLATE,
                                 LIBRARY_PREPARATION_TEMPLATE, LIBRARY_QC_TEMPLATE, SAMPLE_POOLING_TEMPLATE, LIBRARY_CAPTURE_TEMPLATE,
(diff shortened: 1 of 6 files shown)


backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -170,6 +170,12 @@ def labwork_info(self, request, *args, **kwargs):
             if protocol.id not in sample_next_step_by_protocol.keys():
 
                 protocolSteps = Step.objects.filter(protocol=protocol)
+
+                # Make sure that at least one workflow uses one of the steps for this protocol, since
+                # labwork doesn't need protocols that are not used by any workflow (Infinium...)
+                workflows = Workflow.objects.filter(steps__in=protocolSteps)
```
UlysseFG, line 176: `if Workflow.objects.filter(steps__in=protocolSteps).exists():` should be lighter and clearer.

frontend/src/modules/labwork/services.ts
```diff
@@ -154,4 +173,99 @@ export function findChangedStepsInSummary(oldSummary: LabworkSummary, newSummary
...
+				const distanceA = entriesA.reduce((acc, entry) => {
+					return acc + (entry.stepOrder / entry.numSteps)
+				}, 0) / entriesA.length
+				const distanceB = entriesB.reduce((acc, entry) => {
+					return acc + (entry.stepOrder / entry.numSteps)
+				}, 0) / entriesB.length
+
+				return distanceA - distanceB
+			} else {
+				return numAfter - numBefore
```
UlysseFG, line 261: It's likely going to give wonky results as we are not using a common metric for all steps... Both metrics are not on the same scale... This seem to be working for now but it's likely to need rework down the line. Let's roll with it, but ordering usually use only a single criterion (or a weighted combination).

backend/fms_core/viewsets/sample_next_step.py
ckostiw: Excellent - fixed. 👍

frontend/src/modules/labwork/services.ts
ckostiw: Yes, I'm sure that there is a better algorithm - this is a bit hacky, I agree, but it got a decent result for now.

UlysseFG (approved):

PR #505, opened 2023-03-17
Generate filter options dynamically

ckostiw:
The sample KIND column filter needs to be initialized from the sample kinds contained in the redux store - we have to generate a list of options for the Select element. This change adds the ability to define a function to generate the options dynamically.

It also cleans up some excessive rendering in the LabworkStep component by making proper use of `useCallback` and `useMemo`.

Finally, I fixed a bug in the filter serializer for SELECT values.

### DIFF
--- a/frontend/src/components/labwork/step/LabworkStep.tsx
+++ b/frontend/src/components/labwork/step/LabworkStep.tsx
@@ -1,6 +1,6 @@
 import { InfoCircleOutlined, SyncOutlined } from '@ant-design/icons'
 import { Alert, Button, Select, Space, Tabs, Typography } from 'antd'
-import React, { useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 import { useAppDispatch } from '../../../hooks'
 import { FMSId } from '../../../models/fms_api_models'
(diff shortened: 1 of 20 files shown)


frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -43,55 +43,64 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
 		}
 	}, [stepSamples])
 
-	function handleSetFilter(filterKey: string, value: FilterValue, description: FilterDescription) {
-		if(typeof description === 'undefined') {
-			return
-		}
-		dispatch(setFilter(step.id, description, value))
-	}
+	const handleSetFilter = useCallback(
```
UlysseFG, line 46: I'm not convinced by all these additional useCallback. My understanding is you save during rendering by reusing the function but i don't really see any gain to use it on such small functions. As far as i can see you cannot reuse the values or the results so in the end you are most likely just adding test to the function with a risk of having the wrong function if the cache is not refreshed when it should. The fact that you do not set dependencies will likely lead to the wrong values being used in the function.

ckostiw: You're absolutely right about the dependencies - there were some that I needed to add. I added the react-hooks linter plugin to check for missing dependencies, to avoid that in the future, and I added all the missing dependencies.

I do want to continue using these hooks and other react best practices though, and to consider rendering performance a bit more.

PR #507, opened 2023-03-20
Return more informative error when extraction type field is omitted.

UlysseFG:


ckostiw (commented):
Screen Shot 2023-03-28 at 1 34 20 PM

Wouldn't it be better if the Association error appears on a Row# line, rather than at the top? Also, I only have 2 bad rows but I got 4 messages. 
Also, unrelated to this PR, it would be better if the Row# error messages mention the bad column. The error message asks for a "Sample kind" but the column name is "Extraction Type".

UlysseFG:
I could patch something for the repeats. The row errors are already there (however cryptic they may be. They are actually not so bad here). The extra errors at the top is generated when i do a scan over the whole dataset outside the row_handler (which does the errors reporting by row). The error is caught both by the service at the importer level and the row_handler in this case. The duplication is due to us propagating our error like good soldiers ... we could dig into it at some point (there is some issues related to errors that do not get caught at row_handler level).

UlysseFG:
Removed the duplication but this is just a partial fix ... I'll add a ticket to add a systematic required field check upon template submission.

PR #513, opened 2023-03-30
Add a transfer step in workflows before lib QC

UlysseFG:


ckostiw (approved):
Tested it out and everything worked as expected.

PR #515, opened 2023-04-04
Study samples filtering and sorting

ckostiw:
Adds filters and sorting to queued samples in the study details page. Each step in the study has it's own set of filters and sort order. Filtering and sorting is not supported in the completed samples tables yet.

* Added a redux state for each step to store filters and sort order.
* The same redux state is used to keep track of the collapsed/expanded state of the step, and the choice between "Ready to Process" or "Completed" samples, so make navigating back to the study less irritating.
* Added a new `summary` endpoint to `sample-next-step-by-study` and `step-history` endpoints to get the total numbers of queued and completed samples (before I was using the number of samples returned, but now that varies depending on filters).
* Used `immer` to write the studySamples reducer, which makes the code much more maintainable.


backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -36,4 +43,25 @@ def destroy(self, request, pk=None):
...
+    study__id__in: The study ID
+
+    Returns:
+    Dictionary of step order ID / sample count pairs, one for each step in the study workflow.
+    """
+    @action(detail=False, methods=["get"])    
+    def summary(self, request):
+        study_id = request.GET.get('study__id__in')
+        study = Study.objects.get(pk=study_id)
+        step_orders = StepOrder.objects.filter(workflow=study.workflow)
```
UlysseFG, line 62: you might want to order them by order ...

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -36,4 +43,25 @@ def destroy(self, request, pk=None):
...
+    the key is a step order ID and the value is the count of the number of samples queued at that step.
+
+    Args:
+    study__id__in: The study ID
+
+    Returns:
+    Dictionary of step order ID / sample count pairs, one for each step in the study workflow.
+    """
+    @action(detail=False, methods=["get"])    
+    def summary(self, request):
```
UlysseFG, line 59: I would rename these endpoints  summary_by_study since they only perform a summary by study. It would be better to have an endpoint that does a full summary (by study) but can optionally take a study_id to filter it to return a single study.

ckostiw: Renamed

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -36,4 +44,51 @@ def destroy(self, request, pk=None):
...
+        The endpoint returns an array of objects. Each object contains a study ID and a list
+        of steps, where each step includes a count of the number of samples queued.
+
+        For each step, the step order ID, order, step name and count is returned.
+
+        The endpoint only returns steps that have at least one sample queued - steps
+        with zero samples are omitted from the results.
+
+        Args:
+        study__id__in: The study ID (optional)
```
UlysseFG, line 65: This is not really an Arg, i would add it as a note in the above description so the auto-doc does not include it as an argument in the docs.

UlysseFG (approved):

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -36,4 +44,51 @@ def destroy(self, request, pk=None):
...
+
+        Returns:
+        Dictionary of step order ID / sample count pairs, one for each step in the study workflow.
+        """
+       
+        study_id = request.GET.get('study__id__in')
+    
+        samplesInStudy = SampleNextStepByStudy.objects.all()
```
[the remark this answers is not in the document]
ckostiw: Doh!

PR #517, opened 2023-04-12
Displaying step order number when showing workflow steps

JMarce01:


ckostiw (changes requested):
I think we should drop the ':' colon and make the number bolder.

ckostiw:
Screen Shot 2023-04-13 at 9 13 33 AM

ckostiw (changes requested):

ckostiw (approved):

ckostiw (approved):

PR #518, opened 2023-04-12
Moving static data fetch into new function

JMarce01:
Splitting up data fetches in "App.js" fetching static and dynamic data separately.

### DIFF
--- a/frontend/src/components/App.js
+++ b/frontend/src/components/App.js
@@ -34,12 +34,8 @@ import useUserInputExpiration from "../utils/useUserInputExpiration";
 import { useAppDispatch, useAppSelector } from "../hooks";
 import { setAppInitialized } from "../modules/app/actions";
 import { logOut } from "../modules/auth/actions";
-<<<<<<< HEAD
-import { fetchInitialData, fetchLabworkSummary, fetchSummariesData } from "../modules/shared/actions";
-=======
 import { refreshLabwork } from "../modules/labwork/actions";
 import { fetchInitialData, fetchSummariesData, fetchStaticData } from "../modules/shared/actions";
->>>>>>> 2218b284 (moving static data fetch into new function)
 import { get } from "../modules/users/actions";
 import { selectAppInitialzed, } from "../selectors";
 import DatasetsPage from "./datasets/DatasetsPage";
--- a/frontend/src/components/App.js
+++ b/frontend/src/components/App.js
@@ -34,10 +34,9 @@ import useUserInputExpiration from "../utils/useUserInputExpiration";
 import { useAppDispatch, useAppSelector } from "../hooks";
 import { setAppInitialized } from "../modules/app/actions";
 import { logOut } from "../modules/auth/actions";
-import { refreshLabwork } from "../modules/labwork/actions";
-import { fetchInitialData, fetchSummariesData, fetchStaticData } from "../modules/shared/actions";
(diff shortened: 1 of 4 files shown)


ckostiw:
I found an issue. The `fetchInitalData` action is dispatched by the `performAuth` auth action after the user logs in. The static data no longer gets loaded, since we removed that from `fetchInitialData`.

`fetchInitialData` is repeating the same things that are done in the `fetchSummariesData` and `fetchListedData`, and `fetchListedData` is repeating stuff done in `fetchSummariesData`, so maintenance will be a bit harder. We should clean that up to have a set of actions that each does one thing.

Here's what I propose:

1. Change `fetchListedData` so that it only handles the "list" stuff.
2. In `TemplateFlow.js`, update `onSubmit` to dispatch fetchSummariesData after it dispatches fetchListedData.
3. In `performAuth`, remove the dispatch of the `fetchInitialData` action - the app should take care of that, not the auth action.
4. Use the auth state in `App.js`, to detect when the auth token changes. Add that as a dependency to the useEffect() so that the useEffect() is called whenever auth changes. That way the initial data will be loaded after the user logs in.
5. Delete the fetchInitialData function
6. In `App.js` change the useEffect clause to dispatch `fetchSummariesData` and `fetchListedData` instead of calling `fetchInitalData`

We should end up with a clean set of `fetch...Data` actions without code duplication.

frontend/src/modules/auth/actions.js
```diff
@@ -1,32 +1,29 @@
 import jwtDecode from "jwt-decode";
 
 import api from "../../utils/api";
-import {createNetworkActionTypes, networkAction} from "../../utils/actions";
-import {fetchInitialData} from "../shared/actions";
-import {constVal} from "../../utils/functions";
+import { createNetworkActionTypes, networkAction } from "../../utils/actions";
+import { fetchInitialData } from "../shared/actions";
```
ckostiw, line 5: You have a dead import here. Your linter should be have picked that up. You can run the vscode command 'organize imports' to remove dead imports automatically on an open source file (cmd-shift-P opens the commands chooser in vscode).

frontend/src/modules/shared/actions.js
```diff
@@ -124,6 +103,8 @@ export const fetchListedData = () => async (dispatch, getState) => {
         Projects.listTable,
         ProcessMeasurements.listTable,
         Protocols.list,
+        Datasets.listTable,
+        DatasetFiles.listTable,
     ].map(a => dispatch(a())))
 
     // Lower priority - Fetch summaries
```
ckostiw, line 110: fetchListedData should not be loading summaries - that should only be done in fetchSummariesData. You should remove this and then check everywhere that `fetchListedData` is used to add a call to fetchSummariesData (ie. TemplateFlow.js and App.js)

frontend/src/components/App.js
```diff
@@ -152,13 +152,15 @@ const App = ({userID, usersByID, logOut, get}) => {
 
   const dispatch = useAppDispatch()
   const isInitialized = useAppSelector(selectAppInitialzed)
-
+  const token = useAppSelector(selectAuthTokenAccess)
```
ckostiw, line 155: 👍

ckostiw (changes requested):
Getting close :)

ckostiw (approved):
Looks great

PR #519, opened 2023-04-12
Fixing SampleEditForm bug

JMarce01:
- replaced "formData" with "form" using useForm to keep track of all changed data
- changed onValueChange to listen for sample_kind change to set the tissue_source to empty if extracted
- changed serialize to serializeFormData to format the Form information properly (could be better optimized/cleaner but some values need to be formatted either empty string because values can't be null)


ckostiw:
There seems to be a bug. When I select a plate as a container, the coordinates field should be enabled, but it remains greyed out.

Screen Shot 2023-04-13 at 4 17 45 PM

ckostiw:
The sample form is used for creating new samples, but it is also used for editing existing samples. When I tried to edit an existing sample I got an empty form, rather than a form filled with the current sample values.


https://user-images.githubusercontent.com/104078019/232058573-41aa21ba-149e-476f-978e-a0e2b59b6980.mov

ckostiw (changes requested):
I've reported two bugs

ckostiw (changes requested):

ckostiw (changes requested):
When I edit an existing sample, the coordinate field is disabled even if the container type is set and there's already a coordinate value. I can't update the sample coordinate.

Screen Shot 2023-04-17 at 4 07 03 PM

ckostiw (changes requested):
If I open an existing sample for editing, and the sample kind is RNA or DNA (ie. it's extractable) then the tissue type field should be enabled. Currently, it's disabled and I have to switch the sample kind to force it to become enabled.
Screen Shot 2023-04-18 at 9 14 44 AM

ckostiw (changes requested):
If I am adding a new sample, as soon as I choose a coordinate the coordinate field becomes disabled and I can no longer edit it.
Screen Shot 2023-04-18 at 9 24 11 AM

frontend/src/components/samples/SampleEditContent.js
```diff
@@ -158,14 +155,30 @@ const SampleEditContent = ({ token, samplesByID, sampleKinds, add, update, listT
...
+        form.setFieldValue('tissue_source', '')
+        setisTissueEnabled(false)
+      } else {
+        setisTissueEnabled(true)
+      }
+    }
+    if (values["container"]) {
+      setIsCoordRequired(true)
+    } else {
+      setIsCoordRequired(false)
```
ckostiw, line 176: The coordinates field gets disabled here every time a value other than 'container' changes.

ckostiw (changes requested):

frontend/src/components/samples/SampleEditContent.js
```diff
@@ -158,14 +155,30 @@ const SampleEditContent = ({ token, samplesByID, sampleKinds, add, update, listT
     onSearchContainer(newData.container, { exact_match: true })
     onSearchCoordinate(newData.coordinate, { exact_match: true })
     onSearchSampleKind(newData.sample_kind)
+    if (!isAdding)
```
ckostiw, line 158: I'm not sure why it makes a difference whether we are adding a new sample or editing an existing sample. Couldn't we call `form.setFieldsValue` either way? `newData` will have whatever fields it got from EMPTY_SAMPLE. Does that cause an error?

ckostiw:
Still a bug. If I try to edit an existing sample, which is in a plate at coordinate A01, the coordinates field is disabled, so if I wanted to change the coordinates I would not be able to. If I try to change anything else (such as the sample kind), when I submit I get an error saying that the coordinate is required, and only then does the field become enabled.

https://user-images.githubusercontent.com/104078019/234326747-59fca0f0-6401-49ca-94f0-45e99184d088.mov

ckostiw:
Also, if I submit a change and there is an error, the form gets reset and I lose all the changes I had made. This is likely to frustrate the users and possible lead to errors.

PR #525, opened 2023-04-17
Display samples removed from study

ckostiw:
Display samples that are removed from a study in the study view, using the step history to identify those samples.

### DIFF
--- a/frontend/src/components/studySamples/StudySamples.tsx
+++ b/frontend/src/components/studySamples/StudySamples.tsx
@@ -118,7 +118,7 @@ function StepPanel({step, studyID, uxSettings} : StepPanelProps) {
 	const removedSamples = step.completed.filter(completed => completed.removedFromWorkflow === true)
 	const hasRemovedSamples = removedSamples.length > 0
 
-	const removedTitle = removedSamples.length === 1 ? `1 sample was removed from study` : `${removedSamples.length} samples were removed from study`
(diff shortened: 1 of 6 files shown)


frontend/src/components/studySamples/StudySamples.tsx
```diff
@@ -96,64 +95,81 @@ function StudySamples({ studyID, studySamples, refreshSamples }: StudySamplesPro
...
 	const dispatch = useAppDispatch()
 
+	const countString = `${step.completedCount} / ${step.sampleCount + step.completedCount}`
+	const countTitle = `${step.completedCount} of ${step.sampleCount + step.completedCount} samples are completed`
+	
+	const completedSamples = step.completed.filter(completed => completed.removedFromWorkflow === false)
+	const removedSamples = step.completed.filter(completed => completed.removedFromWorkflow === true)
+	const hasRemovedSamples = removedSamples.length > 0
+
+	const removedTitle = removedSamples.length === 1 ? `1 sample was removed from study` : `${removedSamples.length} samples were removed from study`
```
UlysseFG, line 121: I think we should specify in this message that the sample was removed after the step was completed. It is to make sure it is clear the sample is not missing from this step listing but will from further study steps.

UlysseFG (approved):

frontend/src/components/studySamples/StudySamples.tsx
ckostiw: I added "at this step" to make it a bit more clear. We could also add an info message in the removed samples section to make it even more clear "These samples were removed from the study at this step and will not appear in any later steps."

PR #527, opened 2023-04-17
Addition of Container Comment column to Container Creation Template.

UlysseFG:


ckostiw (approved):
Works as expected. My only comment would be that the template column title "Container Comment" is inconsistent with the comment columns in our other templates, which are all named "Comment". It might be better to just use "Comment", like we do in the sample submission form for the sample comment.

UlysseFG:
Yeah i was uncertain if the Comment alone would be understood as an action operation but there is really no operation related to the container creation. I'll rename it (i named it Comment originally).

PR #528, opened 2023-04-18
Fix unique key errors in list pages

ckostiw:
Fixes a react 'unique key' error in  pages that include the actions dropdown and prefilled templates dropdown. This is to reduce noise in the console.

I converted the functions that generate these dropdowns to proper React components so that I could assign a key to the dropdown elements where they are used in page headers.

~~There shouldn't be any change in the existing behaviour of the dropdowns.~~

The template actions and prefilled template menu were tweaked so that menu item buttons span the entire width of the menu (except for ant padding).


UlysseFG:
This is mildly unrelated to the PR content but i would be nice tweak to the dropdown. If the button could be the fullwith of the dropdowns and aligned to the left it would prevent clicking on the area not covered by the button (especially for button with short text) and having the dropdown close without selected the intended option,

PR #532, opened 2023-04-24
Filters update

JMarce01:
- adding filter warning and clear filter to workflow steps in Labwork section
- adding filter warning and clear filter to study sample steps


frontend/src/components/shared/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -1,53 +1,61 @@
...
-	filterDefinitions?: FilterDescriptionSet
-	filterKeys?: FilterKeySet
-	filters?: FilterSet
-	setFilter?: SetFilterFunc
-	setFilterOptions?: SetFilterOptionFunc
-	sortBy?: SortBy
-	setSortBy?: SetSortByFunc
-	pagination?: PaginationParameters
+	hasFilter: boolean,
+	stepNumber: number,
```
ckostiw, line 27: `stepNumber` doesn't seem to be used by WorkflowSamplesTable, so this prop should be removed. Also, the samples table should remain as generic as possible, so adding a stepNumber prop isn't a good idea.

frontend/src/components/shared/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -1,53 +1,61 @@
...
+	clearFilters?: any,
```
ckostiw, line 28: Please define a type for the clearFilters function - `any` isn't a good idea.

frontend/src/components/filters/FiltersWarningTS.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { QuestionCircleOutlined } from "@ant-design/icons";
+import { Tooltip, Typography } from "antd";
+import FiltersInfo from "./FiltersInfoTS";
+const { Text } = Typography
+
+interface FiltersWarning {
+    filters: any,
```
ckostiw, line 8: There is a type defined for filters, so you should use that, not any. See `FilterSet` in `paged_items.ts`.

frontend/src/components/shared/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -84,34 +92,51 @@ function WorkflowSamplesTable({sampleIDs, columns, filterDefinitions, filterKeys
...
 	}
-	
+
 	return (
 		<>
-			{tableColumns && 
+			{tableColumns &&
 				<>
+					{
+						hasFilter &&
```
ckostiw, line 114: Let's create a single stand-alone component that combines the div, FiltersWarning and Button and call it `FiltersBar` (I can't think of a better name that doesn't already conflict with the existing components :) ). 

I would also like to move the filters bar out of WorkflowSamplesTable. I think the filters bar can exist independently of the samples table. You would create a `<FiltersBar>` component in the parent component that is using the samples table (eg. LabworkStep).

frontend/src/modules/studySamples/reducers.ts
```diff
@@ -184,6 +184,14 @@ export const studySamplesReducer = (state: WritableDraft<StudySamplesState>, act
 			}
 			break
 		}
+		case CLEAR_FILTERS: {
```
ckostiw, line 187: 👍

frontend/src/modules/studySamples/actions.ts
```diff
@@ -78,6 +78,18 @@ export function initStudySamplesSettings(studyID: FMSId, stepIDs: FMSId[]) {
 	}
 }
 
+export function clearFilters(studyID: FMSId, stepID: FMSId) {
```
ckostiw, line 81: 👍

ckostiw (changes requested):

frontend/src/components/filters/FiltersWarningTS.tsx
```diff
@@ -0,0 +1,31 @@
...
+    nFilters: number
+}
+
+const FiltersWarning = ({ filters, nFilters }: FiltersWarning) => {
+    if (nFilters === 0)
```
ckostiw, line 13: Let's implement the FiltersBar component (see below) and handle this nFilters stuff internally so the caller doesn't have to mess around counting the number of filters for this component.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,81 @@
+import React from "react";
+import { FILTER_TYPE } from "../../constants";
+
+
+const FiltersInfos = ({
+    filters,
```
ckostiw, line 6: Please define the type for `filters`

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    <strong>{description.label}: </strong>
+                }
+            </>
+        );
+        const valuesArray = [].concat(filterValue)
+        let labels: any = [];
+        let value = ""
+        switch (description.type) {
+            case FILTER_TYPE.SELECT:
```
ckostiw, line 24: I get a big linter error here `Unexpected lexical declaration in case block`. It is complaining about declaring local variables in the case statement. The fix is to put the case handling code in `{}` brackets, eg.
```
case THING: {
  const someVar = ...
  break;
}
```

Screen Shot 2023-04-25 at 2 07 13 PM

If you're not getting that error I can help you get the linter set up.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,81 @@
...
+                labels = valuesArray.map((val: any) => {
+                    if (description.options) {
+                        const option = description.options.find((option: { value: any }) => option.value === val)
+                        valueJSX = option.label
```
ckostiw, line 28: Since this is an array mapping function, it should be returning a value, not setting a variable. This is why the filter strings for Sample Kind and Depleted aren't showing up. The `labels` variable should end up holding an array of value strings, but right now it is just empty.

ckostiw (changes requested):

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+interface FiltersInfosProps {
+    filters: FilterSet
+}
+
+const FiltersInfos = ({ filters }: FiltersInfosProps) => {
+    const appliedFilters = Object.keys(filters).filter((key: any) => filters[key]?.value)
+    const getValue = (key: any) => {
+        const filterValue: FilterValue = filters[key].value ?? {}
+        const description: any = filters[key].description
```
ckostiw, line 13: The type here is `FilterDescription | undefined`. You'll need to add a null check somewhere to handle the case where the description is undefined.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
...
+        let valueJSX: any;
```
ckostiw, line 14: The type here should be `string`

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
...
+        const descriptionJSX: any = (
```
ckostiw, line 15: You can leave the type off here, as it can be inferred by the compiler to be JSX.Element.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
...
+            </>
+        );
+        const valuesArray = [filterValue]
+        let labels: any = [];
+        let value = ""
+        switch (description.type) {
+            case FILTER_TYPE.SELECT: {
+                labels = valuesArray.map((val: any) => {
+                    if (description.options) {
+                        const option = description.options.find((option: { value: any }) => option.value === val)
```
ckostiw, line 29: Freezeman crashed (white screen of death) when I selected a Sample Kind filter (`BLOOD`) and then hovered the mouse over the filter info.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
+import React from "react";
+import { FILTER_TYPE } from "../../constants";
+import { FilterSet, FilterValue, RangeFilterValue } from "../../models/paged_items";
+
+interface FiltersInfosProps {
+    filters: FilterSet
+}
+
+const FiltersInfos = ({ filters }: FiltersInfosProps) => {
+    const appliedFilters = Object.keys(filters).filter((key: any) => filters[key]?.value)
```
ckostiw, line 10: There is no need for this `any` typing on `key` - the compiler knows what type to expect from Object.keys

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const getValue = (key: any) => {
```
ckostiw, line 11: You should be able to use `string` for the key typing.

ckostiw (changes requested):

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,98 @@
...
+                    if (isRangeFilterValue(filterValue)) {
+                        let value = "";
+                        if (filterValue.min !== undefined) { value += ` min: ${filterValue.min}` }
+                        if (filterValue.max !== undefined) { value += ` max: ${filterValue.max}` }
+                        if (value !== "") {
+                            valueJSX = value
+                            break;
+                        }
+                        throw new Error('MIN and MAX values not defined for Filter range')
+                    }
```
ckostiw, line 53: This case is missing a `break`

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,98 @@
...
+                <>
+                    {
+                        <strong>{description.label}: </strong>
+                    }
+                </>
+            );
+
+            switch (description.type) {
+                case FILTER_TYPE.SELECT: {
+                    if (isStringArrayFilterValue(filterValue)) {
```
ckostiw, line 26: This needs to support both a string array filter value, and a single string filter value. Right now, the Depeleted filter value is being displayed as "true" or "false", but the old JS code displayed the labels "Yes" or "No". The value is a boolean, but we want to display the label to match the Yes or No labels the user sees in the filter control.

frontend/src/components/filters/FiltersInfoTS.tsx
```diff
@@ -0,0 +1,98 @@
+import React from "react";
+import { FILTER_TYPE } from "../../constants";
+import { FilterDescription, FilterSet, FilterValue, RangeFilterValue, StringArrayFilterValue, isMetadataFilterValue, isRangeFilterValue, isStringArrayFilterValue } from "../../models/paged_items";
```
ckostiw, line 3: There are a couple of unused imports. You can use the VS Code command "Organize Imports" to clean out any unused imports in your source file (cmd-shift-P opens the command palette).

frontend/src/modules/labwork/actions.ts
```diff
@@ -5,17 +5,17 @@ import { refreshSamplesAtStep } from '../labworkSteps/actions'
 import { findChangedStepsInSummary, processFMSLabworkSummary } from './services'
 
 export const GET_LABWORK_SUMMARY = createNetworkActionTypes('SAMPLE-NEXT-STEP.GET_LABWORK_SUMMARY')
+export const LIST_TABLE = createNetworkActionTypes("SAMPLE-NEXT-STEP.LIST_TABLE");
```
ckostiw, line 8: I'm not sure why you added this action definition - it doesn't seem to be used anywhere.

ckostiw (changes requested):

ckostiw (approved):

PR #533, opened 2023-04-24
Experiment run validation tab

ckostiw:
WIP


frontend/src/components/datasets/DatasetDetailContent.tsx
```diff
@@ -40,20 +43,20 @@ const getTableColumns = (setReleaseStatus, releaseStatusOption) => {
...
             title: "Release Status",
             dataIndex: "release_status",
-            render: (release_status, file) => {
+            render: (release_status: number, file) => {
                 const { id } = file;
                 const releaseStatus = releaseStatusOption.specific[id] ?? releaseStatusOption.all ?? release_status
                 const changed = (releaseStatusOption.all && releaseStatusOption.all !== release_status && !releaseStatusOption.specific[id]) || (!releaseStatusOption.all && releaseStatusOption.specific[id])
                 return <>
-                    <Button style={{ color: changed ? "red" : "grey", width: "6em" }} onClick={() => setReleaseStatus(id, OPPOSITE_STATUS[releaseStatus])}>{RELEASE_STATUS_STRING[releaseStatus]}</Button>
+                    <Button style={{ color: changed ? "red" : "grey", width: "6em" }} onClick={() => toggleReleaseStatus(id, OPPOSITE_STATUS[releaseStatus])}>{RELEASE_STATUS_STRING[releaseStatus]}</Button>
```
UlysseFG, line 51: You can toggle the button even if the validation status is not set..

frontend/src/components/datasets/DatasetDetailContent.tsx
```diff
@@ -157,7 +193,7 @@ const DatasetDetailContent = () => {
     const extraButtons = <>
         <Button
             style={{ margin: 6 }}
-            onClick={(ev) => {
+            onClick={() => {
```
UlysseFG, line 196: The behaviour here is a bit strange for the user. Having to select Release all first to get access to Block all is not intuitive. (i know it is not new but it could be improved without doing too much.

frontend/src/components/experimentRuns/ExperimentRunValidation.tsx
```diff
@@ -0,0 +1,231 @@
...
+									</a>
+								</List.Item>
+							)
+						})}
+					</List>
+				)}
+
+				<Title level={5}>{title}</Title>
+				<Space>
+					<Button disabled={isValidationInProgress}
```
UlysseFG, line 202: I don't really like that the status can be flipped freely around. This makes it easy to set but it doesn't make the decision final ... We want to easily be able to see if a run was validated but this makes it possible to toggle it at a later time obscuring the original value at a given time to the user. Validation should not flip value every day. I would suggest 2 buttons (passed - failed) with modal confirmation. Reset would have to  be requested as a curation (or we could have the reset button be only visible to staff (us) for easy curation).

backend/fms_core/viewsets/experiment_run.py
```diff
@@ -80,7 +80,7 @@ def set_experiment_run_lane_validation_status(self, _request):
         run_name = _request.data.get("run_name", None)
         lane = _request.data.get("lane", None)
         validation_status = _request.data.get("validation_status", None)
-        validation_status = int(validation_status) if validation_status else None
+        validation_status = int(validation_status) if validation_status is not None else None
```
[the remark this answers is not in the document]
ckostiw: In the frontend I was expecting to get AVAILABLE, PASSED, or FAILED. Instead I was getting null. I thought it was a bug, so I fixed it. :) For the frontend, it is much clearer if we use AVAILABLE, so that we can always expect some value in this field. With the UX change I just made, users will no longer be able to reset the status to AVAILABLE - only staff will be able to clear the validation status, so we shouldn't have to worry about the problem you are describing here.

frontend/src/components/datasets/DatasetDetailContent.tsx
ckostiw: I've fixed that, so that both Release All and Block All are available.

frontend/src/components/experimentRuns/ExperimentRunOverview.tsx
```diff
@@ -0,0 +1,68 @@
...
+				<Descriptions.Item label="Experiment Start Date" span={3}>
+					{experimentRun.start_date}
+				</Descriptions.Item>
+				<Descriptions.Item label="Container Barcode" span={3}>
+					{experimentRun.container && <Link to={`/containers/${experimentRun.container}`}>{container.barcode}</Link>}
+				</Descriptions.Item>
+				<Descriptions.Item label="Comment" span={3}>
+					{process.comment}
+				</Descriptions.Item>
+				<Descriptions.Item label="Launch Date" span={3}>{formatDate(experimentRun.run_processing_launch_time)}</Descriptions.Item>
```
ckostiw, line 58: We have `run_processing_launch_date` and `run_processing_launch_time` but the launch date property is empty, and the launch time property contains a complete data/time timestamp, so I have ignored `run_processing_launch_date`...

UlysseFG: Not sure why there would be a launch date ... there is no matching backend attribute. Maybe an artifact of an old name,

frontend/src/components/experimentRuns/ReadsPerSampleGraph.tsx
```diff
@@ -0,0 +1,131 @@
...
+	const { ref: resizeRef, size: componentSize } = useResizeObserver(800, 0)
+	const [graphWidth, setGraphWidth] = useState<number>(DEFAULT_GRAPH_WIDTH)
+
+	useEffect(() => {
+		if (!lane.readsPerSample) {
+			dispatch(loadReadsPerSample(lane.runName, lane.laneNumber))
+		}
+	}, [lane, dispatch])
+
+	useLayoutEffect(() => {
```
nafiz1001, line 33: > useLayoutEffect can hurt performance. Prefer [useEffect](https://react.dev/reference/react/useEffect) when possible.

Is this the risk you're willing to take 🤔

frontend/src/components/shared/WithItemRenderComponent.tsx
```diff
@@ -79,7 +79,9 @@ function WithItemRenderComponentFactory<W extends WithItemFunc, T extends FMSTra
             return render(object)
         } else {
             if (placeholder) {
-                return placeholder
+                if (typeof placeholder === 'string') {
+                    return <span>placeholder</span>
+                } else return placeholder
```
nafiz1001, line 84: Doesn't just returning placeholder work? Because `string` should be part of `ReactNode`.

frontend/src/components/shared/WorkflowSamplesTable/getFilterPropsTS.tsx
```diff
@@ -23,7 +23,37 @@ interface FreezemanColumnType<T> extends TableColumnType<T> {
...
+ */
+export function getFilterPropsIncludingDescriptions(
+	column: FreezemanColumnType<any>,
+	descriptions: FilterDescriptionSet, 
+	filters: FilterSet,
+	setFilter: SetFilterFunc,
+	setFilterOption: SetFilterOptionFunc) {
+	
+	const key = column.dataIndex
+	if (key && typeof key === 'string') {
```
nafiz1001, line 48: can't typescript enforce the `string` type?

frontend/src/models/paged_items_reducers.ts
nafiz1001, line 1: I get type errors in this file. I believe the fixes in branch [ux-run-validation-1642-type](https://github.com/c3g/freezeman/tree/ux-run-validation-1642-type) might be able to fix that

frontend/src/modules/experimentRunLanes/services.ts
```diff
@@ -0,0 +1,81 @@
...
+        Group them by lane number.
+        Determine the list of lanes for the experiment.
+        For each lane:
+            Request the validation status for the lane
+    */
+    // Request all of the datasets associated with the experiment run
+    const datasetOptions = {
+        run_name: experimentRunName
+    }
+    const response = await store.dispatch(api.datasets.list(datasetOptions))
```
nafiz1001, line 20: why do you sometimes `useAppDispatch` and other times use the `store.dispatch` directly?

nafiz1001 (approved):

frontend/src/modules/experimentRunLanes/services.ts
ckostiw: useAppDispatch is a hook that can only be used from components. Services are just plain functions, so I have to dispatch actions using the imported `store`. I'm only dispatching actions to be able to use the api, which is based on thunks that have to be dispatched to be executed.

frontend/src/models/paged_items_reducers.ts
ckostiw: Yes, I didn't try to fix those typing problems since we are going to overhaul the paged items stuff.

frontend/src/components/shared/WorkflowSamplesTable/getFilterPropsTS.tsx
ckostiw: The antd column definition allows the key to be either a string or a string array[], so I have to make sure the key is a string. I don't support mapping string arrays to filters.

frontend/src/components/shared/WithItemRenderComponent.tsx
ckostiw: TBH, I don't remember why I wrapped the string with a span. I think I ran into a problem somewhere and the span fixed it.

frontend/src/components/experimentRuns/ReadsPerSampleGraph.tsx
ckostiw: useLayoutEffect is run before rendering. We have to compute the graph size before it is rendered, so this is the kind of thing useLayoutEffect was designed for.

frontend/src/components/experimentRuns/ExperimentRunOverview.tsx
ckostiw: Yeah, I probably wrote that while the run validation was still in development.

PR #536, opened 2023-04-26
Run validation

UlysseFG:


backend/fms_core/serializers.py
```diff
@@ -164,7 +167,33 @@ class ExperimentRunExportSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ExperimentRun
-        fields = ('experiment_run_id', 'experiment_run_name', 'run_type', 'instrument', 'container_kind', 'container_name', 'container_barcode', 'start_date', 'run_processing_launch_date')
+        fields = ('experiment_run_id',
```
ckostiw, line 170: Maybe the serializer should list the lanes associated with the experiment run, the same way as done for external experiments. Otherwise, I'll have to make an extra api call to get the lanes associated with a freezeman experiment run.

PR #537, opened 2023-05-01
Move library size to sample

UlysseFG:


frontend/src/components/samples/details/SampleDetailsPool.js
```diff
@@ -65,11 +65,6 @@ const getTableColumns = (sampleKinds) => {
           dataIndex: "library_selection_target",
           sorter: true
         },
-        {
```
ckostiw, line 68: I'm not sure why this column is removed. Is it just because we don't want to display the sizes of individual libraries once they have been pooled?

UlysseFG: We don't have to ... The pooled samples will not have a library size. We could display a column that comes from the sample model (all the same one), but we already display it on the sample detail page.

ckostiw (approved):

PR #540, opened 2023-05-04
Fix react errors sidebar

ckostiw:
Fixes the remaining react state errors for all of the items loaded from the sidebar, plus a couple of unique key errors.

NOTE: This PR contains an important fix for study samples, when workflows contain the sample step more than once. I originally used the step ID to group samples, but then the UX got confused when a workflow contained the same step two or more times. Now the step order is used, which is unique, and samples are grouped properly at each step in the workflow.


frontend/src/components/App.js
nafiz1001, line 1: At line 198, you're missing dispatch as part of the useEffect dependency.

nafiz1001:
When I create the following study I get some errors.

image

PR #543, opened 2023-05-04
Taxon definition

JMarce01:
- Added component to `TaxonListContent.tsx`.
- Added `EditTaxon.tsx`.
- Added to `actions.js`, and `api.js` to call the api and backend.
- Modified `taxon.py` added create and update in TaxonViewSet (note: would like feedback on error handling, I haven't been able to find a standardized version in the project).
- Update `TaxonsRoute.tsx` for to point to add and edit component routes.


frontend/src/components/taxons/TaxonsRoute.tsx
```diff
@@ -1,15 +1,26 @@
...
 import TaxonsListContent from './TaxonsListContent'
+import PageContainer from '../PageContainer'
+import { Navigate, Route, Routes } from 'react-router-dom'
+import { AddTaxonRoute, EditTaxonRoute } from './EditTaxon'
 
-function TaxonsRoute() {
+const TaxonsRoute = () => {
 	const taxonsState = useAppSelector(selectTaxonsByID)
-	const taxons = getAllItems(taxonsState)
+	const taxons: Taxon[] = getAllItems(taxonsState)
```
ckostiw, line 12: `taxons` and `taxonsState` are not being used, so these two lines can be removed.

frontend/src/components/taxons/EditTaxon.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const props = (name: string) =>
+        !formErrors[name] ? { name } : {
+            name,
+            hasFeedback: true,
+            // validateStatus: "error", 
+            help: formErrors[name],
+        }
+    const onFinish = () => {
+        const new_taxon: Taxon = taxon;
```
ckostiw, line 52: Careful! You are creating a variable that is pointing to the original taxon object and then modifying the object. You should be creating a clone instead, `const new_taxon: Taxon = {...taxon}`.

frontend/src/components/taxons/EditTaxon.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React, { useCallback, useState } from "react";
+import { requiredRules } from "../../constants";
+import { useAppDispatch, useAppSelector } from "../../hooks";
+import { selectAppInitialzed, selectTaxonsByID } from "../../selectors";
+import { useNavigate, useParams } from "react-router-dom";
+import { add, update } from "../../modules/taxons/actions";
+import AppPageHeader from "../AppPageHeader";
+import PageContent from "../PageContent";
+
+export interface Taxon {
```
ckostiw, line 11: The type `Taxon` is already defined in `frontend_models.ts`. I would name this "EditableTaxon" to avoid confusion. Also, I don't think you need to export this interface, since it will only be used in this file.

frontend/src/components/taxons/EditTaxon.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    return appInitialzed ? <EditTaxon taxon={emptyTaxon} /> : null
+
+}
+export const EditTaxonRoute = () => {
+    const taxons = useAppSelector(selectTaxonsByID)
+    const { id } = useParams();
+    const appInitialzed = useAppSelector(selectAppInitialzed)
+
+    return appInitialzed ? <EditTaxon taxon={taxons[id ?? -1]} /> : null
```
ckostiw, line 32: I'm not sure what `taxons[-1]` is going to result in, but we should definitely not be creating the EditTaxon component if the id is undefined, nor if the taxon with the given id cannot be found. In those cases, the route should return null. The point of the route is to avoid ever passing an undefined Taxon prop to EditTaxon, so that we can write the EditTaxon code without having to check for an undefined `taxon` value every time we use it.

It's also a good idea to pass a clone of the taxon object as a prop, just to be sure the original copy doesn't get modified by accident.

ckostiw, line 32:
```suggestion
    const taxon = id ? taxons[id] : undefined
    return (appInitialzed && taxon) ? <EditTaxon taxon={{...taxon}} /> : null
```

frontend/src/components/taxons/EditTaxon.tsx
```diff
@@ -0,0 +1,109 @@
...
+        !formErrors[name] ? { name } : {
+            name,
+            hasFeedback: true,
+            // validateStatus: "error", 
+            help: formErrors[name],
+        }
+    const onFinish = () => {
+        const new_taxon: Taxon = taxon;
+        const fieldValues = form.getFieldsValue();
+        Object.keys(fieldValues).forEach((fieldName: string) => {
```
ckostiw, line 54: I'm not crazy about using Object.keys here. It's defeating typescript's type checking. Since there are only a few fields, it might be better to just assign the id, ncbi_id and name properties explicitly.

frontend/src/components/taxons/EditTaxon.tsx
ckostiw: Another option would be to use Pick<> to define a new type from the existing Taxon definition:
`type EditableTaxon = Pick<Taxon, 'id'|'ncbi_id'|'name'>`
The only trouble is that you need id to be optional, so you would have to also use `Partial<>`, which makes all properties optional:
`type EditableTaxon = Partial<Pick<Taxon, 'id'|'ncbi_id'|'name'>>`

This is the point where some devs start to hate typescript :)

The advantage to doing it this way is that if the `Taxon` type is changed, and one of those properties no longer exists, then the code will break and the compiler will discover the problem for future us.

ckostiw (changes requested):

frontend/src/components/App.js
```diff
@@ -223,53 +223,54 @@ const App = ({userID, usersByID, logOut, get}) => {
...
+              {MENU_ITEMS.map(renderMenuItem)}
+            </Menu>
+            {isLoggedIn &&
               <Menu
                 theme="dark"
                 mode="inline"
                 selectedKeys={matchingMenuKeys(MENU_ITEMS)}
-                style={{flex: 1}}
+                style={{ flex: 1 }}
                 defaultOpenKeys={['definitions']} // Submenus should be open by default
```
ckostiw, line 264: This Menu doesn't contain the Definitions section - that's in MENU_ITEMS, so you should move `defaultOpenKeys` to the previous Menu.

menuItems just contains the About, Profile, and Sign Out items.

ckostiw (changes requested):

frontend/src/modules/taxons/actions.js
```diff
@@ -12,6 +14,24 @@ export const get = id => async (dispatch, getState) => {
...
+    if (getState().taxons.isFetching)
+        return;
+
+    return await dispatch(networkAction(
+        ADD, api.taxons.add(taxon), { meta: { ignoreError: 'APIError' } }
+    ));
+};
+
+export const update = (id, taxon) => async (dispatch, getState) => {
+    if (getState().taxons.itemsByID[id].isFetching)
```
ckostiw, line 27: I don't think this is necessary either. We should be able to call the backend to update even if the frontend is busy fetching this taxon. Did this fix a bug you ran into?

frontend/src/components/taxons/TaxonsListContent.tsx
```diff
@@ -1,23 +1,52 @@
...
+import { selectTaxonsByID } from '../../selectors'
 
 export interface TaxonsListContentProps {
-	taxons: Taxon[]
+	taxons: Taxon[],
 }
 
-function TaxonsListContent({taxons} : TaxonsListContentProps) {
+function TaxonsListContent() {
+	const taxonsState = useAppSelector(selectTaxonsByID)
```
ckostiw, line 17: Naming this `taxonsByID` would match our naming conventions a bit better.

frontend/src/components/taxons/TaxonsListContent.tsx
```diff
@@ -1,23 +1,52 @@
...
+	const taxons: Taxon[] = getAllItems(taxonsState)
+	const [taxonColumns, setTaxonColumns] = useState<ObjectWithTaxon[]>();
```
ckostiw, line 19: These aren't columns, they are data objects, so I would use `const [taxonsData, setTaxonsData]` here. 
I might be a good idea to move `const taxons: Taxon[] = getAllItems(taxonsState)` into the useEffect, since that variable is only used inside the useEffect. You would need to change the useEffect dependency to `[taxonsState]` (or `[taxonsByID]`) in that case.

ckostiw (changes requested):

ckostiw (approved):

PR #544, opened 2023-05-04
Sort order on coordinates column for selected samples

ckostiw:
Add ascending/descending sort order on the coordinates column for selected samples in labwork step page.


UlysseFG:
- The labwork sample list (previous to selection) do not have the same coordinate sorting option which is a bit strange. 
- The unselected sample coordinate ordering do not order within a plate which is likely to cause some confusion.
(longer term we'll likely have to make coordinate search apply everywhere in the app.)
- could be nice to have an independant ascending descending sorter for container so we can get them in an inverted order (both sort would always be active default would be ascending).

PR #547, opened 2023-05-08
Reduce redux logger noise in console

ckostiw:
I'd like it if redux-logger could be configured to log all redux actions at the debug console level, so that we could turn them off if we don't need to see the logged actions - the actions add a lot of noise to the console. Unfortunately, actions are logged as console groups, and groups are always logged at the 'info' level. Browsers do not support logging groups at other levels. So, this PR just reduces the noise a bit by having the action groups collapsed by default, and by not logging the SUMMARY actions, which otherwise pollute the log every 30 seconds.


UlysseFG:
Browsers do not support logging at lower levels... does it mean you could log summary at the debug level and have the browser tools ignore them ? this might allow for use of additional tool to debug the summary spam if required ?

ckostiw:
> Browsers do not support logging at lower levels... does it mean you could log summary at the debug level and have the browser tools ignore them ? this might allow for use of additional tool to debug the summary spam if required ?

That's what I wanted to do, but there's no way to stop redux-logger from logging each action at the 'info' level, because it uses console groups to log actions and groups always appear at the 'info' level. I do log any summary ERROR actions, if errors occur, and we can use the redux-devtools chrome extension to debug the summaries actions if we really need to (it tracks every action that is dispatched).

UlysseFG:
Redux console is the main tool i use to track the frontend. You are welcome to reduce the spam, but it would be nice to still see the information at a different level (if debug is not available there seem to be a few options). If there is no log for for the summaries ... at some point i expect it will stop working and we won't notice ...

ckostiw:
> 

I don't think that's a good enough reason to continue logging summaries every 30 seconds. It makes it really hard to use the console log for anything else. If I could just move the redux action logging to the 'debug' level then I wouldn't have to suppress the summaries actions, but that's not possible. With the change I made, any summaries errors will be logged, but REQUEST and RECEIVE are ignored.

UlysseFG (approved):

PR #549, opened 2023-05-08
Ref genomes definition

JMarce01:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
```
ckostiw, line 2: It's better not to commit vscode settings changes with your PR's. Adopting a python formatter would have to be a team decision.

frontend/src/components/referenceGenomes/ReferenceGenomesListContent.tsx
```diff
@@ -1,21 +1,42 @@
...
+import { getColumnsForReferenceGenome } from './ReferenceGenomeTableColumns'
+import { ReferenceGenome } from '../../models/frontend_models'
 
-export interface ReferenceGenomesListContentProps {
-	referenceGenomes: ReferenceGenome[]
-}
 
-function ReferenceGenomesListContent({referenceGenomes} : ReferenceGenomesListContentProps) {
+function ReferenceGenomesListContent() {
+	const [referenceGenomeColumns, setReferenceGenomeColumns] = useState<ReferenceGenome[]>();
```
ckostiw, line 14: This is super confusing because this state is storing ReferenceGenome objects, not columns. I guess the names should be `[referenceGenomes, setReferenceGenomes]` here.

frontend/src/components/referenceGenomes/ReferenceGenomeTableColumns.tsx
```diff
@@ -0,0 +1,120 @@
...
+    },
+    {
+        columnID: ReferenceGenomeID.TAXON_ID,
+        title: 'Taxon',
+        dataIndex: ['referenceGenome', 'taxon_id'],
+        sorter: (a, b) => a.taxon_id - b.taxon_id,
+        render: (_, { taxon_id }) =>
+            taxon_id && (
+                <Link to={`/taxons/update/${taxon_id}`}>
+                    <div>{taxonsByID[taxon_id].name}</div>
```
ckostiw, line 105: Here it would be safer to use a "WithItemRenderComponent". That's a component that takes an object id and will fetch the object, if necessary. This code would bomb if the taxon with `taxon_id` doesn't exist. You will have to add a new `WithTaxonRenderComponent` for this in the file `WithItemRenderComponent.tsx`.
You won't need to pass the taxonsByID parameter to this function anymore, once you have a WithTaxonRenderComponent here.

frontend/src/components/referenceGenomes/EditReferenceGenomes.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    const [form] = Form.useForm()
+    const [formErrors, setFormErrors] = useState({})
+
+    const isAdding = referenceGenome === undefined
+    const navigate = useNavigate();
+    const dispatch = useAppDispatch();
+
+    const taxonsByID = useAppSelector(selectTaxonsByID)
+    const taxonOptions = Object.keys(taxonsByID).map((id) => taxonsByID[id])
```
ckostiw, line 41: I wrote a function to get the array of items out of an ItemsByID object, `getAllItems` in `frontend_models.ts` so you could use that here.

ckostiw (changes requested):

ckostiw (approved):

PR #551, opened 2023-05-11
Feature #1471: To be able to edit a Study when no samples have been assigned to it

nafiz1001:


backend/fms_core/serializers.py
```diff
@@ -729,10 +729,20 @@ class Meta:
         fields = ("id", "assembly_name", "synonym", "genbank_id", "refseq_id", "taxon_id", "size")
 
 class StudySerializer(serializers.ModelSerializer):
+    removable = serializers.SerializerMethodField(read_only=True)
     class Meta:
         model = Study
-        fields = ("id", "letter", "project_id", "workflow_id", "start", "end")
-    
+        fields = ("id", "letter", "project_id", "workflow_id", "start", "end", "removable")
```
UlysseFG, line 735: No... if the object is not updated in the frontend you'll have issues ...

backend/fms_core/serializers.py
```diff
@@ -729,10 +729,20 @@ class Meta:
...
+
+    def get_removable(self, instance: Study):
+        if StepHistory.objects.filter(study=instance.pk).exists():
+            return False
+        if SampleNextStep.objects.filter(studies__id=instance.pk).exists():
+            return False
+        if SampleNextStepByStudy.objects.filter(study=instance.pk).exists():
+            return False
+        return True
```
nafiz1001, line 744: service it

backend/fms_core/services/study.py
```diff
@@ -36,9 +36,17 @@ def create_study(project, workflow, start, end):
...
+    letter = chr(ord(max(occupied_letters)) + 1) if len(occupied_letters) > 0 else 'A'
+    
+    # Find earlier missing letter
+    for c in range(len(occupied_letters)):
+        expected_letter = chr(ord('A') + c)
+        # assumes occupied_letters is sorted alphabetically
+        if expected_letter != occupied_letters[c]:
+            # expected_letter missing
+            letter = expected_letter
+            break
```
nafiz1001, line 49: use constant for 'A'

backend/fms_core/viewsets/study.py
```diff
@@ -55,4 +57,23 @@ def create(self, request, *args, **kwargs):
...
+
+        if pk is None:
+            errors['Study'].append("pk cannot be None")
+        else:
+            if StepHistory.objects.filter(study=pk).exists():
+                errors['StepHistory'].append("At least one StepHistory is associated with the Study")
+            if SampleNextStep.objects.filter(studies__id=pk).exists():
+                errors['SampleNextStep'].append("At least one SampleNextStep is associated with the Study")
+            if SampleNextStepByStudy.objects.filter(study=pk).exists():
+                errors['SampleNextStepByStudy'].append("At least one SampleNextStepByStudy is associated with the Study")
```
nafiz1001, line 73: service it

frontend/src/modules/studies/reducers.ts
```diff
@@ -63,6 +66,13 @@ export const studies = (
 		}
 		case LIST_PROJECT_STUDIES.ERROR:
 			return { ...state, isFetching: false, error: action.error, };
+		
+		case REMOVE.REQUEST:
+			return merge<StudiesState>(state, ['itemsByID', action.meta.id], { id: action.meta?.id, isRemoving: true })
+		case REMOVE.RECEIVE:
+			return del<StudiesState>(state, ['itemsByID', action.meta.id])
```
nafiz1001, line 73: remove item from `items`

nafiz1001:
@nafiz1001 oh and don't forget to test

backend/fms_core/services/study.py
```diff
@@ -97,3 +99,15 @@ def get_study(project_obj: Project, study_letter: str):
         errors.append(f"Both a project and a study letter are required to retrieve a study.")
 
     return study, errors, warnings
+
+def delete_study_errors(study: int) -> Dict[str, List[str]]:
```
UlysseFG, line 103: rename it to tell what it actually do (is_study_used, is_study_referenced, can_remove_study, ...)

backend/fms_core/services/study.py
```diff
@@ -97,3 +99,15 @@ def get_study(project_obj: Project, study_letter: str):
...
+    errors = defaultdict(list)
+
+    if StepHistory.objects.filter(study=study).exists():
+        errors['StepHistory'].append("At least one StepHistory is associated with the Study")
+    if SampleNextStep.objects.filter(studies__id=study).exists():
+        errors['SampleNextStep'].append("At least one SampleNextStep is associated with the Study")
+    if SampleNextStepByStudy.objects.filter(study=study).exists():
+        errors['SampleNextStepByStudy'].append("At least one SampleNextStepByStudy is associated with the Study")
+
+    return errors
```
UlysseFG, line 113: I would prefer you return a more "standard" tuple here. return a boolean to answer to the question and errors and warning (even if there is no warning yet) just to keep the common service format.

backend/fms_core/services/study.py
```diff
@@ -1,8 +1,29 @@
...
+from typing import Dict, List, Tuple
 import string
 
+def new_letter(project):
+    def skip_char(letter: str, distance: int):
+        return chr(ord(letter) + distance)
+
+    FIRST_LETTER = 'A'
+    occupied_letters = sorted(Study.objects.filter(project=project).values_list('letter', flat=True))
+    new_letter = skip_char(max(occupied_letters), 1) if len(occupied_letters) > 0 else FIRST_LETTER
```
UlysseFG, line 14: if you sort your list (as you do here), you should slice here instead of using max. The last item will be your max,

backend/fms_core/services/study.py
```diff
@@ -80,3 +99,19 @@ def get_study(project_obj: Project, study_letter: str):
         errors.append(f"Both a project and a study letter are required to retrieve a study.")
 
     return study, errors, warnings
+
+def can_remove_study(study: int) -> Tuple[bool, Dict[str, List[str]], Dict[str, List[str]]]:
```
UlysseFG, line 103: If you pass an int here you are not passing a Study but an int. Rename param to study_id to clarify

backend/fms_core/services/study.py
```diff
@@ -80,3 +99,19 @@ def get_study(project_obj: Project, study_letter: str):
...
+    warnings = {}
+    errors = defaultdict(list)
+    is_removable = True
+
+    if StepHistory.objects.filter(study=study).exists():
```
UlysseFG, line 108: filter using study_id=study_id instead of study=study_id... I am not convinced an instance of study can be equated to an int in django. This will not fail it will just always be false.

backend/fms_core/services/study.py
```diff
@@ -80,3 +99,19 @@ def get_study(project_obj: Project, study_letter: str):
...
+        errors['StepHistory'].append("At least one StepHistory is associated with the Study")
+    if SampleNextStep.objects.filter(studies__id=study).exists():
+        errors['SampleNextStep'].append("At least one SampleNextStep is associated with the Study")
+    if SampleNextStepByStudy.objects.filter(study=study).exists():
```
UlysseFG, line 112: Same as above about study_id...

backend/fms_core/utils.py
```diff
@@ -33,6 +33,10 @@
 
 TRUTH_VALUES = frozenset({"TRUE", "T", "YES", "Y"})
 
+T = TypeVar("T") # generic variable type
```
UlysseFG, line 36: Please don't start typing python ...

backend/fms_core/viewsets/study.py
```diff
@@ -55,4 +58,19 @@ def create(self, request, *args, **kwargs):
...
+            return Response(serializer.data)
+
+    def destroy(self, request, pk=None):
+        errors = defaultdict(list)
+
+        if pk is None:
+            errors['Study'].append("pk cannot be None")
+        else:
+            _, newerrors, _ = can_remove_study(pk)
+            errors = merge_dicts_of_list(newerrors, errors)
```
UlysseFG, line 70: This is a very busy way of doing this. It is quite generic but i think a bit overkill i would suggest you pass a list from errors and limit you service to returning a single error list (as is done elsewhere in the application).  For example, removable, errors['Removable'], _ = can_remove_study(pk).

ckostiw (changes requested):
Please add a confirmation popup before deleting the study.

backend/fms_core/services/study.py
```diff
@@ -80,3 +99,19 @@ def get_study(project_obj: Project, study_letter: str):
         errors.append(f"Both a project and a study letter are required to retrieve a study.")
 
     return study, errors, warnings
+
+def can_remove_study(study: int) -> Tuple[bool, Dict[str, List[str]], Dict[str, List[str]]]:
+    warnings = {}
+    errors = defaultdict(list)
```
UlysseFG, line 105: You can use a simple list for errors and warnings you don't need the extra structure (and hassle).

frontend/src/modules/studies/actions.ts
```diff
@@ -55,6 +55,12 @@ export const listProjectStudies = (projectId: number) => {
 	}
 }
 
+export const remove = (id: number) => async (dispatch: AppDispatch, getState: () => RootState) => {
```
ckostiw, line 58: Please use `FMSId` as the type, rather than `number` when the type is an ID.

backend/fms_core/services/study.py
nafiz1001: you sure it's not two underscores?

backend/fms_core/services/study.py
nafiz1001: so `errors = []` and `warnings = []` ?

frontend/src/components/projects/ProjectsDetailedContent.tsx
```diff
@@ -62,10 +54,14 @@ interface ProjectsDetailedContentProps {
 }
 
 const ProjectsDetailedContent = ({project, studies} : ProjectsDetailedContentProps) => {
+	const dispatch = useAppDispatch()
+
 	const navigate = useNavigate()
 
 	const [activeKey, setActiveKey] = useHashURL('overview')
 
+	const currentStudy = activeKey.startsWith("study") ? Number(activeKey.split("-")[1]) : null
```
ckostiw, line 63: This seems very hacky to me and will break if someone decides to tweak the key string for some reason. I suggest creating a map of keys to study id's, and looking up the study ID that matches `activeKey`. That way, it won't break if the key format changes.

frontend/src/components/projects/ProjectsDetailedContent.tsx
```diff
@@ -78,12 +74,21 @@ const ProjectsDetailedContent = ({project, studies} : ProjectsDetailedContentPro
 
 	const title = `Project ${project ? project.name : ''}`
 
+	const handleRemoveStudy = useCallback(async () => {
+		if (currentStudy !== null) {
+			await dispatch(removeStudy(currentStudy))
```
ckostiw, line 79: Is the `await` necessary?

ckostiw (changes requested):

ckostiw:
I had studies A, B, C then I deleted A. When I created a new study, I had a new Study A, but now appearing as the last study. That will look a bit strange to the users.
Screen Shot 2023-05-17 at 11 22 47 AM

backend/fms_core/services/study.py
UlysseFG: two underscore means it follows the FK to the next model, but django as a strange way of defining FK it creates an attribute by appending an "_id" after the name you gave the FK field. So the study fk fields has a study_id attribute which does not require following the relation. It is equivalent, but i am not sure Django does not query the foreign model if you use the "__" notation.

frontend/src/components/projects/ProjectsDetailedContent.tsx
```diff
@@ -92,7 +97,7 @@ const ProjectsDetailedContent = ({project, studies} : ProjectsDetailedContentPro
 			{project && (
 				<PageContent loading={false} style={undefined}>
 					{ project && 
-						<Tabs activeKey={activeKey} onChange={setActiveKey} size="large" type="card" style={tabsStyle} tabBarExtraContent={addStudyButton}>
+						<Tabs activeKey={activeKey} onChange={setActiveKey} size="large" type="card" style={tabsStyle} tabBarExtraContent={<>{addStudyButton}{removeStudyButton}</>}>
```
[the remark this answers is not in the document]
UlysseFG: I personally don't like having the delete button alongside the add study button. This button makes no sense if you are on the detail tab or on the sample by project tab ... it's also likely that someone will expect this to pop a question about which study. I would prefer something tied to a specific study tab.

frontend/src/components/projects/ProjectsDetailedContent.tsx
```diff
@@ -60,7 +67,10 @@ const ProjectsDetailedContent = ({project, studies} : ProjectsDetailedContentPro
 
 	const [activeKey, setActiveKey] = useHashURL('overview')
 
-	const currentStudy = activeKey.startsWith("study") ? Number(activeKey.split("-")[1]) : null
+	const studyTabKeyToID = useMemo(() => Object.fromEntries(studies.map((study) => [createStudyTabKey(study.id), study.id])), [studies])
```
ckostiw: You should use the mapped keys as the `key` below where the TabPanes for the studies are created, just to be sure the same key prop is set on the TabPane (right now, we are calling `createStudyTabKey` to generate the keys)

ckostiw (changes requested):

ckostiw:
Could we only render the "ⓘ Studies containing samples cannot be removed." message when the button is disabled?

ckostiw (approved):

backend/fms_core/viewsets/study.py
```diff
@@ -55,4 +57,18 @@ def create(self, request, *args, **kwargs):
...
+
+        if pk is None:
+            errors['Study'].append("pk cannot be None")
+        else:
+            _, errors['Removable'], _ = can_remove_study(pk)
+
+        if any(bool(error) for error in errors.values()):
+            raise ValidationError(errors)
+        else:
+            Study.objects.filter(pk=pk).delete()
```
UlysseFG, line 73: Check if it calls the Study overloaded delete() function. This could be calling the manager delete which will not create a version in reversion. If this is the case chain .first() before delete() to the model delete().

UlysseFG (approved):

PR #557, opened 2023-05-15
Hide Actions button for templates if no actions are present

JMarce01:
Added a quick check to see if the actions list is an empty array or not. Hides button if no actions are present in the component.

### DIFF
--- a/frontend/src/utils/templateActions.js
+++ b/frontend/src/utils/templateActions.js
@@ -36,9 +36,10 @@ export const actionsToButtonList = (urlBase, actions, fullWidth = false) =>
 
 export function ActionDropdown({ urlBase, actions, fullWidth = true }) {
   const history = useNavigate();
-  const actionMenu = (
+  const renderActions = actions.items.length > 0;
+  const actionMenu = renderActions ? (
     <Menu>
-      {actions.items ? actions.items.map((a) =>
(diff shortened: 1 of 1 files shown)


frontend/src/utils/templateActions.js
```diff
@@ -25,37 +27,39 @@ export const actionIcon = a => {
...
+        <Menu.Item key={a.id.toString()}>
+          <Button
+            icon={actionIcon(a)}
+            onClick={() => history(`${urlBase}/actions/${a.id}/`)}
+            {...(fullWidth ? { style: { width: "100%", border: 0, textAlign: 'left' } } : { style: { border: 0 } })}
+          >
+            {a.name}
+          </Button>
+        </Menu.Item>) :
+        <Menu.Item>Loading ...</Menu.Item>
```
ckostiw, line 51: I'm pretty sure that this code is creating one "Loading..." menu item but the user will never see it because we don't render the Dropdown if the actions.items array is empty, so adding the Loading item is redundant.
It might be better to just have a single check for `actions.items.length > 0` at the top of the function. If it's empty just return null. If it has items then create the menu and render the Dropdown.

ckostiw (changes requested):

ckostiw (approved):

PR #559, opened 2023-05-18
Bug #1656: Lineage graph is clipped with latest ant 4.x version

nafiz1001:


ckostiw:
I can no longer drag nodes around in the graph. Clicking and dragging a node just drags the whole graph. Is that intentional?

ckostiw:
Now the bounds of one of the elements in the lineage graph tab has become too large, and the page content is now unecessarily scrollable.

https://github.com/c3g/freezeman/assets/104078019/65ac165d-4f82-440d-b3bb-c7c865b0c818

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -352,7 +352,7 @@ const SampleDetailsContent = ({
           </Descriptions>
         </TabPane>
 
-        <TabPane tab={`Lineage`} key="lineage" style={tabStyle}>
+        <TabPane tab={`Lineage`} key="lineage" style={{...tabStyle, minHeight: "100vh", maxheight: "100vh"}}>
```
ckostiw, line 355: Setting the tab height to 100vh is setting the height to the window height, so the tab bottom extends beyond the bottom of the window, making it scrollable.

ckostiw (approved):

frontend/src/components/containers/ContainersDetailContent.js
```diff
@@ -22,6 +22,7 @@ const pageStyle = {
 
 const tabsStyle = {
   marginTop: 8,
+  overflow: 'auto'
```
ckostiw, line 25: This might be a good opportunity to clean up the copy/pasted `tabsStyle` and `tabStyle` stuff, and define the style once in a shared file, rather than adding `overflow:auto` everywhere.

ckostiw (changes requested):

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -180,10 +182,11 @@ const SampleDetailsContent = ({
     })
   }, [sample])
 
-  const tabsStyle = {
-    marginTop: 8,
-    overflow: activeKey === 'lineage' ? 'hidden' : 'auto'
-  }
+  const tabsClass = classNames({
```
ckostiw, line 185: This is generating bad class names: `ant-tabs ant-tabs-top ant-tabs-large ant-tabs-card .ant-tabs .ant-tabs-content.ant-tabs-content-top .lineage-tab-active`. Those dots shouldn't be there, and it's duplicating the properties that are already there. I'm not sure how classNames works, but it's not doing the right thing here.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -205,7 +208,7 @@ const SampleDetailsContent = ({
       {error &&
         <ErrorMessage error={error} />
       }
-      <Tabs activeKey={activeKey} onChange={setActiveKey} size="large" type="card" style={tabsStyle}>
+      <Tabs activeKey={activeKey} onChange={setActiveKey} size="large" type="card" className={tabsClass}>
```
ckostiw, line 211: You only need to add your new class name here:
`className='lineage-tab-active'`
It will add your new class name to the existing ant class names (note that you don't include the dot here).

There's no need to use the classNames package for this.

ckostiw (changes requested):

ckostiw (approved):

PR #563, opened 2023-05-19
Fix completed samples 100 limit

ckostiw:
Fixes the issue with missing completed samples in the project study samples tables - only the first 100 samples were retrieved from the server because no limit was specified and the default limit of 100 was used.

### DIFF
--- a/frontend/src/modules/studySamples/services.ts
+++ b/frontend/src/modules/studySamples/services.ts
@@ -203,7 +203,7 @@ export async function fetchSamplesAtStepOrder(studyID: FMSId, stepOrderID: FMSId
 		options = {ordering, ...serializedFilters}
 	}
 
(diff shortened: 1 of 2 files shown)


frontend/src/modules/studySamples/services.ts
```diff
@@ -46,7 +46,7 @@ export async function loadStudySamples(studyID: FMSId) {
 	
 	// Get samples that have completed the process at a step
 	let completedSamplesByStudy : FMSStepHistory[] | undefined
-	const sampleHistoryResponse = await store.dispatch(api.stepHistory.getCompletedSamplesForStudy(studyID))
+	const sampleHistoryResponse = await store.dispatch(api.stepHistory.getCompletedSamplesForStudy(studyID, {limit: 100000}))
```
nafiz1001, line 49: 🥲

nafiz1001 (approved):

PR #566, opened 2023-05-19
Bug #1634: Jumpbar: including an apostrophe in search query string causes 500 error

nafiz1001:
https://206.12.92.46/issues/1634

### DIFF
--- a/backend/fms_core/viewsets/_utils.py
+++ b/backend/fms_core/viewsets/_utils.py
@@ -34,10 +34,10 @@ class FZY(Func):
     template = "%(function)s(%(expressions)s::cstring)"
     function = "fzy"
 
-    def __init__(self, expression, search_term, **extras):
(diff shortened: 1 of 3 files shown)


backend/fms_core/viewsets/_utils.py
```diff
@@ -31,13 +31,13 @@ def _list_keys(d: Dict[str, Any]) -> Dict[str, Any]:
 
 
 class FZY(Func):
-    template = "%(function)s('%(search_term)s', %(expressions)s::cstring)"
+    template = "%(function)s(%(expressions)s::cstring)"
     function = "fzy"
 
     def __init__(self, expression, search_term, **extras):
```
UlysseFG, line 37: rename expression to search_field and invert search_field and search_term (don't forget to flip the FZY call too).

UlysseFG (approved):

PR #567, opened 2023-05-23
[QC]  New curations tools to delete Samples and Containers

UlysseFG:


ckostiw (approved):
LGTM. I was thinking that if a sample is deleted then the individual should be deleted as well (if there are no other samples associated with the individual), but then I saw that there is a curation for deleting individuals, so I guess the idea would be to delete individuals as a separate curation after the samples are deleted.

PR #569, opened 2023-05-25
Sample submission updates

UlysseFG:


UlysseFG:
Built on top of the additional library types branch to keep the updates included in that version of the template.

PR #570, opened 2023-05-29
Feature #1614: Notification management

nafiz1001:
https://206.12.92.46/issues/1614

### DIFF
--- a/frontend/src/components/ExportButton.js
+++ b/frontend/src/components/ExportButton.js
@@ -30,9 +30,10 @@ const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...res
               downloadFromFile(name, text)
             })
             .catch(err => {
-              dispatch(notifyError(err.message, {
-                title: err.message,
-                description: <pre>{err.stack}</pre>,
+              const key = 'Failed to export'
+              dispatch(notifyError(key, {
+                title: key,
+                description: err.message,
               }))
             })
             .then(() => {
--- a/frontend/src/components/ExportButton.js
+++ b/frontend/src/components/ExportButton.js
@@ -4,9 +4,7 @@ import {Button, Modal} from "antd";
 
 import { DownloadOutlined, ExclamationCircleOutlined } from "@ant-design/icons";
 import { useAppDispatch } from "../hooks";
-import { notify } from "../modules/notification/actions";
-import { NotificationType } from "../modules/notification/models";
-
+import { notifyError } from "../modules/notification/actions";
 const { confirm } = Modal;
 
 const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...rest }) => {
@@ -32,10 +30,9 @@ const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...res
               downloadFromFile(name, text)
             })
             .catch(err => {
-              dispatch(notify(err.message, {
+              dispatch(notifyError(err.message, {
                 title: err.message,
                 description: <pre>{err.stack}</pre>,
-                type: NotificationType.ERROR,
               }))
             })
             .then(() => {
--- a/frontend/src/components/ExportButton.js
+++ b/frontend/src/components/ExportButton.js
@@ -1,13 +1,17 @@
 import React, {useState} from "react";
 import {downloadFromFile} from "../utils/download";
-import {Button, Modal, notification} from "antd";
+import {Button, Modal} from "antd";
 
 import { DownloadOutlined, ExclamationCircleOutlined } from "@ant-design/icons";
+import { useAppDispatch } from "../hooks";
+import { notify } from "../modules/notification/actions";
+import { NotificationType } from "../modules/notification/models";
 
 const { confirm } = Modal;
 
 const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...rest }) => {
   const [loading, setLoading] = useState(false);
+  const dispatch = useAppDispatch()
 
   const name = filename + '_' + new Date().toISOString().slice(0, 10) + '.csv'
 
@@ -28,10 +32,11 @@ const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...res
               downloadFromFile(name, text)
             })
             .catch(err => {
-              notification.error({
-                message: err.message,
+              dispatch(notify(err.message, {
+                title: err.message,
                 description: <pre>{err.stack}</pre>,
-              });
+                type: NotificationType.ERROR,
+              }))
             })
             .then(() => {
               setLoading(false);
--- a/frontend/src/modules/notification/actions.tsx
+++ b/frontend/src/modules/notification/actions.tsx
@@ -25,7 +25,7 @@ export const notify = (id: NotificationID, props: NotificationProps) => (dispatc
     notification[convertToAntdNotificationAPI[props.type]]({
         message: props.title,
         description: <pre style={{ fontSize: '0.8em', whiteSpace: 'pre-wrap' }}>{props.description}</pre>,
-        duration: props.duration,
+        duration: props.duration ?? 4.5, // 4.5 is the default duration in antd
         key: id,
         onClose: () => dispatch(closeNotification(id))
     });
--- a/frontend/src/modules/notification/actions.tsx
+++ b/frontend/src/modules/notification/actions.tsx
@@ -17,7 +17,12 @@ const hasNotification = (state: RootState, id: NotificationID) => {
     return state.notifications.some((notification) => notification.id === id)
 }
 
-export const notify = (id: NotificationID, props: NotificationProps) => (dispatch: AppDispatch, getState: () => RootState) => {
+export interface NotifyProps extends NotificationProps {
+    id: NotificationID
+}
+export const notify = (props: NotifyProps) => (dispatch: AppDispatch, getState: () => RootState) => {
+    const { id } = props
+
     if (hasNotification(getState(), id)) {
         dispatch(closeNotification(id))
     }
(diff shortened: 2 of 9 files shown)


frontend/src/modules/notification/models.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export interface AlertProps {
+    type: AlertType,
+    description: ReactNode
+    title?: string,
+    error?: Error
+}
+
+export type AlertID = string
```
ckostiw, line 21: I like using a string for an ID. For network errors (failed api calls) we could use the URL as an id string. If the same error occurs 500 times, it would ensure that the existing notification is replaced by the new notification, rather than generating 500 alert components. 👍

frontend/src/modules/notification/models.ts
```diff
@@ -0,0 +1,37 @@
...
+export const ALERT_REMOVE = `${ALERT}.REMOVE`
+
+export enum AlertType {
+    SUCCESS = 'success',
+    INFO = 'info',
+    WARNING = 'warning',
+    ERROR = 'error'
+}
+
+export interface AlertProps {
```
ckostiw, line 14: We should have two kinds of alerts - alerts that stay open until the user dismisses them (like we have now) and alerts that stay open for a number of seconds and then are closed automatically. That's useful for info alerts like "Your changes were submitted successfully" which are tiresome to close manually every time they pop up.

frontend/src/modules/notification/reducers.ts
```diff
@@ -0,0 +1,19 @@
+import { Reducer } from "redux"
+import { ALERT_REMOVE, AlertAction, AlertRemoveAction, ALERT, NotificationState } from "./models"
+import {del, merge} from 'object-path-immutable'
+
+export const notifications: Reducer<NotificationState, AlertAction | AlertRemoveAction> = (state, action) => {
+    switch(action.type) {
+        case ALERT: {
+            const alertAction = action as AlertAction
+            return merge<NotificationState>(state ?? {}, [alertAction.id], alertAction.props)
```
ckostiw, line 9: It's important that alerts are displayed in the correct and consistent order (newest to oldest). It might be better to use an array to store them in order.

If a newer notification replaces an existing one with the same ID it should appear as the most recent alert and the older one should disappear. If we just use object properties then the it will not be promoted to the most recent alert, as the order of the keys won't change.

ckostiw (commented):
I like what I see so far. 👍

frontend/src/modules/notification/models.ts
nafiz1001: the existing solution is to set duration to 0 to represent infinity (until user closes it). what do you think of that?

frontend/src/modules/notification/models.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface NotificationAction extends AnyAction {
+    id: NotificationID,
+    props: NotificationProps
+}
+
+export interface NotificationRemoveAction extends AnyAction {
+    id: NotificationID
+}
+
+export interface NotificationState extends NotificationProps {
```
ckostiw, line 32: NotificationItem

frontend/src/modules/notification/actions.tsx
```diff
@@ -0,0 +1,54 @@
...
+    [NotificationType.WARNING]: 'warning',
+    [NotificationType.ERROR]: 'error'
+}
+
+const hasNotification = (state: RootState, id: NotificationID) => {
+    return state.notifications.some((notification) => notification.id === id)
+}
+
+export const notify = (id: NotificationID, props: NotificationProps) => async (dispatch: AppDispatch, getState: () => RootState) => {
+    if (hasNotification(getState(), id)) {
```
ckostiw, line 19: Should close existing notification and replace with the new instance at the head of the list.

frontend/src/modules/notification/actions.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const notify = (id: NotificationID, props: NotificationProps) => (dispatch: AppDispatch, getState: () => RootState) => {
+    if (hasNotification(getState(), id)) {
+        dispatch(closeNotification(id))
+    }
+
+    notification[convertToAntdNotificationAPI[props.type]]({
+        message: props.title,
+        description: <pre style={{ fontSize: '0.8em', whiteSpace: 'pre-wrap' }}>{props.description}</pre>,
+        duration: props.duration ?? 4.5, // 4.5 is the default duration in antd
```
UlysseFG, line 28: Maybe you should not shadow antd default duration by setting 4.5 but instead leave the duration undefined so the default takes over ?

frontend/src/store.ts
```diff
@@ -1,18 +1,62 @@
...
+
+	const error = getErrorDescription(action.error)
+	dispatch(showNotification(error.message, error.details))
+
+	return next(action)
+}
+
 const store = configureStore({
 	reducer: rootReducer,
 	// Disable immutable check because it throws an exception if withItem() tries to modify
```
UlysseFG, line 51: If we have no longer any withItem breaking the immutable check it might be a good time to turn this back on to prevent more breaking of the rule.

nafiz1001: ah i see you're referring to the `configureStore`

frontend/src/modules/notification/actions.tsx
UlysseFG: If antd changes the default it might lead to baffled devs trying to figure why the duration is 4.5 and not the default ;). I typically try not to override a default if i can. It is not a fault though.

frontend/src/store.ts
UlysseFG: I read on Chris comment on a different PR that there is still some unwrapped withItems around. You can ignore my comment.

frontend/src/store.ts
```diff
@@ -1,18 +1,62 @@
...
 
+const notificationError: ThunkMiddleware = ({ dispatch }) => next => action => {
+	const TOKEN_EXPIRED_MESSAGE = 'Given token not valid for any token type'
+
+	function getErrorDescription(error: any) {
+		/* HTTP errors are handled specially because they include the URL
+		 * in the error message and we don't want many very similar errors
+		 * to show up, just show one.
+		 */
+		if (typeof error.status === 'number')
```
ckostiw, line 20: I think for network errors it would be better if we use the URL as the notification ID, rather than the status code. For example, we might get 500 errors from multiple endpoints, and I think it would make sense to show a separate notification for each failed request, rather than replacing the alert each time a 500 occurs.

frontend/src/modules/notification/actions.tsx
```diff
@@ -0,0 +1,60 @@
...
+        type: NOTIFICATION_REMOVE,
+        id
+    })
+
+    // notification has onClose config which dispatches closeNotification.
+    // Closing after dispatch should prevent duplicate remove action.
+    notification.close(id)
+}
+
+export const showNotification = (
```
ckostiw, line 54: I'm not sure this action is a good idea. Mixing the title and ID might be a poor practice. It might be better to simply use `notify` to pop alerts. It might be nice to have `errorNotification`, `warningNotification`, `infoNotification`, `successNotification` actions as a short cut:
```
export const errorNotification = (id: NotificationID, title: NotificationProps['title'], description: NotificationProps['description']) {
    return notify(id, {
        title,
        description,
        type: NotificationType.ERROR
    })
}
```

frontend/src/store.ts
```diff
@@ -1,18 +1,62 @@
...
+	if (action.error.message.includes(TOKEN_EXPIRED_MESSAGE)) {
+		dispatch(logOut())
+		return next(action)
+	}
+
+	if (shouldIgnoreError(action))
+		return next(action)
+
+	const error = getErrorDescription(action.error)
+	dispatch(showNotification(error.message, error.details))
```
ckostiw, line 44: Might be better to just use `notify` here to keep it clear what is being used for an id, and that an error alert is being popped.

frontend/src/modules/notification/models.ts
```diff
@@ -0,0 +1,34 @@
...
+export enum NotificationType {
+    SUCCESS = 'SUCCESS',
+    INFO = 'INFO',
+    WARNING = 'WARNING',
+    ERROR = 'ERROR',
+}
+
+export interface NotificationProps {
+    type: NotificationType
+    description: ReactNode
```
ckostiw, line 16: Oops! You can't put ReactNodes in the redux store. I believe that you end up with an object containing a reference to the entire DOM and the redux serializer goes bananas trying to serialize the whole DOM. I think we will have to limit title and description to strings.

ckostiw (changes requested):

frontend/src/modules/notification/actions.tsx
nafiz1001: Hopefully you like my approach to writing those extra actions ;)

frontend/src/components/ExportButton.js
```diff
@@ -28,10 +30,10 @@ const ExportButton = ({ exportType, exportFunction, filename, itemsCount, ...res
               downloadFromFile(name, text)
             })
             .catch(err => {
-              notification.error({
-                message: err.message,
+              dispatch(notifyError(err.message, {
+                title: err.message,
                 description: <pre>{err.stack}</pre>,
```
nafiz1001, line 35: uh oh

frontend/src/modules/notification/models.ts
```diff
@@ -0,0 +1,35 @@
...
+export const NOTIFICATION_REMOVE = `${NOTIFICATION}.REMOVE`
+
+export enum NotificationType {
+    SUCCESS = 'SUCCESS',
+    INFO = 'INFO',
+    WARNING = 'WARNING',
+    ERROR = 'ERROR',
+}
+
+export type SerializableReactNode = string | number | boolean | null | undefined
```
ckostiw, line 14: Calling this "ReactNode" is confusing and asking for trouble.

ckostiw (changes requested):

ckostiw (approved):

PR #571, opened 2023-05-29
Feature #1661: Add advanded filter section to containers with Comment search

nafiz1001:
https://206.12.92.46/issues/1661

### DIFF
--- a/frontend/src/components/containers/ContainerFilters.tsx
+++ b/frontend/src/components/containers/ContainerFilters.tsx
@@ -20,7 +20,7 @@ const ContainerFilters = (rest: ContainerFiltersProps) => {
 
   return (
     <div className='SamplesFilters' {...rest}>
-      <Collapse defaultActiveKey={[]} ghost>
+      <Collapse defaultActiveKey={[]} ghost collapsible={"header"}>
         <Collapse.Panel
           header='Show advanced filters'
           key={0}
--- a/frontend/src/components/containers/ContainerFilters.tsx
+++ b/frontend/src/components/containers/ContainerFilters.tsx
@@ -1,0 +1,39 @@
+import React from "react";
+import {filter as filterObject} from "rambda";
+import {Collapse} from "antd";
+
+import {setFilter} from "../../modules/containers/actions";
+import FilterGroup from "../filters/filterGroup/FilterGroup";
+import {CONTAINER_FILTERS} from "../filters/descriptions";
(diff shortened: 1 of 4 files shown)


nafiz1001:
I forgot to fix the abort abuse.

frontend/src/components/containers/ContainerFilters.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type ContainerFiltersProps = Pick<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'style'>
+
+const ContainerFilters = (rest: ContainerFiltersProps) => {
+  const filters = useAppSelector((state) => state.containers.filters as FilterSetting[])
+  const dispatch = useAppDispatch()
+
+  const onChangeFilter = (filter: FilterDescription, value: FilterValue) => {
+    dispatch(setFilter(filter.key, value))
+  }
```
nafiz1001, line 19: use some sort of debounce to fix abort abuse.

ckostiw: DebouncedInput.jsx has a debounce example, and JumpBar.js too.

frontend/src/components/containers/ContainerFilters.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <div className='SamplesFilters' {...rest}>
+      <Collapse defaultActiveKey={[]} ghost>
```
ckostiw, line 23: It would be better if we use `collapsible="header"` so that the Collapse only expands if the user clicks on the 'Show advanced filters' text. I think that should also be fixed in the Samples table. I don't know how many times I've clicked in the white area at the top of the view and accidentally expanded the custom filters section...

ckostiw (changes requested):

ckostiw (changes requested):
Is there any way to search for multiple keywords in the comments, something like "tube, error, client" to match comments with all of those keywords. It's a bit tough if you can only filter on a single word, or a phrase that matches a substring of the comment exactly.

nafiz1001:
> Is there any way to search for multiple keywords in the comments, something like "tube, error, client" to match comments with all of those keywords. It's a bit tough if you can only filter on a single word, or a phrase that matches a substring of the comment exactly.

I have a feeling that will require playing with the backend, configure the input_filter do the splitting and searching. Or maybe create a new filter type 😨

PR #572, opened 2023-05-31
Individual Associated Samples

JMarce01:
- Adding `IndividualDetailContentTS.tsx`, `IndividualOverview.tsx`, `IndividualAssociatedSamples.tsx` to divide the individual section into tabs.
- Reworking `WorkflowSamplesTable.tsx` to take in a list of samples instead of a list of sampleID's so it doesn't rely on samples redux state. (also by removing the useEffect to read the redux state, it has been moved up a level to where it's needed)
- Adding a `reducer.ts` and `action.ts` for individualDetails. It contains an individual as well as individualSamples which is a `pagedItems` object.


ckostiw (changes requested):

frontend/src/modules/samples/actions.js
```diff
@@ -1,25 +1,25 @@
-import {createNetworkActionTypes, networkAction} from "../../utils/actions";
+import { createNetworkActionTypes, networkAction } from "../../utils/actions";
```
ckostiw, line 1: The changes in this file just seem to be formatting changes. Are there any code changes?

frontend/src/components/individuals/IndividualsDetailContentTS.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { useEffect, useState } from "react";
```
ckostiw, line 1: If we decide to port a component to TS then we just change the extension of the original JS file to `.tsx` and make whatever changes are needed for TS. You should trash the old js file and rename this file as "IndividualsDetailContent.tsx". I was using the "TS" naming convention for cases where I needed to keep the original JS code, and implement a new TS version with breaking changes. Otherwise, you just need to rename the file.

ckostiw (changes requested):

frontend/src/modules/individualDetails/reducers.ts
```diff
@@ -0,0 +1,102 @@
...
+export const SET_SORT_BY = 'INDIVIDUAL_DETAILS.SET_SORT_BY'
+export const SET_INDIVIDUAL_DETAILS_SAMPLES_FILTER = 'INDIVIDUAL_DETAILS.SET_INDIVIDUAL_SAMPLES_FILTER'
+export const CLEAR_FILTERS = "INDIVIDUAL_DETAILS.CLEAR_FILTERS"
+
+export interface Individual {
+    individual: FMSIndividual,
+    samplesByIndividual: PagedItems<FMSSample>
+}
+export interface IndividualDetailsState {
+    [key: number]: Readonly<FetchedState<Individual>>
```
ckostiw, line 21: Whenever I define a key map like this I add a comment to indicate what is expected as a key value (eg. an individual id, sample id, etc). Also, is the key an FMSId?

frontend/src/modules/samples/actions.js
JMarce01: No, did you want me to revert these changes? I must've accidentally formatted. Trying my best not to format existing pages. :/

ckostiw (changes requested):

frontend/src/modules/individualDetails/reducers.ts
```diff
@@ -0,0 +1,102 @@
...
+            }
+            break;
+        }
+        case SET_SORT_BY: {
+            const { individualID, sortBy } = action
+            const samplesByIndividual = state[individualID]?.data?.samplesByIndividual
+            if (samplesByIndividual) {
+                samplesByIndividual.sortBy = sortBy
+            }
+        }
```
ckostiw, line 97: Missing a `break` here

frontend/src/modules/individualDetails/reducers.ts
```diff
@@ -0,0 +1,102 @@
...
+import { PagedItems } from "../../models/paged_items";
+import { createItemsByID } from "../../models/frontend_models";
+import { clearFiltersReducer, setFilterReducer } from "../../components/shared/WorkflowSamplesTable/FilterReducers";
+
+export const LIST_TABLE = createNetworkActionTypes('INDIVIDUAL_DETAILS.LIST_TABLE')
+export const SET_SORT_BY = 'INDIVIDUAL_DETAILS.SET_SORT_BY'
+export const SET_INDIVIDUAL_DETAILS_SAMPLES_FILTER = 'INDIVIDUAL_DETAILS.SET_INDIVIDUAL_SAMPLES_FILTER'
+export const CLEAR_FILTERS = "INDIVIDUAL_DETAILS.CLEAR_FILTERS"
+
+export interface Individual {
```
ckostiw, line 16: There is already an Individual type defined in frontend_models.ts, so this could cause some confusion for the next developers to work in this code. I would give this type a unique name, such as IndividualDetails or IndividualWithSamples - anything unique.

frontend/src/modules/individualDetails/actions.ts
```diff
@@ -0,0 +1,54 @@
...
+			type: SET_SORT_BY,
+			individualID,
+			sortBy
+		})
+		dispatch(listTable(individualID))
+	}
+}
+export const listTable = (individualID: FMSId) => {
+	return async (dispatch, getState) => {
+		if (!getState().individuals.itemsByID[individualID]) {
```
ckostiw, line 35: Should use a selector here, `selectIndividualsByID(getState())`

frontend/src/modules/individualDetails/actions.ts
```diff
@@ -0,0 +1,54 @@
...
+			await dispatch(get(individualID));
+		}
+		const individual = getState().individuals.itemsByID[individualID]
```
ckostiw, line 38: Selector should be used here too. Also, the `individual` variable should be typed as an Individual (from frontend_models.ts).

ckostiw (changes requested):

frontend/src/components/individuals/IndividualsDetailContent.tsx
```diff
@@ -0,0 +1,82 @@
...
+            const individualInstance: FetchedState<Individual> = {
+                ...individualDetailsById[Number(id)]
+            }
+            if (individualInstance) {
+                setIndividual(individualInstance)
+            }
+        }
+    }, [individualDetailsById])
+
+    const tabsStyle = {
```
ckostiw, line 41: Nafiz changed some CSS and we don't need to set the `tabsStyle` and `tabPanelStyle` styles on our Tab and TabPane components anymore, so you can trash this.

ckostiw (changes requested):

frontend/src/modules/samples/actions.js
ckostiw: If you can revert this file, sure. 👍

frontend/src/modules/individualDetails/reducers.ts
```diff
@@ -49,7 +49,7 @@ const individualDetailsReducer = (state: WritableDraft<IndividualDetailsById>, a
             const { individualID } = action.meta
             if (!state[individualID]) {
                 state[individualID] = {
-                    individual: undefined,
+                    individual: preprocess(undefined),
```
ckostiw, line 52: Better not use preprocess here, because it sets the `loaded` property to `true`, but the individual hasn't been received from the backend yet, so it's not loaded. What you want to is an object like this: 
```
{
  id: <individual id goes here>,
  isFetching: true,
  isLoaded: false
}
```

ckostiw (changes requested):

ckostiw (changes requested):
Could you fix this import error as part of the PR? It is currently in master.
```
src/components/taxons/EditTaxon.tsx:10:10 - error TS2305: Module '"./TaxonTableColumns"' has no exported member 'ObjectWithTaxon'.

10 import { ObjectWithTaxon } from "./TaxonTableColumns";
            ~~~~~~~~~~~~~~~
```

frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -38,6 +39,31 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
 	const SAMPLES_TAB_KEY = 'samples'
 	const SELECTION_TAB_KEY = 'selection'
 	const [selectedTab, setSelectedTab] = useState<string>(SAMPLES_TAB_KEY)
+	const samplesByID = useAppSelector(selectSamplesByID)
+	const librariesByID = useAppSelector(selectLibrariesByID)
+	const [samples, setSamples] = useState<SampleAndLibrary[]>([])
+	const [selectedSamples, setSelectedSamples] = useState<SampleAndLibrary[]>([])
+
+	const getSampleList = (sampleIDs) => {
```
ckostiw, line 47: You should put this function inside the useEffect that is calling it, since it isn't called anywhere else. I am getting a linter error because `getSampleList` is not listed in the dependencies array of the useEffect. Moving the function definition inside the useEffect will solve that.

PR #573, opened 2023-05-31
Modified template submission error alert

ckostiw:
If the user modifies a template after it has been uploaded then when they try to submit a 'failed to fetch' error is thrown. Catch the error and pop an error notification to explain why submit has failed.

### DIFF
--- a/frontend/src/components/templateFlow/TemplateFlow.js
+++ b/frontend/src/components/templateFlow/TemplateFlow.js
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import { connect } from "react-redux"
 import PropTypes from "prop-types";
-import { Button, Steps, Row, Col } from "antd";
+import { Button, Steps, Row, Col, notification } from "antd";
 import { downloadFromFile } from "../../utils/download";
 
 import {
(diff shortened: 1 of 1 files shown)


ckostiw:
Screen Shot 2023-05-31 at 2 03 38 PM

frontend/src/components/templateFlow/TemplateFlow.js
```diff
@@ -87,6 +87,21 @@ const TemplateFlow = ({ fetchListedData, fetchSummariesData, ...props }) => {
...
+        if (error.message === 'Failed to fetch') {
+          // If the template file was changed on disk since it was uploaded
+          // then a 'failed to fetch' error is thrown. This can happen if the user saves a change to
+          // the template after it has been uploaded.
+          const description = 
+            <div>
+              <p>The template has changed since it was uploaded and cannot be submitted.</p>
+              <p>Please upload the template again, and ensure that it is not saved or otherwise modified until after it has been submitted.</p>
+            </div>
+          notification.error({
```
nafiz1001, line 99: you should've used my brand new api 👎🏼 

but otherwise it looks fine and i can add it myself later.

PR #574, opened 2023-05-31
Edit def. auth + Individuals associated to Taxon

JMarce01:
- adding authentication for edit Taxons and reference genomes. (checks if logged_user is a superuser to allow them to edit Taxons and reference genomes)
- changing Taxon serializer to append boolean to FMSTaxon, to see if it's ncbi_id can be edited. (checks if there are existing individuals associated to a certain Taxon)

### DIFF
--- a/backend/fms_core/serializers.py
+++ b/backend/fms_core/serializers.py
@@ -221,13 +221,13 @@ class RunTypeSerializer(serializers.ModelSerializer):
 
 
 class TaxonSerializer(serializers.ModelSerializer):
-    edit = serializers.SerializerMethodField()
+    editable = serializers.SerializerMethodField()
     class Meta:
         model = Taxon
         fields = "__all__"
-        extra_fields = ("edit")
-    def get_edit(selft, obj):
-        return Individual.objects.filter(taxon_id=obj.id).count() == 0
+        extra_fields = ("editable")
+    def get_editable(selft, obj):
+        return Individual.objects.filter(taxon_id=obj.id).exists()
 
 
 class IndividualSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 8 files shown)


frontend/src/components/taxons/EditTaxon.tsx
```diff
@@ -19,10 +22,14 @@ export const EditTaxonRoute = () => {
     const taxons = useAppSelector(selectTaxonsByID)
     const { id } = useParams();
     const appInitialzed = useAppSelector(selectAppInitialzed)
-    return (id && taxons[id] && appInitialzed) ? <EditTaxon taxon={{ ...taxons[id] }} /> : null
+    const authState = useAppSelector(selectAuthState)
+    const users = useAppSelector(selectUsersByID)
+    const isAuthenticated = (authState.currentUserID ? users[authState.currentUserID].is_superuser : false);
```
UlysseFG, line 27: Would rename isAuthenticated to hasWritePermission or something else to more specifically define the variable use.

backend/fms_core/serializers.py
```diff
@@ -221,9 +221,13 @@ class Meta:
 
 
 class TaxonSerializer(serializers.ModelSerializer):
+    edit = serializers.SerializerMethodField()
```
UlysseFG, line 224: use editable instead, it's a bit clearer what it indicate.

ckostiw:
If I reload the taxon list page the application crashes and I get the white page of death.
Screen Shot 2023-06-02 at 3 15 06 PM

ckostiw:
I think the 'editable' property is backwards. It is true for taxons that are in use (homo sapiens), and false for taxons that have no individuals (mus musculus). Did I misunderstand this field?
Screen Shot 2023-06-02 at 3 18 29 PM

PR #579, opened 2023-06-05
Optimizing jumpbar 

JMarce01:
adding async to jumpbar fetch and moving setIsFetching

### DIFF
--- a/frontend/src/components/JumpBar.js
+++ b/frontend/src/components/JumpBar.js
@@ -44,17 +44,20 @@ const JumpBar = (props) => {
   const history = useNavigate();
   const selectRef = useRef();
 
-  const search = useMemo(() => debounce(150, query => {
+  const search = useMemo(() => debounce(150, async query => {
     setValue(null)
-    setIsFetching(true)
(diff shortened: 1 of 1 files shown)


ckostiw (approved):

frontend/src/components/JumpBar.js
```diff
@@ -44,17 +44,19 @@ const JumpBar = (props) => {
...
-    setIsFetching(true)
-    withToken(token, api.query.search)(query)
-      .then(response => { setItems(response.data) })
-      .catch(err => {
-        if (err.name === 'AbortError')
-          return
-        setError(err.message)
-      })
-      .then(() => setIsFetching(false))
+  const search = useMemo(() => debounce(150, async query => {
```
ckostiw, line 47: No matter how fast I type, a call to the backend is being triggered on every character I type. 150ms for the debounce time is too small. Let's change it to 500ms.

ckostiw (changes requested):

ckostiw (approved):

PR #581, opened 2023-06-06
Feature #1606: Close table column input filter boxes on 'Return'

nafiz1001:
https://206.12.92.46/issues/1606


ckostiw (approved):
Looks fine

PR #584, opened 2023-06-07
api - fix redirects

ckostiw:
Add trailing `/` characters to api calls to avoid generating redirects. If we send a request to django without a trailing `/` then django returns a redirect to a url containing the missing `/` character (eg `http://localhost:9000/api/samples/list?...` gets redirected to `http://localhost:9000/api/samples/list/...`).

Fixes an error with fetching study samples, which can generating a long URL because the URL contains the ID's of all of the samples being fetched. The `api.samples.list` function was missing the trailing `/` character. Django is configured to handle long URL's but apparently it cannot handle long redirect URLs, and a 502 error was generated.


ckostiw:
Examples of redirects in freezeman:

Screen Shot 2023-06-07 at 1 08 52 PM

nafiz1001:
So I decided to grep to see if there are any links with missing slash end, and here's what I found.
```
nafiz@localhost:~/src/freezeman/frontend/src/utils> grep -n '"/[^/"]*"' api.js 
20:    list: (options, abort) => get("/containers", options, { abort }),
82:    list: (options, abort) => get("/indices", options, { abort }),
111:    list: (options, abort) => get("/libraries", options, { abort }),
163:    list: (options, abort) => get("/projects", options, { abort }),
193:    list: (options, abort) => get("/samples", options, { abort }),
278:    list: (options, abort) => get("/users", options, { abort }),
289:    list: (options, abort) => get("/groups", options, { abort }),
```
Not sure if it's worth adding slashes to those.

PR #596, opened 2023-06-14
403 error on users api call, forced logout

JMarce01:
- changed expiration time of refresh token
- after reading docs, adding 'simplejwt' to 'installed_apps' definition in `settings.py` to make sure that the expiration is properly set, not entirely sure if it does anything but in all setup docs they say to add it.


backend/fms/settings.py
```diff
@@ -265,7 +266,7 @@
 # Mechanism in order to automatically logout the user after 4 hours
 SIMPLE_JWT = {
 'ACCESS_TOKEN_LIFETIME': timedelta(hours=4),
-'REFRESH_TOKEN_LIFETIME': timedelta(minutes=10),
+'REFRESH_TOKEN_LIFETIME': timedelta(days=10),
```
UlysseFG, line 269: I think this was to enforce the user to show sign of life every 10 minutes. The 10 days refresh token make it useless. I would prefer you find another fix that one that breaks the intended functionality.

ckostiw: I think it was configured backwards originally, according to this [documentation](https://django-rest-framework-simplejwt.readthedocs.io/en/latest/settings.html) 

The refresh token is supposed to live longer than the access token. The access token should have a short lifetime (eg. 10 minutes) and the refresh token a longer one (eg. 1 day). The token will be refreshed until the refresh token expires after 1 day. Setting it to one day will force users to login each day. Setting it to 7 days would force them to refresh only once a week. I'm not sure what policy we want in the lab.

ckostiw: To test this, you can set the access token lifetime to 1 minute, and the refresh token lifetime to maybe 10 minutes and verify that the token gets refreshed every minute. You will have to delete any existing tokens you might have by going in chrome devtools -> Application and deleting the token stored under 'persist:auth'

ckostiw: Also, in `frontend/src/modules/auth/actions.js`, the `refreshAuthToken` action checks the expiry date of the access token and if it still has more than 30 seconds of lifetime left then it skips the refresh. The problem is that refresh gets called every 30 seconds (as a side effect of refreshing summaries every 30 seconds). If the access token actually has 30.001 seconds left, and it takes 30.9 seconds before the loop runs (since the intervals are never precise in JS) the token will expire before we can refresh and it's possible that freezeman will make an api call during this period, the call with fail with an auth error, and the user will get bounced into the login page. This has happened to me a few times and I've always wondered why it seems to happen rarely and randomly.

I think we should refresh if the access token has less than 60 seconds of life left.

ckostiw (changes requested):

backend/fms/settings.py
UlysseFG: I like the idea of preventing the summaries from keeping people logged in artificially... Since the mechanism is hidden to the user i would rather prevent the summaries from refreshing the auth token than use an invisible countdown that will likely frustrate the users. That way we could use a generous duration for the refresh to kick in.

ckostiw (approved):

UlysseFG (approved):

PR #598, opened 2023-06-16
Feature #1750:  Group Similar Warnings for Template Submission

nafiz1001:
https://206.12.92.46/issues/1750


nafiz1001:
image
image

ckostiw:
I think if that there are more than 10 rows with errors then we should display something like: 
"(Row # 11, 12, 13, 14, 15, 16, 17, 18, 19, 20... (+231 others)"

The lab's templates will usually have at least a hundred sample, and I'm pretty sure there will be a lot of cases where the same warning is applied to every sample.

Screen Shot 2023-07-04 at 3 04 52 PM

nafiz1001:
image
image
image

UlysseFG:
The index validation tool warnings are not updated to the new format...

UlysseFG (approved):

PR #599, opened 2023-06-21
adding indexes to optimize template submission

JMarce01:
Adding indexes to models to improve template submission, major improvements:
- sample_submission average 10.10 sec. -> 5.34 sec.

Note: 
- to improve indexing (filtering with case insensitive, newer version of django is required)

indexes:
- Container: 'coordinate', 'barcode', 'name'
- Coordinate: 'name'
-  Individual: 'name'
- Instrument: 'type'
- Project: 'name'
- SampleNextStep: 'step+sample'
- SampleNextStepByStudy: 'sample_next_step+step_order+study'
- StepOrder: 'order+workflow'


UlysseFG:
I think we would need to add a couple of common usage indexes
- Sample -> name
- Sample -> creation-date
- ...

Create your migration file for these indices (`python manage.py makemigrations` should be able to generate the code you can rename the file generated to fit our naming.).

The index you created make sense. We'll have to see if we need more as we go but we can get started with these.

Don't forget to remove your timing code once you are done ;).

backend/fms_core/migrations/0054_v4_3_0.py
```diff
@@ -0,0 +1,57 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('fms_core', '0053_v4_3_0'),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='container',
+            index=models.Index(fields=['coordinate'], name='fms_core_co_coordin_27d459_idx'),
```
UlysseFG, line 15: I would recommend using human readable names for your indices... You can use a similar pattern than i use for unique keys (TableName_Field1_Field2_..._key) replacing key by idx for index... It will make it easier to understand traces and logs that might come up if the naming is clear (the default naming is truncated which fails to carry the information we want).

UlysseFG (approved):

PR #600, opened 2023-06-21
Remove redundant request for template actions in Dashboard

ckostiw:
Possible fix for broken action template download links in QC when connecting through an ssh tunnel.

We load the template actions when the app launches, but Dashboard was firing off it's own request to load the template actions. It's possible that this causes the strangeness on QC.

The second request for template actions by the dashboard is redundant and has been removed.

"Template actions url's are broken on QC"
https://206.12.92.46/issues/1765


JMarce01 (approved):

UlysseFG:
You can merge it if it is approved.

PR #604, opened 2023-06-27
Feature #1465: Study Samples View: Display number of samples in pool

nafiz1001:
https://206.12.92.46/issues/1465

It was a while ago that I worked on this. So I'm not sure if it's working or up to date.


JMarce01 (approved):

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
```diff
@@ -99,12 +99,24 @@ function StudyStepSamplesTable({ studyID, step, settings }: StudyStepSamplesTabl
 			// is already in the project details page.
 			return [
 				...getColumnsForStudySamplesStep(stepDefinition, protocol),
-				actionColumn,
-			]
+				{
+					columnID: 'SAMPLE_COUNT',
+					title: 'Sample Count (as Pool)',
```
ckostiw, line 104:
```suggestion
					title: 'Sample Count (in Pool)',
```

frontend/src/modules/studySamples/services.ts
```diff
@@ -125,6 +125,10 @@ export async function buildStudySamplesFromWorkflow(
 			}
 			const samples = sampleNextSteps.map(nextStep => nextStep.sample)
 
+			const sampleCountByPooledSampleID =
```
ckostiw, line 128: We should avoid calling this endpoint if there are zero samples at the current step (happens often).

ckostiw: Also, it's a shame that we are forced to call the endpoint even if there are no pooled samples in the list (since we don't know if the sample is pooled or not).
I know the story was about displaying the number of samples in a pool for a study, but maybe we might want to display the same thing elsewhere. I wonder if it would be better to add a new field to Sample, `pooledSampleCount` and have the samples endpoint return a value in that field if the sample is a pool (and undefined/null if not).

ckostiw (changes requested):

frontend/src/modules/studySamples/services.ts
UlysseFG: We could replace is_pool by an integer pooled_count (or other name) (1 -> x) and provide an is_pool function in the frontend to convert 1 to false and anything more to true... We currently use a count to test for pool in the endpoint but only return a boolean.

backend/fms_core/serializers.py
```diff
@@ -422,40 +422,47 @@ class Meta:
...
                   'concentration', 'concentration_nm', 'quantity_ng', 'creation_date', 'quality_flag',
-                  'quantity_flag', 'project', 'depleted', 'library_type', 'platform', 'index', 'library_size')
+                  'quantity_flag', 'project', 'depleted', 'library_type', 'platform', 'index', 'library_size', 'derived_samples_counts')
 
 
 class LibraryExportSerializer(serializers.Serializer):
     coordinates = serializers.CharField(read_only=True, source="coordinate.name")
     library_size = serializers.DecimalField(max_digits=20, decimal_places=0, read_only=True, source="fragment_size")
+    derived_samples_counts = serializers.IntegerField(read_only=True, source="count_derived_samples")
+
```
UlysseFG, line 461: Maybe put this one next to is_pool ?

backend/fms_core/serializers.py
```diff
@@ -422,40 +422,47 @@ class Meta:
...
 class SampleSerializer(serializers.Serializer):
+    derived_samples_counts = serializers.IntegerField(read_only=True, source="count_derived_samples")
+
     class Meta:
         fields = ('id', 'biosample_id', 'name', 'alias', 'volume', 'depleted', 'concentration', 'child_of',
                   'extracted_from', 'individual', 'container', 'coordinate', 'sample_kind', 'is_library', 'is_pool', 'project',
                   'process_measurements', 'tissue_source', 'creation_date', 'collection_site', 'experimental_group',
                   'quality_flag', 'quantity_flag', 'created_by', 'created_at', 'updated_by', 'updated_at', 'deleted', 
-                  'comment')
+                  'comment', 'derived_samples_counts')
```
UlysseFG, line 432: Next to is_pool would be a nice place ...

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
```diff
@@ -99,8 +99,20 @@ function StudyStepSamplesTable({ studyID, step, settings }: StudyStepSamplesTabl
...
-			]
+				{
+					columnID: 'SAMPLE_COUNT',
+					title: 'Sample Count (in Pool)',
+					dataIndex: ['sample', 'id'],
+					render: (_, { sample }: SampleAndLibrary) => {
+						return (
+								sample && sample.is_pool
+								? sample.derived_samples_count
+								: '-'
```
ckostiw, line 110: I'd prefer it if we just render an empty string, rather than "-" to be consistent with the rest of the UX.

backend/fms_core/serializers.py
```diff
@@ -422,38 +422,45 @@ class Meta:
...
     class Meta:
         fields = ('id', 'biosample_id', 'name', 'alias', 'volume', 'depleted', 'concentration', 'child_of',
-                  'extracted_from', 'individual', 'container', 'coordinate', 'sample_kind', 'is_library', 'is_pool', 'project',
-                  'process_measurements', 'tissue_source', 'creation_date', 'collection_site', 'experimental_group',
-                  'quality_flag', 'quantity_flag', 'created_by', 'created_at', 'updated_by', 'updated_at', 'deleted', 
-                  'comment')
+                  'extracted_from', 'individual', 'container', 'coordinate', 'sample_kind', 'is_library', 'is_pool',
+                  'derived_samples_counts', 'project', 'process_measurements', 'tissue_source', 'creation_date',
+                  'collection_site', 'experimental_group', 'quality_flag', 'quantity_flag', 'created_by', 'created_at',
+                  'updated_by', 'updated_at', 'deleted',  'comment')
```
UlysseFG, line 432: 2 spaces before comment

ckostiw (approved):

UlysseFG:
Made a ready to sequence pool, with 11 samples ... the sample count (in pool) column tells me 121 ... You'll have to get closer to the mark.

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
```diff
@@ -99,8 +99,20 @@ function StudyStepSamplesTable({ studyID, step, settings }: StudyStepSamplesTabl
 			// is already in the project details page.
 			return [
 				...getColumnsForStudySamplesStep(stepDefinition, protocol),
-				actionColumn,
-			]
+				{
+					columnID: 'SAMPLE_COUNT',
+					title: 'Sample Count (in Pool)',
```
UlysseFG, line 104: "Samples in pool" would be simpler and more readable

UlysseFG (changes requested):
The count is not right.

backend/fms_core/viewsets/_utils.py
```diff
@@ -29,6 +29,13 @@ def _prefix_keys(prefix: str, d: Dict[str, Any]) -> Dict[str, Any]:
 def _list_keys(d: Dict[str, Any]) -> Dict[str, Any]:
     return [k  for k, v in d.items()]
 
+def fix_count_derived_samples(queryset: QuerySet) -> QuerySet:
+    # https://stackoverflow.com/a/69031027
+    return queryset.annotate(count_derived_samples=Subquery(
+        DerivedBySample.objects
+            .filter(sample=OuterRef("pk"))
+            .annotate(my_count=Func(F('id'), function='Count'))
```
UlysseFG, line 37: why not just Count("id") ? i think it actually does the same as what you are doing.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -9,6 +9,7 @@
 from fms_core.services.library import convert_library_concentration_from_ngbyul_to_nm
 
 from ..utils import decimal_rounded_to_precision
+from ._utils import fix_count_derived_samples
```
UlysseFG, line 12: do not call this fix_ ... just name the variable to describe what it really is (count_derived_by_sample).

UlysseFG (approved):

PR #605, opened 2023-06-28
Bug #1763 + Feature #1762: Fix ref Genome form and taxon form

JMarce01:
- fixing check for disable field on taxon form,
- adding editable field on ref genome to see if it's associated with an individual.


ckostiw (changes requested):
I'm getting a 500 error when I try to update the Assembly Name of a reference genome. I'm changing `ASM886v2` to `ASM886v2CK`.

Screen Shot 2023-07-05 at 9 56 31 AM

backend/fms_core/serializers.py
```diff
@@ -226,7 +226,7 @@ class Meta:
         model = Taxon
         fields = "__all__"
         extra_fields = ("editable")
-    def get_editable(selft, obj):
+    def get_editable(self, obj):
         return not Individual.objects.filter(taxon_id=obj.id).exists()
```
UlysseFG, line 230: This would be cleaner to create a can_edit_taxon service in ../service/taxon.py to make sure we always have the same version of this test. This would include a small unit test also for the service.

backend/fms_core/serializers.py
```diff
@@ -784,11 +784,15 @@ def get_steps_order(self, instance):
...
+    taxon_id = serializers.SerializerMethodField()
     class Meta:
         model = ReferenceGenome
-        fields = ("id", "assembly_name", "synonym", "genbank_id", "refseq_id", "taxon_id", "size", "editable")
-    def get_editable(selft, obj):
-        return not Individual.objects.filter(reference_genome_id=obj.id).exists()
+        fields = "__all__"
+        extra_fields = ("taxon_id", "editable")
+    def get_editable(self, obj):
+        return not Individual.objects.filter(reference_genome_id = obj.id).exists()
```
UlysseFG, line 793: This would be cleaner to create a can_edit_reference_genome service in ../services/reference_genome.py to make sure we always have the same version of this test. This would include a small unit test also for the service(.../tests/test_services/reference_genome_test.py).

backend/fms_core/services/taxon.py
```diff
@@ -0,0 +1,5 @@
+from fms_core.models import Individual
+
+
+def can_edit_taxon(id):
```
UlysseFG, line 4: Service functions have docstrings to document them.

backend/fms_core/services/referenceGenome.py
```diff
@@ -0,0 +1,5 @@
+from fms_core.models import Individual
+
+
+def can_edit_referenceGenome(id):
```
UlysseFG, line 4: Service functions have docstrings to document them.

backend/fms_core/serializers.py
```diff
@@ -783,16 +785,14 @@ def get_steps_order(self, instance):
         return serialized_data.data
 
 class ReferenceGenomeSerializer(serializers.ModelSerializer):
-    editable = serializers.SerializerMethodField()
-    taxon_id = serializers.SerializerMethodField()
+    editable = serializers.BooleanField(read_only=True)
```
UlysseFG, line 788: I might be wrong but i think you need the SerializerMethodField() to call a method for a Field.

backend/fms_core/tests/test_services/taxon_test.py
```diff
@@ -0,0 +1,9 @@
+from django.test import TestCase
+
+from fms_core.services.taxon import can_edit_taxon
+
+class TaxonServicesTestCase(TestCase):
+    def setup(self) -> None:
+        ex = 1
+    def test_get_individual(self):
+        editable = can_edit_taxon()
```
UlysseFG, line 9: Similar to above ...

backend/fms_core/services/referenceGenome.py
```diff
@@ -0,0 +1,5 @@
+from fms_core.models import Individual
+
+# returns a boolean if an individual is referencing this reference genome
```
UlysseFG, line 3: Use the same format found in other services. The automatic doc generator rely on that format.

backend/fms_core/services/taxon.py
```diff
@@ -0,0 +1,5 @@
+from fms_core.models import Individual
+
+# returns a boolean if an individual is referencing this taxon
```
UlysseFG, line 3: Use the same format found in other services. The automatic doc generator rely on that format.

backend/fms_core/serializers.py
```diff
@@ -221,13 +223,13 @@ class Meta:
 
 
 class TaxonSerializer(serializers.ModelSerializer):
-    editable = serializers.SerializerMethodField()
+    editable = serializers.BooleanField(read_only=True)
     class Meta:
         model = Taxon
         fields = "__all__"
         extra_fields = ("editable")
```
UlysseFG, line 230: This extra_field is not present in the response... you likely need to add this field to the fields instead of extra_fields ...

backend/fms_core/serializers.py
```diff
@@ -783,9 +785,14 @@ def get_steps_order(self, instance):
         return serialized_data.data
 
 class ReferenceGenomeSerializer(serializers.ModelSerializer):
+    editable = serializers.BooleanField(read_only=True)
+    taxon_id = serializers.IntegerField(read_only=True, source='taxon.id')
     class Meta:
         model = ReferenceGenome
-        fields = ("id", "assembly_name", "synonym", "genbank_id", "refseq_id", "taxon_id", "size")
+        fields = "__all__"
+        extra_fields = ("taxon_id", "editable")
```
UlysseFG, line 793: Taxon id is present twice once as taxon and once as taxon_id editable is not present ... This would suggest the extra_field are returned but editable is a BooleanField() not a SerializerMethodField() (the boolean field is tied to no value).

ckostiw (approved):

UlysseFG (approved):

PR #607, opened 2023-07-06
Add Readset api

nafiz1001:


backend/fms_core/filters.py
```diff
@@ -213,4 +215,15 @@ def quantity_ng_gte_filter(self, queryset, name, value):
 
     class Meta:
         model = SampleNextStep
-        fields = _sample_next_step_filterset_fields
\ No newline at end of file
+        fields = _sample_next_step_filterset_fields
+
+class ReadsetFilter(GenericFilter):
+    id__not_in = django_filters.CharFilter(field_name='id', method='filter_id_not_in')
```
UlysseFG, line 221: This is an oddly specific filter. What is the use case related to this ?

nafiz1001: @ckostiw can elaborate on this

But the reason why i've introduced it is so that we can avoid pulling in hundreds of readsets that the users don't want to see, and the Readset viewset for some reason did not support id__not_in

nafiz1001:
This is for triton's sake.

backend/fms_core/viewsets/_constants.py
```diff
@@ -249,6 +249,13 @@
     "validation_status_timestamp": DATE_FILTERS,
 }
 
+_readset_filterset_fields: FiltersetFields = {
+    **_prefix_keys("dataset__", _dataset_filterset_fields),
```
UlysseFG, line 253: Put the model own fields first in the filterset. Move this at the end of the list ...

backend/fms_core/filters.py
```diff
@@ -213,4 +215,10 @@ def quantity_ng_gte_filter(self, queryset, name, value):
 
     class Meta:
         model = SampleNextStep
-        fields = _sample_next_step_filterset_fields
\ No newline at end of file
+        fields = _sample_next_step_filterset_fields
+
+class ReadsetFilter(GenericFilter):
```
UlysseFG, line 220: You don't need to define a ReadsetFilter if it has no specific filter ...

backend/fms_core/viewsets/readset.py
```diff
@@ -0,0 +1,22 @@
...
+class ReadsetViewSet(viewsets.ModelViewSet):
+    queryset = Readset.objects.all()
+
+    serializer_class = ReadsetSerializer
+
+    ordering_fields = (
+        *_list_keys(_readset_filterset_fields),
+    )
+
+    filterset_class = ReadsetFilter
```
UlysseFG, line 22: define filterset_fields here instead of the the filterset_class unless you need specific filters.

backend/fms_core/viewsets/_constants.py
```diff
@@ -249,6 +249,13 @@
     "validation_status_timestamp": DATE_FILTERS,
 }
 
+_readset_filterset_fields: FiltersetFields = {
+    "id" : PK_FILTERS,
+    "name": CATEGORICAL_FILTERS_LOOSE,
+    "sample_name": CATEGORICAL_FILTERS_LOOSE,
```
UlysseFG, line 255: you should include derived_sample as FK_FILTERS since it is part of the serialized data.

UlysseFG (commented):
You should have Chris review this too to make sure it has all you guys need for Triton. Otherwise it looks fine to me.

ckostiw (approved):

PR #611, opened 2023-07-18
Rename run info fields

UlysseFG:


backend/fms_core/schema_validators.py
```diff
@@ -122,9 +122,10 @@ def deconstruct(self):
                             "items": {
                                 "type": "object",
                                 "properties": {
+                                    "external_project_id": {"type": "string"},
                                     "PROJECT": {"type": "string"},
                                 },
-                                "required": ["PROJECT"]
+                                "required": ["external_project_id", "PROJECT"]
```
nafiz1001, line 128: we're gonna expect datasets to include external_project_id field?

UlysseFG: yeah the external_project_id should always be there ... it is not currently for Clarity runs (or even for most of ours, but we'll change the project required fields in the frontend to enforce that)

nafiz1001 (approved):
I don't see anything missing or wrong so far ;)
Tests passed too.

PR #612, opened 2023-07-19
changing index name for samplenextstepbystudy

JMarce01:
- renaming index for SampleNextStepbyStudy in models, and migrations file


ckostiw (approved):
That fixed the migration and server startup issues for me.

PR #613, opened 2023-07-20
Better DatasetFile serializer

nafiz1001:


backend/fms_core/serializers.py
```diff
@@ -591,12 +593,22 @@ def get_blocked_status_count(self, obj):
...
 
+class ReadsetSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Readset
+        fields = ("id", "name", "dataset", "sample_name", "derived_sample")
+
 class DatasetFileSerializer(serializers.ModelSerializer):
-    dataset = serializers.IntegerField(read_only=True, source='readset.dataset.id')
-    sample_name = serializers.CharField(read_only=True, source='readset.sample_name')
+    readset = serializers.SerializerMethodField()
```
UlysseFG, line 602: you can just plug in readset = ReadsetSerializer(read_only=True) here i think instead of using a method.

frontend/src/models/fms_api_models.ts
```diff
@@ -78,11 +78,15 @@ export interface FMSDataset extends FMSTrackedModel {
     metric_report_url?: string          // An external url to a report containing metrics for the dataset run
 }
 
+export interface FMSReadset extends FMSTrackedModel {
+    sample_name: string                // The name of the sample that was processed to produce this readset
```
UlysseFG, line 82: You should still include in this model the rest of the Readset content even if you don't use it.

UlysseFG (approved):

PR #614, opened 2023-07-20
New index on sample name and removal of validation to increase speed

UlysseFG:


ckostiw (approved):
Sample submission is twice as fast as before.

PR #616, opened 2023-07-20
Fix missing sidebar

ckostiw:
Fixes an issue where the sidebar would mysteriously disappear from freezeman.

App.js only renders the sidebar if it finds a user ID in the auth state in redux. If there is no user ID then it skips the sidebar, but still renders the dashboard (which makes no sense - we should fix that someday).

Normally if there is no user ID then there will be no tokens either and users are bounced to the login page.

Somehow, we were ending up with a null user ID but valid tokens in the auth state, so most of the app functioned normally but the sidebar wouldn't render.

I think this was caused by a race condition. When the app starts up it checks any tokens it has (loaded from local storage into redux by the persist framework). If the access token is expired but the refresh token is still valid then it will use the refresh token to request a new access token. 

At the same time, App.js would attempt to load the current user object with the user `get` action, but with either the old expired token, or no token at all (not 100% sure), and this would result in an error response from the server.

The new notifications middleware checks all actions for errors. If it discovers a token error then it dispatches the logOut action automatically.

logOut flushes the auth state, setting the current user id to null, and flushing the tokens.

I think what happened was that there was a pending refresh api call that was completed after the auth state was flushed. The refresh was successful and the auth reducer ended up writing the new tokens into the auth state. Refresh only writes new tokens to the auth state, and keeps whatever user id was already in the state.

We end up with a null user id and the set of tokens, and freezeman falls apart. Since this gets written to local storage, the user ends up stuck with no sidebar until the tokens expire.

The fix is to stop trying to load the current user with `get` and simply wait for the app to finish loading the list of users into redux. Eventually the user will be loaded, the app will re-render, and all will be well.

### DIFF
--- a/frontend/src/components/App.js
+++ b/frontend/src/components/App.js
@@ -34,14 +34,14 @@ import useUserInputExpiration from "../utils/useUserInputExpiration";
 import { useAppDispatch, useAppSelector } from "../hooks";
 import { setAppInitialized } from "../modules/app/actions";
 import { logOut } from "../modules/auth/actions";
-import { fetchLabworkSummary, fetchListedData, fetchStaticData, fetchSummariesData } from "../modules/shared/actions";
+import { fetchSummariesData, fetchStaticData, fetchLabworkSummary, fetchListedData } from "../modules/shared/actions";
 import { get } from "../modules/users/actions";
 import { selectAppInitialzed, selectAuthTokenAccess, } from "../selectors";
 import DatasetsPage from "./datasets/DatasetsPage";
 import LabworkPage from "./labwork/LabworkPage";
+import WorkflowDefinitionsRoute from "./workflows/WorkflowDefinitionsRoute";
 import ReferenceGenomesRoute from "./referenceGenomes/ReferenceGenomesRoute";
 import TaxonsRoute from "./taxons/TaxonsRoute";
-import WorkflowDefinitionsRoute from "./workflows/WorkflowDefinitionsRoute";
 
 
 const { Title } = Typography;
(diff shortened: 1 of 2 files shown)


ckostiw:
I added a console error message to the store middleware so that we will know if the store middleware is kicking the users to the login page, in case we run into other problems with this mechanism.

ckostiw:
To test the refresh stuff, you can edit `settings.py` and set the access token time to 1 minute and the refresh token time to 2 minutes, then see how the app behaves if the access token expires but the refresh token is still valid, and also what happens if both tokens are expired.

frontend/src/components/App.js
```diff
@@ -204,12 +204,6 @@ const App = ({userID, usersByID, logOut, get}) => {
   const isLoggedIn = userID !== null;
   const user = usersByID[userID];
 
-  useEffect(() => {
-    if (!user && isLoggedIn) {
-      get(userID)
```
ckostiw, line 209: This was also causing the irritating 403 error :)

nafiz1001:
> Refresh only writes new tokens to the auth state, and keeps whatever user id was already in the state.

If it contains user ID, we could also overwrite user ID with what is in the refresh token. Is there a problem with that kind of solution?

nafiz1001:
I tested this with:
```python
'ACCESS_TOKEN_LIFETIME': timedelta(seconds=1),
'REFRESH_TOKEN_LIFETIME': timedelta(seconds=10),
```

And I not only reproduced the missing sidebar, but it would also rapid fire API calls.

nafiz1001 (approved):
I'm not quite sure how to properly test this on Firefox. But it at least still works ;)

PR #617, opened 2023-07-24
Project Form changes.

UlysseFG:


frontend/src/components/projects/ProjectEditContent.tsx
nafiz1001, line 1: Maybe introduce type definitions for the `serialize` and `deserialize` functions that are at the bottom of the code?

frontend/src/models/empty_models.ts
nafiz1001, line 1: You might be able to type those empty models with interfaces in frontend/src/models/frontend_models.ts or frontend/src/models/fms_api_models.ts

UlysseFG: I think i'll let this to Chris ... He is the one that pushes for typing and i really don't want to get in his path on that by creating ad hoc types just to make the damn thing stop complaining ...

frontend/src/components/containers/ContainerEditContent.js
```diff
@@ -132,10 +132,13 @@ const ContainerEditContent = ({ token, containerKinds, containersByID, add, upda
...
         >
-          <Item label="Name" {...props("name")} rules={nameRules.concat(requiredRules)}>
+          <Item label="Name" {...props("name")} rules={nameRules.concat(requiredRules)}
+            tooltip="Use [a-z], [A-Z], [0-9], or [ - ][ _ ][ . ]. Space not allowed."
+            extra="Unique name given to a container." >
             <Input />
           </Item>
-          <Item label="Kind" {...props("kind")} rules={requiredRules}>
+          <Item label="Kind" {...props("kind")} rules={requiredRules}
+            extra="Format of the container." >
```
nafiz1001, line 141: Curious to know why you say "Format" here when it could've been "Kind"

nafiz1001 (approved):

frontend/src/components/containers/ContainerEditContent.js
UlysseFG: Defining or explaining a word or an expression using the same expression is not accomplishing much.

PR #618, opened 2023-07-25
Add InstrumentType API

nafiz1001:


backend/fms_core/viewsets/instrument_type.py
```diff
@@ -0,0 +1,27 @@
...
+    ordering_fields = (
+        *_list_keys(_instrument_type_filterset_fields),
+    )
+
+    filterset_fields = {
+        **_instrument_type_filterset_fields,
+    }
+
+    @action(detail=False, methods=["get"])
+    def list_names(self, _request):
```
UlysseFG, line 25: I don't think we need that ... or at least not specifically here ... It could be nice to provide a list_name to most endpoints but it would have to be present on each endpoint that has a name not just once here or there. Could you remove it from this PR and make a new ticket to add this feature to most endpoint. One thing that make a good API is that it is consistent.

UlysseFG:
The instrument type drop down in index validation is not working so i assume there is something going wrong ...

nafiz1001:
@UlysseFG I fixed your problem 🤗

UlysseFG (approved):

PR #619, opened 2023-07-25
Adding information column to locate tube in rack for normalization planning.

UlysseFG:


backend/fms_core/templates.py
```diff
@@ -17,6 +17,7 @@
     "LIBRARY_PREPARATION_TEMPLATE",
     "LIBRARY_QC_TEMPLATE",
     "NORMALIZATION_TEMPLATE",
+    "NORMALIZATION_PLANNING_TEMPLATE",
```
nafiz1001, line 20: Looks like I'll have to update the template prefiller branch after this gets merged ;(

PR #623, opened 2023-07-31
Fix to user not loaded, Import export removal and typo correction

UlysseFG:


nafiz1001:
many different changes in one PR 🥲

UlysseFG:
Yeah this is not the finest PR ever ... The only thing that is possibly not harmless is the removal of import-export but i don't think there is anything using this package anymore. I left the acknowledgment in the readme since we took inspiration from that package for our importer.

nafiz1001 (approved):
no more error 🙌🏼

PR #627, opened 2023-08-02
Fix missing dashboard part 2

nafiz1001:


UlysseFG (approved):

UlysseFG:
Haven't managed to get it to break so i assume it's at least as good as previous version ;)

PR #628, opened 2023-08-03
Prefill more columns for Extraction and Transfer

nafiz1001:


backend/fms_core/templates.py
```diff
@@ -489,7 +489,25 @@
       },
   ],
   "user prefill info": {
+      # "Extraction Type": ["DNA", "RNA"], # already prefilled by workflow
+      "Volume Used (uL)": "number",
+      # borrowed from extraction template
+      "Destination Container Kind": [
```
UlysseFG, line 495: you should cast SAMPLE_CONTAINER_KINDS from containers.py into a list instead. The less places we have to update the same information the best.

backend/fms_core/templates.py
```diff
@@ -515,7 +533,23 @@
       },
   ],
   "user prefill info": {
+      # borrowed from transfer template
+      "Destination Container Kind": [
+        "Tube",
```
UlysseFG, line 538: you should cast SAMPLE_CONTAINER_KINDS from containers.py into a list instead. The less places we have to update the same information the best.

backend/fms_core/templates.py
nafiz1001: i'm not quite sure how it applies in here. i picked only the containers that are relevant to sample extraction. is this just for improving existing codes?

UlysseFG: It applies that w can typically extract in any container that can receive samples. These are listed in the above mentioned constant. This is meant to make the code require less maintenance. This list will be automatically updated with new values as new containers are created while your hard coded list will need to be updated specifically each time.

backend/fms_core/templates.py
```diff
@@ -489,7 +490,14 @@
       },
   ],
   "user prefill info": {
+      # "Extraction Type": ["DNA", "RNA"], # already prefilled by workflow
+      "Volume Used (uL)": "number",
+      # borrowed from extraction template
+      "Destination Container Kind": list(SAMPLE_NON_RUN_CONTAINER_KINDS),
+      "Volume (uL)": "number",
+      "Source Depleted": ["YES", "NO"],
```
UlysseFG, line 498: Remove the "NO" from the list there is really just an option here...

backend/fms_core/templates.py
```diff
@@ -515,7 +523,12 @@
       },
   ],
   "user prefill info": {
+      # borrowed from transfer template
+      "Destination Container Kind": list(SAMPLE_NON_RUN_CONTAINER_KINDS),
+      "Source Depleted": ["YES", "NO"],
```
UlysseFG, line 528: Remove the "NO" from the list there is really just an option here...

UlysseFG (approved):

PR #629, opened 2023-08-03
Prevent crash when instrument does not exist.

UlysseFG:


backend/fms_core/template_importer/row_handlers/sample_pooling/pool.py
nafiz1001, line 1: Did you just add an `if` and indent a whole paragraph?

backend/fms_core/template_importer/row_handlers/sample_submission/pool.py
nafiz1001, line 1: I'm guessing the same could be said about this?

nafiz1001 (approved):
Those two codes look suspiciously similar; there might be a way to reuse those two codes into a function.
Aside from that, there's nothing wrong that stood out ;)

backend/fms_core/template_importer/row_handlers/sample_pooling/pool.py
UlysseFG: yeah i did ... The errors was that when failing to find an instrument we would continue as if we actually did find one ... This fixes that.

UlysseFG:
> Those two codes look suspiciously similar; there might be a way to reuse those two codes into a function. Aside from that, there's nothing wrong that stood out ;)

yeah they are pretty similar ... We'll see about this once the code stabilize some and we are confident we can be sure these things will stay ... I do not really want to work too much on refactoring until the lab has used the system a bit and agrees those functionality work as intended.

PR #632, opened 2023-08-08
Feature #1802:  List only instrument_types we have instrument listed in the instrument table in the indice validation form.

nafiz1001:
https://206.12.92.46/issues/1802


backend/fms_core/viewsets/_constants.py
```diff
@@ -142,7 +142,8 @@
 _instrument_type_filterset_fields: FiltersetFields = {
     "id": PK_FILTERS,
     "type": CATEGORICAL_FILTERS_LOOSE,
-    **_prefix_keys("platform__", _platform_filterset_fields)
+    "instruments": ["isnull"],
```
UlysseFG, line 145: Maybe "NULLABLE_FK_FILTERS" would match ? Otherwise i would recommend creating a NULLABLE_FILTER category

frontend/src/components/indices/IndicesValidate.js
```diff
@@ -18,27 +18,31 @@ import PageContent from "../PageContent";
...
 
-const actionCreators = { list, validate };
+  const listCallback = useCallback((options) => {
+    return dispatch(list(options))
+  }, [dispatch])
+  const validateCallback = useCallback((options) => {
+    return dispatch(validate(options))
+  }, [dispatch])
+  const listInstrumentTypes = useCallback(() => {
+    return dispatch(api.instrumentTypes.list({ instruments__isnull: false }))
```
UlysseFG, line 42: This returns one instrument_type for each instrument... This means you have duplicates in the dropdown in the frontend... You could try adding .distinct() to the queryset in the viewset. If it does not work you may have to resolve the duplications in the frontend.

frontend/src/components/indices/IndicesValidate.js
```diff
@@ -18,27 +18,31 @@ import PageContent from "../PageContent";
...
-  token: state.auth.tokens.access,
-  indicesTotalCount: state.indices.totalCount,
-  isFetching: state.indices.isFetching,
-});
+  const dispatch = useAppDispatch()
 
-const actionCreators = { list, validate };
+  const listCallback = useCallback((options) => {
+    return dispatch(list(options))
+  }, [dispatch])
```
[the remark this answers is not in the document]
nafiz1001: no, `options` comes from the argument of the function, not from variables declared inside the component.

UlysseFG (approved):

frontend/src/components/indices/IndicesValidate.js
```diff
@@ -116,7 +125,7 @@ const IndicesValidate = ({ token, indicesTotalCount, isFetching, list, validate
   const onSubmit = () => {
     setValidationLoading(true)
     const data = serialize(formData)
-    validate(data)
+    validateCallback(data)
```
ckostiw, line 128: I'm not sure why `validateCallback` is defined as a callback, since it is being used like a simple function here. Ideally, `onSubmit` should be a callback.

frontend/src/components/indices/IndicesValidate.js
```diff
@@ -80,7 +89,7 @@ const IndicesValidate = ({ token, indicesTotalCount, isFetching, list, validate
     setIndexCount(prevIndexCount => prevIndexCount + numIndicesInSet)
 
     // load options lazily
-    list({ ...query }).then(response => {
+    listCallback({ ...query }).then(response => {
```
ckostiw, line 92: I'm not sure why `listCallback` is a callback. It's just being used as a function here, but `loadData` should probably be a callback.

frontend/src/components/indices/IndicesValidate.js
nafiz1001: I just called it that because I used useCallback.

ckostiw (approved):

PR #633, opened 2023-08-09
Feature #1824:  Display sample count column in specific protocols

nafiz1001:
https://206.12.92.46/issues/1824


UlysseFG (commented):
This works and seem ok to me i will leave Chris see if it matches his expectations on the side of the code.

ckostiw (approved):

PR #634, opened 2023-08-09
Fix login page use effect

ckostiw:
Should fix the issues we have had with random logouts and UX confusion about the current login state.

* Added a `isUserLoggedIn` function to have a consistent way to check the auth redux state to see if the user is logged in. Fixes the problem with `PrivateNavigate` and `LoginPage` having different opinions about whether or not the user is logged in. :)
* Added a hook to refresh the user's access token on app startup after the persisted auth state has been loaded from local storage.
* Added a hook to run a loop in the background that checks the auth token and fetches summaries.
* Removed an obsolete `didInvalidate` property from the auth state that was not used for anything.
* Fixed a race condition problem in `refreshAuthToken` where the tokens would be stored in the auth state after the user had been logged out.
* Deleted the `selectToken` selector which was duplicated by `selectAuthTokenAccess`
* The `fetchSummaries`, `fetchStaticData`, etc. actions are no longer responsible for ensuring the auth token in refreshed. The refresh loop hook is responsible for that now.


frontend/src/components/app/useAuthInit.tsx
```diff
@@ -0,0 +1,45 @@
...
+	const authState = useAppSelector(selectAuthState)
+	const [tokenVerified, setTokenVerified] = useState(false)
+	const isLoggedIn = tokenVerified && hasLoginInfo(authState)
+
+	// At startup we have to verify the tokens that were stored in local storage
+	// for auth, so that we can refresh the user's access token if possible.
+	// If the user's refresh token has expired (or if there was no token to begin with)
+	// then the auth state will be cleared and the user will get the login page.
+	// We only do this once, when the app is loaded.
+	useEffect(() => {
```
UlysseFG, line 31: It's unclear to me why you need this to be a useEffect when you only want to run it once when it is called... I may be missing some dynamic exception where once is not once ...

frontend/src/modules/auth/actions.ts
```diff
@@ -0,0 +1,116 @@
...
+ * The function will dispatch the LOG_OUT action if it is unable to refresh the tokens.
+ * 
+ * It returns true if the access token is currently valid and can be used to make
+ * api requests, and false otherwise.
+ * 
+ * @returns boolean
+ */
+export const refreshAuthToken = () => async (dispatch, getState) => {
+    // Refresh the access token if it is within 2 minutes of expiring
+    const TIME_BEFORE_TOKEN_EXPIRES = 2 * 60
```
UlysseFG, line 45: This should be somewhere centralized so we can find it. Also in order to have this not be synchronized to the summary calls we might want to make both timers be dependants and add an offset of a few seconds to make so they won't trigger at the same time.

UlysseFG: Also this may need to be bigger with templates taking 2-3 minutes to submit it's likely the app will miss the window if a template submission is timed badly.

frontend/src/modules/auth/actions.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    try {
+        const access = jwtDecode(tokens.access);
+        const refresh = jwtDecode(tokens.refresh);
+
+        const now = Date.now() / 1000;
+
+        if (access.exp > now + TIME_BEFORE_TOKEN_EXPIRES) {
+            // Access token is still valid for another while, don't refresh yet.
+            return true;
```
UlysseFG, line 68: I would tend to return false here instead ...

ckostiw: I'm using the return value of the function to know if the token can be used after the refresh check, so I return true if the access token is still valid, or after successfully refreshing the token. The function used to return either the result of the refresh token api call, or false, but that wasn't used anywhere so I "repurposed" the return value. :)

frontend/src/modules/auth/actions.ts
ckostiw: Submitting a template doesn't block the UI thread, so the timer isn't affected by that 2-3 minute wait, and we can refresh the token while we are waiting for the submission response. It wouldn't cause a problem in the frontend, but I'm not sure what would happen in the backend. If the backend received the submission request and was spending several minutes processing it, and in the meantime refreshed the access token would it "cancel" the submit operation because the token associated with the submit request had expired? I have no idea, but I'll change this to 5 minutes just in case.

ckostiw: There's no good place to put that `TIME_BEFORE_TOKEN_EXPIRES` constant right now. I would worry about that later when have time to implement a better system for managing auth.

frontend/src/components/app/useAuthInit.tsx
ckostiw: I am using a state variable to keep track of whether or not I've refreshed the token at startup, and you can only set set variables from within a useEffect.

Ideally, auth would be managed completely independently, rather than trying to do it from within a react component, but that's a much bigger and riskier refactor.

PR #635, opened 2023-08-11
Upgrade Django to 4.2.4

nafiz1001:


nafiz1001:
Currently no other route works except the root.

backend/fms/settings.py
```diff
@@ -167,6 +167,8 @@
              else [os.environ.get("FMS_HOST", ""), "localhost"]) if not DEBUG 
              else [])
 
+CSRF_TRUSTED_ORIGINS = ALLOWED_HOSTS
```
nafiz1001, line 170: I'm not sure if this makes any difference.

backend/fms/urls.py
```diff
@@ -26,14 +26,15 @@
...
     path('api/info/', SoftwareInformationView.as_view()),
+    path('api/csrf/', csrf),
     path('api/api-auth/', include('rest_framework.urls', namespace='rest_framework')),
     path('api/swagger/', schema_view.with_ui('swagger', cache_timeout=0), name='schema-redoc'),
     path('api/token/', jwt_views.TokenObtainPairView.as_view(), name='token_obtain_pair'),
     path('api/token/refresh/', jwt_views.TokenRefreshView.as_view(), name='token_refresh'),
     path('api/password_reset/', include('django_rest_passwordreset.urls', namespace='password_reset')),
     *static(settings.STATIC_URL, document_root=settings.STATIC_ROOT),
     *static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT),
+    path('', admin.site.urls),
```
nafiz1001, line 39: This is absolutely necessary.

backend/fms/views.py
```diff
@@ -1,5 +1,7 @@
+from django.http import JsonResponse
```
nafiz1001, line 1: not sure if JsonResponse is necessary, but since it is a JsonResponse, might as well keep it :)

frontend/src/utils/csrf.ts
```diff
@@ -0,0 +1,16 @@
+import { API_BASE_PATH } from "../config";
+
+// https://fractalideas.com/blog/making-react-and-django-play-well-together-single-page-app-model/
```
nafiz1001, line 3: It is our moral duty to reference where we obtained our code ;)

UlysseFG (approved):

ckostiw (commented):
Might be good to fix this warning:
```
ms_core.Biosample: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the FmsCoreConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```

This article might help:
https://koenwoortman.com/python-django-auto-created-primary-key-used-when-not-defining-primary-key-type/

UlysseFG (approved):

PR #640, opened 2023-08-17
Feature #1677:  Taxon definitions Nice to have.

nafiz1001:
https://206.12.92.46/issues/1677


ckostiw (changes requested):
* Did we decide not to display taxon names in italics?
* The table needs to be updated after the user creates a new taxon.
* The `taxons` state should be refreshed as well

frontend/src/components/taxons/TaxonsListContent.tsx
```diff
@@ -1,52 +1,49 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useMemo, useState } from 'react'
```
ckostiw, line 1: There are a lot of unused imports that should be cleaned up.

ckostiw (changes requested):

nafiz1001:
Oops.

nafiz1001:
I'm surprised `taxons` state is not refreshed
```typescript
        if (isAdding) {
            dispatch(
                add({ ...new_taxon })
            )
                .then(() => {
                    navigate('/taxons')
                })
                .then(() => { dispatch(list()) })
		.then(() => { dispatch(TaxonsTableActions.setStale(true)) })
                .catch(err => setFormErrors({ ...err.data }))
        } else {
            dispatch(
                update(taxon.id, { ...new_taxon })
            )
                .then(() => {
                    navigate('/taxons')
                })
                .then(() => { dispatch(list()) })
		.then(() => { dispatch(TaxonsTableActions.setStale(true)) })		
                .catch(err => setFormErrors({ ...err.data }))
        }
```

ckostiw (approved):

PR #644, opened 2023-09-07
Readset rework (Features #1830, 1873, 1874)

JMarce01:


backend/fms_core/models/dataset_file.py
```diff
@@ -1,23 +1,37 @@
...
 
 @reversion.register()
 class DatasetFile(TrackedModel):
     """ Class to store information about the files associated with their dataset for data deliveries. """
 
     readset = models.ForeignKey(Readset, on_delete=models.PROTECT, help_text="Readset of the file.", related_name="files")
-    file_path = models.CharField(max_length=STANDARD_FILE_PATH_LENGTH, help_text="Path to the dataset file.")
-    release_status = models.IntegerField(choices=ReleaseStatus.choices, default=ReleaseStatus.AVAILABLE, help_text="The release status of the file.")
-    release_status_timestamp = models.DateTimeField(null=True, blank=True, help_text='The last time the release status of the file was changed.')
+    file_path = models.CharField(null=False, max_length=STANDARD_FILE_PATH_LENGTH, help_text="Path to the dataset file.")
```
UlysseFG, line 16: null is false by defaut

backend/fms_core/models/dataset_file.py
```diff
@@ -1,23 +1,37 @@
...
+    size = models.IntegerField(null=False, help_text="Size of the dataset file.")
     validation_status = models.IntegerField(choices=ValidationStatus.choices, default=ValidationStatus.AVAILABLE, help_text="The run validation status of the file.")
```
UlysseFG, line 18: Does not really make sense to move the release status and let the validation status on the dataset_file. The validation status and timestamp should be moved either to readset or to the dataset (we are validating at that level anyway ...).

backend/fms_core/services/dataset.py
```diff
@@ -276,8 +277,12 @@ def ingest_run_validation_report(report_json):
         readset_by_name[readset_name] = readset_obj
         for key in readset:
             if key in ACCEPTED_DATASET_FILE_TYPES and readset[key]:
+                file: DatasetFileReport = readset[key]
+                if file.get('final_path') is None:
+                    warnings.append(("final_path not provided for file type {0} of readset {1}", key, readset_name))
```
UlysseFG, line 282: The path is required (a file without path is nothing) this should raise an error.

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+                return readset && readset.metrics && readset.index ? <div> {readset.index} </div> : ''
+            }
+        },
+        [ReadsetColumnID.NB_READS]: {
+            columnID: ReadsetColumnID.NB_READS,
+            title: "Number of Reads",
+            dataIndex: ['readset', 'nb_reads'],
+            sorter: true,
+            render: (_, { readset }) => {
+                return readset && readset.metrics && readset.metrics['nb_reads'] ? <div> {Number(readset.metrics['nb_reads'].value_numeric).toFixed(3)} </div> : ''
```
ckostiw, line 91: The number of reads is a whole number, so it should not have 3 digits after the decimal.

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+            }
+        },
+        [ReadsetColumnID.NB_BASES]: {
+            columnID: ReadsetColumnID.NB_BASES,
+            title: "Number of bases",
+            dataIndex: ['readset', 'nb_bases'],
+            sorter: true,
+            render: (_, { readset }) => {
+                return readset && readset.metrics && readset.metrics['nb_bases'] ? <div> {Number(readset.metrics['nb_bases'].value_numeric).toFixed(3)} </div> : ''
```
ckostiw, line 100: The number of bases is a whole number, so it should not have 3 digits after the decimal.

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export const READSET_COLUMN_FILTERS: { [key in ReadsetColumnID]: FilterDescription } = {
+    [ReadsetColumnID.ID]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
+        key: UNDEFINED_FILTER_KEY,
+        label: "id",
+    },
+    [ReadsetColumnID.SAMPLE_NAME]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 143: should be FILTER_TYPE.INPUT

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "sample_name",
+    },
+    [ReadsetColumnID.RELEASE_STATUS]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 148: should be FILTER_TYPE.INPUT

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "release_status",
+    },
+    [ReadsetColumnID.LIBRARY_TYPE]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 153: should be FILTER_TYPE.INPUT

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "library_type",
+    },
+    [ReadsetColumnID.INDEX]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 158: should be FILTER_TYPE.INPUT

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "index",
+    },
+    [ReadsetColumnID.NB_READS]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 163: should be FILTER_TYPE.RANGE

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "nb_reads",
+    },
+    [ReadsetColumnID.NB_BASES]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 168: should be FILTER_TYPE.RANGE

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        key: UNDEFINED_FILTER_KEY,
+        label: "nb_bases",
+    },
+    [ReadsetColumnID.MEAN_QUALITY_SCORE]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
+        key: UNDEFINED_FILTER_KEY,
+        label: "mean_qualirt_score",
+    },
+    [ReadsetColumnID.BLAST_HIT]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
```
ckostiw, line 178: should be FILTER_TYPE.INPUT

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+    },
+    [ReadsetColumnID.NB_BASES]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
+        key: UNDEFINED_FILTER_KEY,
+        label: "nb_bases",
+    },
+    [ReadsetColumnID.MEAN_QUALITY_SCORE]: {
+        type: FILTER_TYPE.INPUT_OBJECT_ID,
+        key: UNDEFINED_FILTER_KEY,
+        label: "mean_qualirt_score",
```
ckostiw, line 175: typo 'qualirt'

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+        label: "blast_hit",
+    }
+}
+
+export const READSET_FILTER_KEYS: { [key in ReadsetColumnID]: string } = {
+    [ReadsetColumnID.ID]: 'id',
+    [ReadsetColumnID.SAMPLE_NAME]: 'sample_name',
+    [ReadsetColumnID.RELEASE_STATUS]: 'release_status',
+    [ReadsetColumnID.LIBRARY_TYPE]: 'library_type',
+    [ReadsetColumnID.INDEX]: 'index',
```
ckostiw, line 189: This filter key has to be the path to the index of the derived sample of the readset, ie. 'derived_sample__library__index__name'

frontend/src/components/readsets/ReadsetsTableColumns.tsx
```diff
@@ -0,0 +1,194 @@
...
+    [ReadsetColumnID.NB_READS]: 'nb_reads',
```
ckostiw, line 190: Hmm. To filter by metric value we would need a key here that matches a metric by name, eg 'nb_reads'. I'm not sure how to formulate a key like that. Please check with Ulysse.

UlysseFG: I'm not sure myself if you can have a single key to do that ... it would be the combination of metrics__name for the type of metrics and metrics__value_numeric. Since we do not really want to do something flexible, you can add an annotation on the readset viewset to expose the nb_reads specifically it should allow you to filter using that key (something like readset_reads)...

backend/fms_core/models/dataset_file.py
```diff
@@ -1,23 +1,35 @@
...
+        ]
+    
+    def clean(self):
+        super().clean()
+        errors = {}
+
+        def add_error(field: str, error: str):
+            _add_error(errors, field, ValidationError(error))
+
+        if self.size <= 0:
```
UlysseFG, line 31: A file could technically have size 0 ... I would not block 0 size files.

backend/fms_core/serializers.py
```diff
@@ -583,36 +583,57 @@ class DatasetSerializer(serializers.ModelSerializer):
     released_status_count = serializers.SerializerMethodField()
     blocked_status_count = serializers.SerializerMethodField()
     latest_release_update = serializers.SerializerMethodField()
-
+    total_readsets = serializers.SerializerMethodField()
     class Meta:
         model = Dataset
-        fields = ("id", "external_project_id", "run_name", "lane", "files", "released_status_count", "blocked_status_count", "latest_release_update", "project_name", "metric_report_url")
+        fields = ("id", "external_project_id", "run_name", "lane", "files", "released_status_count", "blocked_status_count", "latest_release_update", "project_name", "metric_report_url", "total_readsets")
```
UlysseFG, line 589: total_readsets name breaks the typical naming pattern ... we typically use "_count".

backend/fms_core/services/dataset.py
```diff
@@ -122,25 +125,17 @@ def create_dataset_file(readset: Readset,
...
-    dataset_file_data = {}
-
     try:
         dataset_file = DatasetFile.objects.create(readset=readset,
                                                   file_path=file_path,
-                                                  validation_status=validation_status,
-                                                  **(dict(validation_status_timestamp=timezone.now()) if validation_status != ValidationStatus.AVAILABLE else dict()), # Set timestamp if setting Status to non-default
-                                                  release_status=release_status,
-                                                  **(dict(release_status_timestamp=timezone.now()) if release_status != ReleaseStatus.AVAILABLE else dict())) # Set timestamp if setting Status to non-default
+                                                  size=size) # Set timestamp if setting Status to non-default
```
UlysseFG, line 138: You can remove the comment it no longer applies.

backend/fms_core/viewsets/readset.py
```diff
@@ -1,25 +1,67 @@
...
+        try:
+            release_status = data.get("release_status")
+            if release_status is not None:
+                data["release_status_timestamp"] = timezone.now()
+            
+            readset_to_update = Readset.objects.filter(pk=data['id']).exclude(validation_status=ValidationStatus.AVAILABLE)
+            if not readset_to_update.exists():
+                return HttpResponseBadRequest(f"Run must first be validated before release status can be changed.")
+            
+            readset_to_update = Readset.objects.select_for_update().get(pk=data['id'], validation_status=ValidationStatus.PASSED)
```
UlysseFG, line 56: This will fail if the readset status is failed ... DoesNotExist raised.

backend/fms_core/viewsets/readset.py
```diff
@@ -1,25 +1,67 @@
...
+            readset_to_update.__dict__.update(data)
```
UlysseFG, line 57: Users should be allowed to release readsets that have Failed validation status.

UlysseFG, line 57: This introduce possible side effects to an endpoint that should only change the release status and release status timestamp ... Data could contain pretty much anything ... it could change the sample_name if data contains that field. I suggest you either make your own dictionary with only the fields that should be updated, or you surcharge the update endpoint instead of creating a specialized endpoint that contain restrictions to what can be changed in which condition.

backend/fms_core/viewsets/readset.py
```diff
@@ -1,25 +1,67 @@
...
+        except Exception as err:
+            raise ValidationError(dict(non_field_errors=err))
+        try:
+
+            serializer = ReadsetSerializer(readset_to_update)
```
UlysseFG, line 62: Careful here is an hardcoded serializer... if you want to make sure you do not override a client side readset with the wrong version of the readset you should make a call to get_serializer_class (the client will be responsible to ask for the right one in their request).

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -0,0 +1,256 @@
...
+    }, releaseStatusOption, canReleaseOrBlockFiles)
+
+    let readsetTableColumns = [
+        columnDefinitions.ID,
+        columnDefinitions.SAMPLE_NAME,
+        columnDefinitions.RELEASE_STATUS,
+        columnDefinitions.LIBRARY_TYPE,
+        columnDefinitions.INDEX,
+        columnDefinitions.NUMBER_READS,
+        // columnDefinitions.NUM_BASES,
```
UlysseFG, line 89: Remove unused code do not comment it out.

frontend/src/utils/api.js
```diff
@@ -45,9 +45,12 @@ const api = {
   datasets: {
     get: id => get(`/datasets/${id}/`),
     list: (options, abort) => get("/datasets/", options, { abort }),
-    setReleaseStatus: (id, release_status, exceptions = [], filters = {}) => patch(`/datasets/${id}/set_release_status/`, { release_status, exceptions, filters })
   },
-
+  readsets: {
+    get: id => get(`/readsets/${id}`),
+    list: (options, abort) => get(`/readsets/`, options, { abort }),
+    setReleaseStatus: readset => post(`/readsets/set_release_status/`, readset),
```
UlysseFG, line 52: You should add options to your actions to be able to get either the regular readset or the readsetWithMetrics...

frontend/src/modules/readsets/reducers.ts
```diff
@@ -0,0 +1,51 @@
...
+        items: [],
+        isFetching: false,
+    }
+    ,
+    action) => {
+    switch (action.type) {
+        case READSETS.GET.REQUEST:
+            return merge(state, ['itemsByID', action.meta.id], { id: action.meta.id, isFetching: true });
+        case READSETS.GET.RECEIVE:
+            return merge(state, ['itemsByID', action.meta.id], { ...action.data, isFetching: false });
```
UlysseFG, line 19: you will likely override withMetrics readsets without readset if you do not have the same mechanism to deal metrics in your get as in your list.

backend/fms_core/models/dataset_file.py
```diff
@@ -1,23 +1,35 @@
...
 class DatasetFile(TrackedModel):
     """ Class to store information about the files associated with their dataset for data deliveries. """
 
     readset = models.ForeignKey(Readset, on_delete=models.PROTECT, help_text="Readset of the file.", related_name="files")
     file_path = models.CharField(max_length=STANDARD_FILE_PATH_LENGTH, help_text="Path to the dataset file.")
-    release_status = models.IntegerField(choices=ReleaseStatus.choices, default=ReleaseStatus.AVAILABLE, help_text="The release status of the file.")
-    release_status_timestamp = models.DateTimeField(null=True, blank=True, help_text='The last time the release status of the file was changed.')
-    validation_status = models.IntegerField(choices=ValidationStatus.choices, default=ValidationStatus.AVAILABLE, help_text="The run validation status of the file.")
-    validation_status_timestamp = models.DateTimeField(null=True, blank=True, help_text='The last time the run validation status of the file was changed.')
+    size = models.IntegerField(null=False, help_text="Size of the dataset file.")
```
UlysseFG, line 17: integer field is likely too small to store the size of the files. should use bigintegerfield.

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -0,0 +1,256 @@
...
+                                                        display: 'flex',
+                                                        flexDirection: 'column',
+                                                    }} key={name}>
+                                                        {<b >
+                                                            {name.replace(/_/g, " ")}
+                                                        </b>
+                                                        }
+                                                        {readset.metrics[name].value_numeric
+                                                            ?
+                                                            Number(readset.metrics[name].value_numeric).toFixed(3)
```
UlysseFG, line 233: pf_clusters, yield, nb_reads, nb_bases are integer numbers. Maybe make a small helper to strip precision numbers if they are 0.

UlysseFG:
The Experiment dataset tab has more existing readsets in readset released denominator than there is actual readsets. It is likely still counting files or there is an issue with the count ...

UlysseFG:
In the release data listing there is a "show advanced filters" that should not be there unless it contains a filter.

UlysseFG:
The "release all" does only a release all of the current page... this could be source of errors and this is a bit counterintuitive... From what i recall the original code did release all and block all ... It used to pass a flag with exceptions ... this should still work as far i know. Let me know if this is a big issue we can possibly delay that to next sprint.

UlysseFG (approved):
Looks good, and working as far as i can see.

PR #650, opened 2023-09-14
Refactor users table

ckostiw:


frontend/src/components/pagedItemsTable/useCanWrite.tsx
```diff
@@ -0,0 +1,21 @@
...
+export function useCanWrite() {
+	const auth = useAppSelector(selectAuthState)
+	const usersByID = useAppSelector(selectUsersByID)
+	const { currentUserID } = auth
+
+	const canWrite = useMemo(() => {
+		if (currentUserID) {
+			const currentUser = usersByID[currentUserID]
+			if(currentUser) {
+				return currentUser.is_staff || currentUser.is_superuser
```
UlysseFG, line 15: I would tend to think we should restrain the user creation to staff ... The idea is to prevent a superuser from creating a staff user and actually upgrading his own permissions ... This is not ideal but until we have a more comprehensive way to deal with permissions we should be careful.

UlysseFG (approved):

PR #651, opened 2023-09-15
Refactor indices table

ckostiw:


frontend/src/modules/indices/reducers.js
```diff
@@ -77,41 +36,15 @@ export const indices = (
...
-
         case INDICES.VALIDATE.REQUEST:
             return { ...state, error: undefined, isFetching: true };
         case INDICES.VALIDATE.RECEIVE:
             return {...state, isFetching: false };
         case INDICES.VALIDATE.ERROR:
             return { ...state, error: action.error, isFetching: false };
+
+        default:
+            return state
```
UlysseFG, line 47: Calling a unknown reducers do not generate an error ?

ckostiw: Reducers receive every action that gets dispatched. If the action isn't one that the reducer is interested in then it must return the unmodified state. I'm not sure how that original reducer worked without causing some kind of error or crash.

UlysseFG: The default is likely set in combineReducers ... looks like the expected value in any case.

PR #657, opened 2023-09-24
Feature #1923:  Add an optional description to Study

nafiz1001:


backend/fms_core/services/study.py
```diff
@@ -24,7 +24,7 @@ def skip_char(letter: str, distance: int):
     
     return new_letter
 
-def create_study(project, workflow, start, end):
+def create_study(project, workflow, start, end, description=''):
```
UlysseFG, line 27: Use None as default instead.

backend/fms_core/models/study.py
```diff
@@ -20,6 +20,7 @@ class Study(TrackedModel):
     workflow = models.ForeignKey(Workflow, on_delete=models.PROTECT, related_name="studies", help_text="Workflow assigned to the study.")
     start = models.PositiveIntegerField(help_text="Index to the order of the start of the assigned workflow for this study.")
     end = models.PositiveIntegerField(help_text="Index to the order of the end of the assigned workflow for this study.")
+    description = models.TextField(blank=True, help_text="Description of the study.")
```
UlysseFG, line 23: Set null=True here and in the migration.

frontend/src/components/studies/StudyEditForm.tsx
```diff
@@ -114,6 +121,16 @@ const StudyEditForm = ({ workflows, isCreatingStudy, onSubmit, formErrors }: Cre
 					onChange={workflowWasSelected}
 				/>
 			</Item>
+			<Item
+				name="description"
+				label="Description"
+				{...itemValidation('description')}
+			>
+				<Input
+					defaultValue={''}
```
UlysseFG, line 130: null might be a better default since it would prevent to populate the DB with empty strings unless you do additional tests.

nafiz1001: Not possible, but I can make it None at the viewset or service.

backend/fms_core/viewsets/study.py
```diff
@@ -43,13 +43,17 @@ def create(self, request, *args, **kwargs):
                 workflow = None
         except Workflow.DoesNotExist:
             errors['workflow'].append(f"Workflow with id {study_data['workflow']} does not exist.")
+        
+        if 'description' not in study_data:
```
UlysseFG, line 47: This is an optional field ...

backend/fms_core/viewsets/study.py
```diff
@@ -43,13 +43,17 @@ def create(self, request, *args, **kwargs):
...
+            errors['description'].append(f"'description' key must be present")
 
         if not any(bool(error) for error in errors.values()): 
             # Call create study service
             study, errors_service, _ = create_study(project=project,
                                                     workflow=workflow,
                                                     start=study_data['start'],
-                                                    end=study_data['end'])
+                                                    end=study_data['end'],
+                                                    description=study_data['description'] or None)
```
UlysseFG, line 56: use if statement they are clearer and safer.

UlysseFG:
The Experiment dataset tab has more existing readsets in readset released denominator than there is actual readsets. It is likely still counting files or there is an issue with the count ...

UlysseFG:
Can you rename your migration file to a match the other migration file naming ? You can add a comment inside the file to summarize the migration content instead.

frontend/src/components/studies/StudyEditForm.tsx
```diff
@@ -114,6 +122,16 @@ const StudyEditForm = ({ workflows, isCreatingStudy, onSubmit, formErrors }: Cre
 					onChange={workflowWasSelected}
 				/>
 			</Item>
+			<Item
+				name="description"
+				label="Description"
+				{...itemValidation('description')}
+			>
+				<Input
```
UlysseFG, line 130: The content may overflow from the single line input field. I would maybe try the textarea with autosize prop instead. This leave the component small if not much text is input, but let the user see the whole content if more than one line of input is required. Might also be nice when entering the description that the "enter" not be captured by the submit button but instead insert a new line in the field (optional) ...

UlysseFG (approved):

PR #664, opened 2023-10-06
Feature #1927:  Sample absorbance QC Tecan-Spark integration

nafiz1001:


backend/fms_core/template_importer/importers/sample_qc_spark.py
```diff
@@ -0,0 +1,127 @@
...
+        # Add the template to the process
+        if self.imported_file is not None:
+            self.preloaded_data['process'].imported_template_id = self.imported_file.id
+            self.preloaded_data['process'].save()
+
+        for row_id, row_data in enumerate(spark_sample_qc_sheet.rows):
+            row_data: SampleQCSparkSheet = row_data
+            process_measurement_properties = self.preloaded_data['process_properties']
+
+            quantity_flag = str_cast_and_normalize(float(row_data['Mass/rxn (ug)']) > 1000)
```
UlysseFG, line 83: Don't convert to float if you don't have to the more conversion we do the more error we introduce. Most calculations are done using decimals.

backend/fms_core/template_importer/row_handlers/sample_qc_spark/sample.py
```diff
@@ -0,0 +1,113 @@
...
+            # Check if sample is not a library or a pool of libraries
+            if sample_obj.is_library:
+                self.errors['source_sample'] = f"Source sample can't be a library or a pool of libraries."
+
+            # Update sample with sample_information
+            if sample_information['concentration'] is None:
+                self.errors['concentration'] = 'Concentration value is required'
+
+            _, self.errors['sample_update'], self.warnings['sample_update'] = \
+                update_sample(sample_to_update=sample_obj,
```
UlysseFG, line 68: Do not update the sample concentration using the spark QC value. The sample should already have a concentration which should be correct. Only the quantity_flag needs to be updated with the one from this step,

backend/fms_core/template_importer/row_handlers/sample_qc_spark/sample.py
```diff
@@ -0,0 +1,113 @@
...
+                              concentration=sample_information['concentration'])
+
+            # Update the sample's flags with sample information
+            _, self.errors['flags'], self.warnings['flags'] = update_qc_flags(sample=sample_obj,
+                                                                              quantity_flag=sample_information['quantity_flag'],
+                                                                              quality_flag=sample_information['quality_flag'])
```
UlysseFG, line 74: We should change update_qc_flags to take only quality or quantity flags. We can return a warning if both are not set at once. In this case we should not update the quality_flag.

backend/fms_core/template_importer/row_handlers/sample_qc_spark/sample.py
```diff
@@ -0,0 +1,113 @@
...
+                        type = process_measurement_properties[instrument]['value']
+                        it = InstrumentType.objects.get(type=type)
+                        # Validate platform and type
+                        if it.platform.name != QC_PLATFORM:
+                            self.errors['instrument_type'] = f'Invalid type: {it.platform} for instrument: {it.type}.'
+                    except Exception as e:
+                        self.errors['instrument'] = f'Invalid instrument {type}.'
+
+                # Validate required RIN for RNA
+                if sample_obj.derived_samples.first().sample_kind.name == 'RNA' and process_measurement_properties['RIN']['value'] is None:
```
UlysseFG, line 110: there is no RIN to set for Axiom.

backend/fms_core/template_importer/importers/sample_qc_spark.py
```diff
@@ -0,0 +1,127 @@
...
+            concentration = float_to_decimal_and_none(float(row_data['Concentration ug/ul']) * 1000)
+
+            process_measurement_properties['Sample Quantity QC Flag']['value'] = quantity_flag
+            process_measurement_properties['Concentration (ng/uL)']['value'] = concentration
+
+            sample = {
+                'coordinates': str_cast_and_normalize(row_data['Well positions']),
```
UlysseFG, line 90: Well positions is in a slightly different format than ours. There is no 0 padding in here so you'll have to create a small helper function (somewhere that can be used elsewhere (template_importer/_utils.py maybe)) to convert the input.

backend/fms_core/template_importer/row_handlers/sample_qc_spark/sample.py
```diff
@@ -0,0 +1,113 @@
...
+                                                                                             step=workflow["step"],
+                                                                                             current_sample=sample_obj,
+                                                                                             process_measurement=process_measurement_obj)
+
+            if process_measurement_obj and properties_obj:
+                # Validate instruments according to platform
+                for instrument in INSTRUMENT_PROPERTIES:
+                    try:
+                        type = process_measurement_properties[instrument]['value']
+                        it = InstrumentType.objects.get(type=type)
```
UlysseFG, line 102: try using readable variable names as much as possible (instrument_type instead of it).

nafiz1001: i just copy pasted what was in sample qc row handler ;)

UlysseFG:
Not much reusability ... Each integration is linked to the instrument the result file comes from. If another workflow need results from the Spark we could expend the integration to save more Metadata, but i don't think there will realistically be much more.

PR #669, opened 2023-10-16
Automation step

UlysseFG:


backend/fms_core/models/step_history.py
```diff
@@ -19,7 +20,8 @@
 class StepHistory(TrackedModel):
     study = models.ForeignKey(Study, on_delete=models.PROTECT, related_name="StepHistory", help_text="Study associated to the process measurement.")
     step_order = models.ForeignKey(StepOrder, on_delete=models.PROTECT, related_name="StepHistory", help_text="Step order in the study that is associated to the process measurement.")
-    process_measurement = models.ForeignKey(ProcessMeasurement, on_delete=models.PROTECT, related_name="StepHistory", help_text="Process measurement associated to the study step.")
+    process_measurement = models.ForeignKey(ProcessMeasurement, null=True, blank=True, on_delete=models.PROTECT, related_name="StepHistory", help_text="Process measurement associated to the study step.")
+    sample = models.ForeignKey(Sample, on_delete=models.PROTECT, related_name="StepHistory", help_text="Source sample that completed the step.")
```
JMarce01, line 24: what is the purpose of this sample field?

frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -24,7 +24,7 @@ import { SAMPLE_COLUMN_FILTERS, SAMPLE_NEXT_STEP_FILTER_KEYS, SampleColumnID } f
 const { Text } = Typography
 
 interface LabworkStepPageProps {
-	protocol: Protocol
+	protocol: Protocol | undefined
```
JMarce01, line 27: just a suggestion, maybe create a type that is for 'Protocol | undefined', and change the name since it's no longer always associated to a protocol. Maybe we can just merge and talk about it with Chris. Maybe we can have checks that always check step.type first instead of having an optional for protocol, that might change how to front-end is going to be setup.

frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -123,6 +126,38 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
...
+  
+  const handleExecuteAutomation = useCallback(
+    async () => {
+      try {
+        const response = await dispatch(requestAutomationExecution(step.id))
+        if (response) {
+          const success = response.data.result.success
+          if (success) {
+            dispatch(flushSamplesAtStep(step.id))
+            const AUTOMATION_SUCCESS_NOTIFICATION_KEY = `LabworkStep.automation-success-${step.id}`
```
JMarce01, line 138: suggestion to store these somewhere as a constant

backend/fms_core/models/step_history.py
UlysseFG: When there is no process_measurement the association between a StepHistory and a specific sample cannot be made otherwise... It replaces process_measurement.source_sample to provide the sample mapping...

frontend/src/components/labwork/step/LabworkStep.tsx
UlysseFG: Yeah i think we can use python formating to make it a constant.

UlysseFG: Looking at how the value is variable it would not make much sense to move part of it outside you still would have to check here then also somewhere else. Maybe i am not getting your meaning.

PR #672, opened 2023-10-25
Addition of Illumina_IDT_Nextera_UDP set to Library preparation and Sample submission templates.

UlysseFG:


UlysseFG:
While checking this make sure i did not break the pool drop down formula ... (the file size diff is suspicious).

PR #674, opened 2023-11-02
change create_container to default container name to barcode

JMarce01:
- name is already defaulted to barcode in `services/container.py`
- overwrite create in container.py viewset to use create container service
- add warnings and errors for create container services
- add errors for create container viewset

### DIFF
--- a/backend/fms_core/services/container.py
+++ b/backend/fms_core/services/container.py
@@ -139,7 +139,12 @@ def get_or_create_container(barcode, kind=None, name=None, coordinates=None, con
                 errors.append(f"Container kind is required to create a container.")
             elif container_parent and CONTAINER_KIND_SPECS[container_parent.kind].requires_coordinates and not coordinates:
                 errors.append(f"Parent container kind {container_parent.kind} requires that you provide coordinates.")
+            elif name is None and Container.objects.filter(name=barcode):
+                    errors.append(f"Missing container name, barcode will replace container name. Container with name {barcode} already exists")
             else:
+
+                if not Container.objects.filter(barcode=barcode).exists() and name is None:
+                    warnings.append(f'Missing container name, container barcode will be used as container name')
                 try:
                     container = Container.objects.create(**container_data, comment=comment)
                     created_entity = True
@@ -163,7 +168,11 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
         if kind:
             if Container.objects.filter(barcode=barcode).exists():
                 errors.append(f"Container with barcode {barcode} already exists.")
+            elif name is None and Container.objects.filter(name=barcode):
+                errors.append(f"Missing container name, barcode will replace container name. Container with name {barcode} already exists")
             else:
+                if not Container.objects.filter(barcode=barcode).exists() and name is None:
+                    warnings.append(f'Missing container name, container barcode will be used as container name')
                 try:
                     coordinate = Coordinate.objects.get(name=coordinates) if coordinates is not None else None
                 except Coordinate.DoesNotExist as err:
--- a/backend/fms_core/services/container.py
+++ b/backend/fms_core/services/container.py
@@ -180,7 +180,7 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
                 container_data = dict(
                     **(dict(location=container_parent) if container_parent else dict()),
                     **(dict(barcode=barcode) if barcode is not None else dict()),
-                    **(dict(name=name) if (name is not None and name) else dict(name=barcode)), # By default, a container name will be his barcode
+                    **(dict(name=name) if name is not None else dict(name=barcode)), # By default, a container name will be his barcode
                     **(dict(coordinate=coordinate) if coordinate is not None else dict()),
                     **(dict(kind=kind) if kind is not None else dict()),
                 )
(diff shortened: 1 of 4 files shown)


backend/fms_core/services/container.py
```diff
@@ -139,7 +139,12 @@ def get_or_create_container(barcode, kind=None, name=None, coordinates=None, con
                 errors.append(f"Container kind is required to create a container.")
             elif container_parent and CONTAINER_KIND_SPECS[container_parent.kind].requires_coordinates and not coordinates:
                 errors.append(f"Parent container kind {container_parent.kind} requires that you provide coordinates.")
+            elif name is None and Container.objects.filter(name=barcode):
+                    errors.append(f"Missing container name, barcode will replace container name. Container with name {barcode} already exists")
```
UlysseFG, line 143: Start with the error part. Finish your error message with a point.

Since you are using the Service for the form creation you may want some bad entry hit the DB and generate a validation error for the form to catch. The validation error for fields with the same name as fields in the form are mapped to it and displayed. You can also patch it up in the viewset by creating your own validation errors. It might be nice in this case to actually have some more to go on that name already exists with an empty field above.

Its important to make sure that all case are caught and the operation is rolled back in case of bad entry.

backend/fms_core/services/container.py
```diff
@@ -139,7 +139,12 @@ def get_or_create_container(barcode, kind=None, name=None, coordinates=None, con
...
             else:
+
+                if not Container.objects.filter(barcode=barcode).exists() and name is None:
+                    warnings.append(f'Missing container name, container barcode will be used as container name')
```
UlysseFG, line 147: missing "."

backend/fms_core/services/container.py
```diff
@@ -163,7 +168,11 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
         if kind:
             if Container.objects.filter(barcode=barcode).exists():
                 errors.append(f"Container with barcode {barcode} already exists.")
+            elif name is None and Container.objects.filter(name=barcode):
+                errors.append(f"Missing container name, barcode will replace container name. Container with name {barcode} already exists")
```
UlysseFG, line 172: missing "."

backend/fms_core/services/container.py
```diff
@@ -163,7 +168,11 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
         if kind:
             if Container.objects.filter(barcode=barcode).exists():
                 errors.append(f"Container with barcode {barcode} already exists.")
+            elif name is None and Container.objects.filter(name=barcode):
```
UlysseFG, line 171: You need the ".exists()" or you will test if you have or not a queryset (which you most definitively have).

backend/fms_core/services/container.py
```diff
@@ -163,7 +168,11 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
...
+                errors.append(f"Missing container name, barcode will replace container name. Container with name {barcode} already exists")
             else:
+                if not Container.objects.filter(barcode=barcode).exists() and name is None:
+                    warnings.append(f'Missing container name, container barcode will be used as container name')
```
UlysseFG, line 175: missing "."

backend/fms_core/services/container.py
```diff
@@ -139,7 +139,12 @@ def get_or_create_container(barcode, kind=None, name=None, coordinates=None, con
                 errors.append(f"Container kind is required to create a container.")
             elif container_parent and CONTAINER_KIND_SPECS[container_parent.kind].requires_coordinates and not coordinates:
                 errors.append(f"Parent container kind {container_parent.kind} requires that you provide coordinates.")
+            elif name is None and Container.objects.filter(name=barcode).exists():
+                    errors.append(f"Container with name {barcode} already exists. Missing container name, barcode will replace container name.")
```
UlysseFG, line 143: Reword the end some so it makes sense in the context that the creation will not go forward ... "... already exists while using barcode as name (container name not provided)." something like that ...

backend/fms_core/services/container.py
```diff
@@ -163,7 +168,11 @@ def create_container(barcode, kind, name=None, coordinates=None, container_paren
         if kind:
             if Container.objects.filter(barcode=barcode).exists():
                 errors.append(f"Container with barcode {barcode} already exists.")
+            elif name is None and Container.objects.filter(name=barcode).exists():
+                errors.append(f"Container with name {barcode} already exists. Missing container name, barcode will replace container name. ")
```
UlysseFG, line 172: same as above... the second sentence does not make much sense in the context.

frontend/src/components/containers/ContainerEditContent.js
```diff
@@ -135,9 +135,9 @@ const ContainerEditContent = ({ token, containerKinds, containersByID, add, upda
           onValuesChange={onValuesChange}
           onFinish={onSubmit}
         >
-          <Item label="Name" {...props("name")} rules={nameRules.concat(requiredRules)}
+          <Item label="Name" {...props("name")}
```
UlysseFG, line 138: Swap Name and Barcode in the form so we don't have the optional field first.

UlysseFG:
Adding a container through the form provoke a 500 server error when it tries to load an undefined container in the frontend ... The returned serialized object might not be what the frontend expects.

UlysseFG (approved):

PR #681, opened 2023-11-09
Transfer LibQC coords

JMarce01:
- use source coords according to template dropdown, using formulas


UlysseFG:
Can you put the formula on the right column (Sample Destination Coord) ? The validation of the cells may get in the way if there is conflict between formula and drop down remove the validation.

UlysseFG:
You need to fix the content of the file too not just the name ... You still have the formula on the barcode column. You also need to remove the old version (v4_4_0) of the template.

UlysseFG:
The change you did to the test template will break the tests ...

UlysseFG:
"Destination Container Coord" is not an helpful cell name ... Try something more telling ... "Keep Source Coord for Destination" or something else you think would be clearer ...

UlysseFG (approved):

PR #689, opened 2023-11-16
Empty row messages, removing extra empty columns

JMarce01:
- add errors when there are missing lines, with data following the missing lines
- removing columns with missing headers when cells with random data inside
- adding test case


backend/fms_core/template_importer/sheet_data.py
```diff
@@ -40,25 +40,32 @@ def prepare_rows(self):
...
                 'validation_error': ValidationError([]),
                 'warnings': [],
             }
             self.rows_results.append(result)
-
+            
+            #checks if entire row has missing data (empty cells)
+            empty_row = row if self.check_for_empty_data_array(row_data) else ''
+            if empty_row:
+                self.empty_row = empty_row
```
UlysseFG, line 59: why not assign directly the above line to self.empty_row ?

backend/fms_core/template_importer/sheet_data.py
```diff
@@ -40,25 +40,32 @@ def prepare_rows(self):
         for row_id in data_row_ids_range(self.header_row_nb + 1, self.dataframe):
             row_data = self.dataframe.iloc[row_id]
             self.rows.append(row_data)
-
-            row_repr = f"#{row_id + 1}"
+            row = row_id + 1
+            row_repr = f"#{row}"
+            row_data = panda_values_to_str_list(row_data)
```
UlysseFG, line 45: Be careful with reallocation of variables ... the python language typically uses the reference when passing data around so you might end up breaking the self.rows by doing this here. I would instead create another variable at that point.

backend/fms_core/template_importer/sheet_data.py
```diff
@@ -67,4 +74,16 @@ def generate_preview_info_from_rows_results(self, rows_results):
...
\ No newline at end of file
+        }
+    
+    def check_for_empty_data_array(self,row_data):
+        for x in (row_data):
+            if x:
+                return False
+        return True
+    
+    def check_for_empty_row_errors(self):
```
UlysseFG, line 85: this does not deserve a function this won't be reused. It is just less readable you have to actually fetch the function to know it just raise an error... you can

backend/fms_core/template_importer/sheet_data.py
```diff
@@ -40,25 +40,33 @@ def prepare_rows(self):
...
             }
             self.rows_results.append(result)
+            
+            #checks if entire row has missing data (empty cells)
+            self.empty_row = row if not any(row_str_data) else self.empty_row
 
+        #checks if empty row exists in sheet_data, appends error with the last filled line in the sheet_data
+        if self.empty_row:
+            erroneous_row = self.empty_row + 1
+            self.base_errors.append(f'Empty line detected. Fill in empty line at row #{str(self.empty_row)} or remove data at row #{str(erroneous_row)}.')
```
UlysseFG, line 62: Given that you might have multiple empty lines you might tweak the message to match possible outcomes ... Maybe "... Fill in empty lines up to row ..."

UlysseFG:
While testing i realized that if you have stuff outside the valid area (the one bellow headers), you still get an error message.
I think the ideal would be to have an empty line error message if there is data in a valid area after an empty line otherwise ignore the empty lines ... (you might have to cut the dataframe manually for this). The reasoning is that it would likely break if people add comments outside the area or do calculations and this may be detrimental to users experience. If we strip the phantom lines (the case where the extra data is not bellow a header), this would not hurt anything (empty cells don't really need validation or warning to be removed).

PR #697, opened 2023-12-04
Break labwork samples by ...

UlysseFG:
Requires an upgrade of antd to fix an issue with collapsible. Raises some deprecations warnings...


frontend/src/modules/labworkSteps/actions.ts
```diff
@@ -260,14 +260,16 @@ export function setFilterOptions(stepID: FMSId, description: FilterDescription,
 	}
 }
 
-export function clearFilters(stepID: FMSId) {
+export function clearFilters(stepID: FMSId, refresh: boolean=true) {
```
JMarce01, line 263: why was it necessary to add a boolean, would you not always need to clear filters if it's being called?

JMarce01 (approved):

frontend/src/modules/labworkSteps/actions.ts
UlysseFG: The boolean addresses the issue that clearFilters by default fetches the sample listed even if this was just meant as a step in setting new filters ... this saves a useless refresh when set to false ... The alternative would be to create a new action that clearFilter and set a new Filter before fetching samples. I did not want  to create a new action unless we really needed one ...

PR #708, opened 2024-01-11
Visualization plate transfer

JMarce01:


frontend/src/components/placementVisuals/AddPlacementContainer.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const [error, setError] = useState<string | undefined>(undefined)
+    const [newContainer, setNewContainer] = useState<any>({})
+
+    const coordinates = useAppSelector(selectCoordinatesByID)
+
+    const dispatch = useAppDispatch()
+    //retrieves container on search
+    const handleContainerLoad = useCallback(async (loadedContainer) => {
+        const container = await dispatch(api.containers.get(loadedContainer))
+        const loadedSamples = await dispatch(api.samples.list({ id__in: container.data.samples.join(',') }))
```
ckostiw, line 31: If the container doesn't contain any samples then the id__in list will be empty and the api will return the first 100 samples from ALL of the samples in the DB.
To reproduce, just create a new plate in Freezeman directly then try to select it as a Placement destination. The api will return 100 of 360,000 samples. When I do this, I end up getting an error saying "Unsupported container type" (???).

ckostiw (changes requested):

frontend/src/components/placementVisuals/AddPlacementContainer.tsx
ckostiw: The code bombs if one of the first 100 samples is a tube with an undefined coordinate. When you try to lookup the coordinate name it causes an "undefined" exception, and the container load fails, so later `handleConfirm` bombs because the selected container is undefined.

Screen Shot 2024-01-24 at 3 57 36 PM

frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -242,6 +248,8 @@ const Placement = ({ sourceSamples, destinationSamples, cycleContainer, saveChan
         updateSampleList(samplesToUpdate, type)
     }, [selectedSamples, sourceSamples.samples, destinationSamples.samples])
 
+    const disableUndo = useMemo(() => Object.values(selectedSamples).some(sample => sample.type == 'source'), [selectedSamples])
```
UlysseFG, line 251: Whenever you have those harcoded values, you should try to import the corresponding constant (SOURCE_STRING here).

frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -67,7 +67,13 @@ const Placement = ({ sourceSamples, destinationSamples, cycleContainer, saveChan
 
     //removes selected samples, unless they're in the source container
     const removeSelectedCells = useCallback(() => {
-        const removed = (Object.keys(selectedSamples).length > 0 ? selectedSamples : destinationSamples.samples)
+        const selected = {}
+        Object.keys(selectedSamples).forEach(id => {
+            if (selectedSamples[id].type != "source") {
```
UlysseFG, line 72: You should try to import the corresponding constant (SOURCE_STRING here).

PR #728, opened 2024-02-29
Placement support more formats

UlysseFG:
Cut short to all that can be done here.. we'll have to come back to it...
The most important change remaining is to add clickable columns and rows for selection and an addition button to select all.

### DIFF
--- a/backend/fms_core/viewsets/container.py
+++ b/backend/fms_core/viewsets/container.py
@@ -174,7 +174,8 @@ class ContainerViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePref
         is_parent = _request.GET.get("parent") == 'true'
         is_sample_holding = _request.GET.get("sample_holding") == 'true'
         is_exact_match = _request.GET.get("exact_match") == 'true'
-        except_kinds = _request.GET.get("except_kinds").split(",") if _request.GET.get("except_kinds") else []
+        qs_except_kinds = _request.GET.get("except_kinds")
+        except_kinds = qs_except_kinds.split(",") if qs_except_kinds else []
 
         if search_input is not None:
             if is_exact_match:
--- a/backend/fms_core/viewsets/container.py
+++ b/backend/fms_core/viewsets/container.py
@@ -170,12 +170,13 @@ class ContainerViewSet(viewsets.ModelViewSet, TemplateActionsMixin, TemplatePref
         """
         Searches for parent containers that match the given query
         """
-        search_input = _request.GET.get("q")
+        search_input = _request.GET.get("q", None)
         is_parent = _request.GET.get("parent") == 'true'
         is_sample_holding = _request.GET.get("sample_holding") == 'true'
         is_exact_match = _request.GET.get("exact_match") == 'true'
+        except_kinds = _request.GET.get("except_kinds").split(",") if _request.GET.get("except_kinds") else []
 
-        if search_input:
+        if search_input is not None:
             if is_exact_match:
                 query = Q(barcode=search_input)
                 query.add(Q(name=search_input), Q.OR)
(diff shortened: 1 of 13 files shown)


backend/fms_core/viewsets/container.py
```diff
@@ -170,12 +170,13 @@ def search(self, _request):
         """
         Searches for parent containers that match the given query
         """
-        search_input = _request.GET.get("q")
+        search_input = _request.GET.get("q", None)
         is_parent = _request.GET.get("parent") == 'true'
         is_sample_holding = _request.GET.get("sample_holding") == 'true'
         is_exact_match = _request.GET.get("exact_match") == 'true'
+        except_kinds = _request.GET.get("except_kinds").split(",") if _request.GET.get("except_kinds") else []
```
nafiz1001, line 177: Using `or` would reduce repetition. ;)

backend/fms_core/viewsets/container.py
```diff
@@ -185,9 +186,11 @@ def search(self, _request):
                 query.add(Q(name__icontains=search_input), Q.OR)
                 query.add(Q(id__icontains=search_input), Q.OR)
             if is_parent:
-                query.add(Q(kind__in=PARENT_CONTAINER_KINDS), Q.AND)
+                kinds = [kind for kind in PARENT_CONTAINER_KINDS if kind not in except_kinds]
+                query.add(Q(kind__in=kinds), Q.AND)
             if is_sample_holding:
-                query.add(Q(kind__in=SAMPLE_CONTAINER_KINDS), Q.AND)
+                kinds = [kind for kind in SAMPLE_CONTAINER_KINDS if kind not in except_kinds]
```
nafiz1001, line 192: maybe `kinds` can be assigned outside the `if`s?

frontend/src/components/placementVisuals/AddPlacementContainer.tsx
```diff
@@ -25,31 +25,115 @@ const AddPlacementContainer = ({ onConfirm, destinationContainerList, setDestina
     const [newContainer, setNewContainer] = useState<any>({})
 
     const coordinates = useAppSelector(selectCoordinatesByID)
+    const containerKinds = useAppSelector(selectContainerKindsByID)
+
+    const getContainerKindOptions = useCallback(() => {
+      if (containerKinds) {
+        const options = Object.keys(containerKinds).filter(id => containerKinds[id].children_ids.length == 0 && id != "tube").sort().map((containerKind) => {
```
nafiz1001, line 32: I would do something like `containerKinds[id]?.children_ids` in case the container kind has not been loaded yet

nafiz1001: Also `===` is better than `==`. Javascript :)

nafiz1001: How can `id` be tube? Shouldn't It be an integer?

frontend/src/components/placementVisuals/AddPlacementContainer.tsx
```diff
@@ -25,31 +25,115 @@ const AddPlacementContainer = ({ onConfirm, destinationContainerList, setDestina
...
+
     //calls addDestination prop with 'New Destination' container
     const handleConfirm = useCallback(() => {
-        const container = selectedTab == 'load' ? loadedContainer : newContainer
-        if (container.container_name && container.container_kind && container.container_kind != 'tube') {
-            onConfirm(container)
-            setIsPopup(false)
-            setNewContainer({})
-            setDestinationIndex(destinationContainerList.length)
+        const AddContainer = (container) => {
```
nafiz1001, line 118: functions and (especially) variables should start with lower case ;(

backend/fms_core/viewsets/container.py
UlysseFG: It would cause an error since None does not have a split() method ... i could store the initial result not sure what is really better ... I'll break it down to save on repeated calls.

backend/fms_core/viewsets/container.py
UlysseFG: Mmm  i don't see much gain you would still need to use an `if` on the KINDS to choose from ... and the query only has kind for those 2 cases ...

frontend/src/components/placementVisuals/AddPlacementContainer.tsx
UlysseFG: "kinds" are still part of an older DB light version of Freezeman... They are defined in a Python file (containers.py) and do not have numerical ids like the stuff that is stored in the DB. The id is the kind name in lowercase.

backend/fms_core/viewsets/container.py
nafiz1001: Oh woops didn't notice they use different constants

backend/fms_core/viewsets/container.py
nafiz1001: Oh woops. Didn't notice the split.

frontend/src/components/placementVisuals/PlacementTab.tsx
nafiz1001, line 1: Why does `cellSample` start with lower case 😭

UlysseFG: there is as many ways of naming stuff as there is people ... Most likely because string and number start with lowercase.

PR #730, opened 2024-03-04
Feature #2053:  Improve loading efficiency for the Project Study tab.

nafiz1001:


frontend/src/components/studySamples/CompletedSamplesTable.tsx
```diff
@@ -67,10 +71,26 @@ const COMMENT: CompletedSampleColumn = {
...
+	tableState?: CompletedSamplesTableState
+	settings?: StudyUXStepSettings
+	workflowAction: WorkflowActionType
 }
 
-function CompletedSamplesTable({completedSamples} : CompletedSamplesTableProps) {
+function CompletedSamplesTable({studyID, step, workflowAction, tableState, settings} : CompletedSamplesTableProps) {
+	const dispatch = useAppDispatch()
+
+	const pageSize = settings?.pageSize ?? 10
```
UlysseFG, line 84: This should use a front-end wide default constant instead of a page specific value (10) i think.

frontend/src/modules/studySamples/services.ts
```diff
@@ -163,64 +129,61 @@ export async function buildStudySamplesFromWorkflow(
...
 	return {
-		steps
+		result: completedStudySamples,
+		count: totalCount
 	}
 }
 
-export async function fetchSamplesAtStepOrder(studyID: FMSId, stepOrderID: FMSId) {
-
+async function fetchSamplesAtStepOrder(studyID: FMSId, stepOrderID: number, offset = 0, limit = 10) {
```
UlysseFG, line 161: limit uses a duplication of a harcoded value ... use reference to the same constant as page_size ...

frontend/src/modules/studySamples/services.ts
```diff
@@ -163,64 +129,61 @@ export async function buildStudySamplesFromWorkflow(
...
 
 	// Get the current set of filters and sort order from UX settings for study and step
-	let options = {}
+	let options = { limit } as any
 	const settings = studySettingsByID[studyID]?.stepSettings[stepOrderID]
 	if (settings) {
 		const serializedFilters = settings.filters ? serializeFilterParamsWithDescriptions(settings.filters) : {}
 		const ordering = settings.sortBy ? serializeSortByParams(settings.sortBy) : undefined
-		options = {ordering, ...serializedFilters}
+		const limit = settings.pageSize
```
UlysseFG, line 170: This is weird here ... you get the page size as prop and also fetch it from the settings ... this is likely to cause issues (someone calls the function with a limit and expects to override the settings and instead get slapped back to the settings value ...). I would expect either the function to have limit as prop or fetch it from the settings not both. If both are used the settings.pagSize should be part of the default value (like the first default the second default being 10) and the default value of limit should be set to undefined in order to be able to tell if a value was passed down.

frontend/src/modules/studySamples/services.ts
```diff
@@ -163,64 +129,61 @@ export async function buildStudySamplesFromWorkflow(
...
 			} else {
 				throw new Error('Failed to fetch study samples - no data in response.')
 			}
 		})
-}
\ No newline at end of file
+}
+
+function timestampStringAsDate(created_at: string): any {
+	throw new Error("Function not implemented.")
```
UlysseFG, line 188: ...

frontend/src/modules/studySamples/services.ts
```diff
@@ -163,64 +129,61 @@ export async function buildStudySamplesFromWorkflow(
...
 	if (settings) {
 		const serializedFilters = settings.filters ? serializeFilterParamsWithDescriptions(settings.filters) : {}
 		const ordering = settings.sortBy ? serializeSortByParams(settings.sortBy) : undefined
-		options = {ordering, ...serializedFilters}
+		const limit = settings.pageSize
+		options = {...options, ordering, limit, ...serializedFilters}
 	}
 
-	return store.dispatch(api.sampleNextStepByStudy.getStudySamplesForStepOrder(studyID, stepOrderID, {...options, limit: 100000}))
+	return store.dispatch(api.sampleNextStepByStudy.getStudySamplesForStepOrder(studyID, stepOrderID, {...options, offset}))
```
UlysseFG, line 174: if offset comes from the props so should the limit (they are expected to go together typically ...) otherwise you are likely to have pages that skip or repeat items ...

frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -34,9 +34,10 @@ export interface WorkflowSamplesTableProps {
 		selectedSampleIDs: FMSId[],
 		onSelectionChanged: (selectedSamples: SampleAndLibrary[]) => void
 	}
+	loading?: boolean
```
UlysseFG, line 37: optional ?

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
```diff
@@ -116,14 +102,16 @@ function StudyStepSamplesTable({ studyID, step, settings }: StudyStepSamplesTabl
...
+			samples={step.ready.samples}
 			columns={columns}
 			filterDefinitions={{ ...SAMPLE_COLUMN_FILTERS, ...LIBRARY_COLUMN_FILTERS }}
 			filterKeys={{ ...SAMPLE_NEXT_STEP_BY_STUDY_FILTER_KEYS, ...SAMPLE_NEXT_STEP_BY_STUDY_LIBRARY_FILTER_KEYS }}
 			filters={settings?.filters ?? {}}
 			setFilter={setFilter}
 			setFilterOptions={setFilterOptions}
 			setSortBy={setSortBy}
+			pagination={{ pageNumber, pageSize, totalCount: step.ready.count, onChangePageNumber, onChangePageSize }}
+			loading={tableState?.isFetching}
```
UlysseFG, line 114: test tableState and send either isFetching or true instead in order not to make loading optional ?

frontend/src/modules/studySamples/actions.ts
```diff
@@ -153,12 +163,59 @@ export function setStudyStepSortOrder(studyID: FMSId, stepOrderID: FMSId, sortBy
...
+
+export function setStudyStepPageSize(studyID: FMSId, stepOrderID: FMSId, pageSize: number) {
+	return async (dispatch: AppDispatch) => {
+		dispatch({
+			type: SET_STUDY_STEP_PAGE_SIZE,
+			studyID,
+			stepOrderID,
+			pageSize
+		})
+		dispatch(setStudyStepPageNumberLight(studyID, stepOrderID, 1)) // FIXME: does not actually reset page number :(
```
UlysseFG, line 178: Did it get fixed ? ...

frontend/src/modules/studySamples/reducers.ts
```diff
@@ -131,10 +146,19 @@ export const studySamplesReducer = (state: WritableDraft<StudySamplesState>, act
 				for (const stepOrderID of stepOrderIDs) {
 					studyUXSettings.stepSettings[stepOrderID] = {
 						stepOrderID,
+						pageSize: 10
```
UlysseFG, line 149: again this hard-coded pagesize ... frontend/src/config.js has DEFAULT_PAGINATION_LIMIT that could serve that purpose ... (or you can add DEFAULT_SMALL_PAGINATION_LIMIT if you wish)

frontend/src/components/studySamples/StudyStepSamplesTable.tsx
nafiz1001:
```ts
			loading={tableState?.isFetching ?? true}
```
?

frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
nafiz1001: Yes

UlysseFG (approved):

PR #731, opened 2024-03-05
Convert SampleDetailsContent to typescript and a few other corrections

nafiz1001:


UlysseFG:
Not sure what is happening but the version section in samples seem to be broken and the samples are loading badly (getting some sample.volume not defined)  but it seem odd given that no changes touch this ... It's also freezing sometimes ... Not sure how much help this comment, but there is something breaking in here unless i did mess my version.

PR #732, opened 2024-03-06
Feature 2072 and 2064: Fix Labwork table loading

nafiz1001:
This PR fixes:

- [[BUG]: Not able to enter a full text on Search option on headers in Lab Work protocols](https://206.12.92.46/issues/2072)
- [[BUG]: Fix sorting on Lab Work Samples](https://206.12.92.46/issues/2064)

### DIFF
--- a/frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
+++ b/frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
@@ -34,9 +34,10 @@ export interface WorkflowSamplesTableProps {
 		selectedSampleIDs: FMSId[],
 		onSelectionChanged: (selectedSamples: SampleAndLibrary[]) => void
 	}
+	loading?: boolean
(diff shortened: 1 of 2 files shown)


frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -34,9 +34,10 @@ export interface WorkflowSamplesTableProps {
 		selectedSampleIDs: FMSId[],
 		onSelectionChanged: (selectedSamples: SampleAndLibrary[]) => void
 	}
+	loading?: boolean
```
UlysseFG, line 37: Is this optional required ? this appear to not be optional (and an optional "loading" is likely to lead to weird behaviour).

UlysseFG:
I think the conditional rendering was to prevent some flickering there may be additional stuff to fix under the hood but this is a good first fix ...

PR #735, opened 2024-03-07
Feature #1938:  Fix remaining react console errors for ContainersDetailContent

nafiz1001:


frontend/src/models/frontend_models.ts
```diff
@@ -106,7 +106,9 @@ export function mergeItemsByID<T extends FMSTrackedModel>(itemsByID: ItemsByID<T
 }
 
 export type ObjectId = FMSId
-export interface Container extends Readonly<FMSContainer>, FetchedObject {}
+export interface Container extends Readonly<FMSContainer>, FetchedObject {
+	parents?: Container["id"][]
```
UlysseFG, line 110: Does not appear right to define it here ... wouldn't this be better in fms_api_model FMSContainer ? this seem like a code bit that is there to feed the lint-monster ...

nafiz1001: They way I look at it, `fms_api_model` is strictly for defining fields that are present the response from the API. The `parents` field is actually created in the frontend.

frontend/src/components/containers/ContainersDetailContent.tsx
```diff
@@ -26,46 +31,37 @@ const tabStyle = {
...
+  useEffect(() => {
+    if (isLoaded && !container?.parents)
+      dispatch(listParents(id));
+  }, [isLoaded, container?.parents, dispatch, id])
 
   return (
     <>
       <AppPageHeader
-        title={`Container ${container.name || id}`}
+        title={`Container ${container?.name ?? id}`}
```
UlysseFG, line 64: The "||" would be safer option here since the "??" operator would take "" as a valid value and return it ... This is not possible normally with the container.name but then the id option would never be called on anything but null or undefined.

frontend/src/models/frontend_models.ts
UlysseFG: Yeah as long as we use that representation everywhere in the frontend that matters ... Not sure it was done that way elsewhere (seeing as you created the only deviation from the backend object).

frontend/src/components/containers/ContainersDetailContent.tsx
```diff
@@ -102,10 +100,10 @@ const ContainersDetailContent = ({
 
           <TabPane tab={`Experiment (${experimentRunsIDs?.length})`} key="experiment" style={tabStyle}>
 
-            {containerKindsByID[container.kind] && containerKindsByID[container.kind].is_run_container ?
+            {container && containerKindsByID[container.kind] && containerKindsByID[container.kind].is_run_container ?
```
UlysseFG, line 103: You should not add this test (`container &&`) on the same level as the others since the error message shown does not make sense if the container is just not loaded ...

nafiz1001: what about this?
```tsx
{container &&
              (containerKindsByID[container.kind] && containerKindsByID[container.kind].is_run_container ?
                <ExperimentRunsListSection experimentRunsIDs={experimentRunsIDs} />
                :
                <div> Experiments are not run directly on containers of kind {container?.kind} </div>
              )
            }
```

UlysseFG: yep

UlysseFG (approved):

PR #744, opened 2024-03-20
Place Row and Column Selector

nafiz1001:


frontend/src/components/placementVisuals/PlacementContainer.tsx
```diff
@@ -204,11 +214,19 @@ const PlacementContainer = ({ containerType, columns, rows, samples, direction,
...
-            const cellSize = columns <= 12 ? "cell" : "tiny-cell"
             for (let i = 0; i < rows; i++) {
-                const rowOfCells: React.ReactElement[] = []
+                const charCopy = char.repeat(1)
+		const rowOfCells: React.ReactElement[] = []
                 rowOfCells.push(
-                    <div key={char} className={cellSize} style={{ backgroundColor: '#001529', color: 'white' }}>
+                    <div key={char} className={cellSize} style={{ backgroundColor: '#001529', color: 'white' }} onClick={(e) => {
+		        e.stopPropagation()
+			const samples = [...Array(12).keys()].map((c) => {
```
nafiz1001, line 223: oops i shouldn't hardcode column numbers

UlysseFG:
The top headers are not aligned properly to the rest of the rows (unless it's an optic illusion).

Can you make it so the top left corner cell select all samples in the container ? This does not really match your logic but it would be nice.

frontend/src/components/placementVisuals/PlacementContainer.tsx
```diff
@@ -181,18 +181,43 @@ const PlacementContainer = ({ containerType, columns, rows, samples, direction,
...
             let char = 'A'
             let coordinates = ''
             //renders header based on the number of columns provided to the component
             const headerCells: React.ReactElement[] = []
-            for (let i = 0; i < columns + 1; i++) {
+            const cellSize = columns <= 12 ? "cell" : "tiny-cell"
+	    for (let i = 0; i < columns + 1; i++) {
                 headerCells.push(
-                    <div key={'header_' + i} className={"header"}>
+                    <div key={'header_' + i} className={cellSize} style={{ backgroundColor: '#001529', color: 'white', cursor: 'grab' }} onClick={i !== 0 ? selectColumn(i) : selectAll}>
```
UlysseFG, line 218: Missing the stoppropagation() that prevent activating selection when clicking headers.

nafiz1001: there used to be a stoppropagation there actually, but i found that removing it works better
if i left stoppropagation there, it would not respond unless you click elsewhere to trigger selecting mode (edited)

PR #747, opened 2024-03-21
Block selection while fetching

nafiz1001:


frontend/src/components/labwork/step/LabworkStepOverview.tsx
```diff
@@ -126,7 +134,7 @@ const LabworkStepOverview = ({step, refreshing, setIsSorted, stepSamples, sample
                 filters={filters}
                 setFilter={setFilter}
                 setFilterOptions={setFilterOptions}
-                selection={selection}
+                selection={loading ? noopSelection : selection}
```
UlysseFG, line 137: I don't think the noop can bridge the gap it is on the same flag as the refreshing. If the refresh does not block the selection i doubt this will do anything for us.

UlysseFG: I'll try to trigger the noop with my spasms ...

nafiz1001: so... how did the spamming go?

UlysseFG (approved):

nafiz1001:
Something's not right, selections don't seem to properly show up in the tables

PR #748, opened 2024-03-22
Removed the __exact lookup from from frontend requests.

UlysseFG:


nafiz1001:
The frontend instead performs `icontains` currently
```
http://localhost:9000/api/sample-next-step/?limit=20&offset=0&ordering=&sample__derived_samples__project__name__icontains=Workshop_Biomodal_Q028212&step__id__in=14
```

nafiz1001:
How about instead of removing the exact condition, we just make it an empty branch in the `if (options)`?

frontend/src/components/pagedItemsTable/serializeFilterParamsTS.ts
```diff
@@ -62,7 +62,9 @@ export default function serializeFilterParamsWithDescriptions(filters: FilterSet
 					if (options) {
 						if (options.recursiveMatch) key += '__recursive'
 						else if (options.startsWith) key += '__startswith'
-						else if (options.exactMatch) key += '__exact'
+            else if (options.exactMatch){
+              /* Left blank because exact not uniformly accepted by the backend */
+            }
```
nafiz1001, line 67: indentation :(

nafiz1001 (approved):

PR #756, opened 2024-03-25
Labwork refactor

nafiz1001:


frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -517,4 +437,77 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
...
+	step: Step
+	protocol: Protocol | undefined
+	selection: WorkflowSamplesTableProps['selection']
+	setSortBy: WorkflowSamplesTableProps['setSortBy']
+}
+
+function SelectionTab({stepSamples, step, protocol, selection, setSortBy}: SelectionTabProps) {
+	const dispatch = useAppDispatch()
+
+	const [pageSize, setPageSize] = useState(10)
```
UlysseFG, line 452: what's up with this pesky hardcoded page size of 10 ?

nafiz1001: Oh nooooo

frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -517,4 +437,77 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
 	)
 }
 
-export default LabworkStep
+
+export interface SelectionTabProps {
```
UlysseFG, line 441: might be worth moving to its own file ...

frontend/src/hooks.ts
```diff
@@ -1,6 +1,44 @@
...
+import { selectLibrariesByID, selectSamplesByID } from './selectors'
+import { SampleAndLibrary } from './components/WorkflowSamplesTable/ColumnSets'
 
 // Use throughout your app instead of plain `useDispatch` and `useSelector`
 export const useAppDispatch: () => AppDispatch = useDispatch
-export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
\ No newline at end of file
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
+
+export function useSampleList(sampleIDs: FMSId[], offset?: number, size?: number) {
```
UlysseFG, line 13: i personally like more precise variable names ... size would be clearer as page_size ... (at least to me). Up to you ...

UlysseFG:
The list of samples in the placement tab under the visualization do not follow the order from the selection tab ... At first i thought it would not matter, but the way it is at the moment is more confusing than anything... I think the source list should be ordered as requested in the Selection Tab. The destination list should retain their relative ordering within the destination container. Even when we hit the selection tab to force ordering the sample orders in the lists are by order of id. The current prod version is doing the same, but maybe this can be improved. How do you stand on that ? We could create a new ticket for next version ...

UlysseFG:
When selecting a full plate in samples then going to selection tab and unselecting a single sample you go from 96 to 9 samples selected ... This is clearly a bug. setSelectedSamplesFromRow only takes the current page selection to set the SamplesSelected.

nafiz1001:
> I realized something about destroying the placement tab when not active ... it also destroys any in progress placement ... This is something maybe we should analyze and discuss with Andres see where we stand on that... I personally liked the idea i could go back to the samples and add or remove a few samples and they would be added/removed to the placement tab without having to start placement over from scratch. This might not be worth the eventual cost but with faster Summary this might be acceptable.

This can be potentially solved by refactoring placement so that it uses redux instead ;)

nafiz1001:
I just realized someone already implemented a service function called `fetchSamplesAndLibraries`. I'd like to incorporate that into `useSamplesAndLibraries` to reduce duplicate code.

nafiz1001:
I should also consider adding another list called sortedSelectedSamples to avoid overwriting selectedSamples. I forgot why.

UlysseFG (approved):

PR #762, opened 2024-04-03
Placement Refactor

nafiz1001:
image


nafiz1001:
Do not fear the large diff. It's mostly package-lock.json 😉

nafiz1001:
Should I get rid of cypress? The tests seem outdated.

frontend/src/reducers.js
```diff
@@ -188,6 +189,7 @@ const allReducers = combineReducers({
   sampleNextStepTemplateActions,
   steps,
   notifications,
```
nafiz1001, line 191: i remember working on that, and I think no one uses that ;'(

UlysseFG:
The list below the visualization do not put the selected samples first ... It makes it harder to know which samples were selected. I think the original version did that. 
The list is also randomly ordered making it very hard to find a sample using the list.
I would order the list by coordinates (as second ordering after the selection status).

UlysseFG:
The popup while trying to place selected samples are distracting more than useful. It would be nice to deactivate popup when a sample are selected to be placed.
Also the popups should follow the currently used format container@coordinates or separate them (the wrapping of text in the popup is not great) instead of coordinates@container.
Sample freezeman ID in the popup is not useful (lab users will not know these). Sample name would be best (You could even include project). Of course, it is already better than nothing so do not remove the popup if these can't be fetched in an efficient manner.

UlysseFG:
When loading an existing container, we need to block the coordinates that are already occupied by samples (and include a visual cue that those are taken and cannot be moved). At the moment, i could do placement to coordinates that are invalid.

UlysseFG (approved):

PR #768, opened 2024-04-15
Feature #2119:  [BUG] When opening sample details for a library when comming from somewhere else than the libray listing, the library details are missing.

nafiz1001:


frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -139,47 +112,56 @@ const SampleDetailsContent = ({
...
+    if (isLoaded && container?.experiment_run) {
+      setExperimentRunIDs((experimentRunsIDs) => [...experimentRunsIDs, container.experiment_run])
+    }
+  }, [container?.experiment_run, experimentRunsIDs, isLoaded])
 
-  if (!librariesByID[id])
-    getLibrary(id)
+  useEffect(() => {
+    if (!librariesByID[id])
+      dispatch(getLibrary(id))
```
nafiz1001, line 156: We just forgot to dispatch `getLibrary`.

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -139,47 +112,56 @@ const SampleDetailsContent = ({
...
+  }, [dispatch, id, librariesByID])
 
   useEffect(() => {
     const biosampleId = sample?.biosample_id
     listSampleMetadata(token, { "biosample__id": biosampleId }).then(metadata => {
       setSampleMetadata(metadata)
     })
-  }, [sample])
+  }, [sample, token])
```
UlysseFG, line 164: do you really want to refresh the sample metadata each time  the token expires and is getting replaced ?

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -139,40 +112,49 @@ const SampleDetailsContent = ({
...
+      })
+    }
+  }, [isLoaded, isProcessesEmpty, sample.process_measurements])
 
-  if (isLoaded && container?.experiment_run) {
-    experimentRunsIDs.push(container.experiment_run)
-  }
+  useEffect(() => {
+    if (isLoaded && container?.experiment_run) {
+      setExperimentRunIDs((experimentRunsIDs) => [...experimentRunsIDs, container.experiment_run])
```
nafiz1001, line 150: Infinite loop of dead

UlysseFG: Looks like it's getting run twice ... adding twice the same id to the list ... then when it reload it loads correctly. 
I really don't see how a single sample could be part of more than one experiment. In the current way it is designed it should always have either zero or one experiment run associated ... I would get rid of the array maybe even of the state ...

nafiz1001: Should be fixed by now.

UlysseFG (approved):

PR #770, opened 2024-04-16
Prefilling override for Axiom experiment template.

UlysseFG:


backend/fms_core/prefilling_functions.py
nafiz1001, line 1: Getting barcode from comment? That's a little strange. Especially when comments are usually free text.

backend/fms_core/template_prefiller/prefiller.py
```diff
@@ -23,9 +23,13 @@ def PrefillTemplate(template_path, template_info, queryset):
     for sheet_name, sheet_dict in position_dict.items():
         current_sheet = workbook[sheet_name]
         for i, entry in enumerate(queryset.values(*filter(lambda x: x is not None, sheet_dict["queryset_column_list"]))):
-            for prefill_sheet_name, template_column, queryset_column, _ in template_info["prefill info"]:
```
nafiz1001, line 26: A function inside `template_info`?

backend/fms_core/prefilling_functions.py
nafiz1001: This probable should be unit tested if it hasn't been done already ;)

UlysseFG: Well it was hand tested but maybe adding a unit test for the regex function might be good ...

backend/fms_core/template_prefiller/prefiller.py
UlysseFG: power of the interpreted language ... The select case used to not exist in python because you can just make a dictionary with functions in it ... Now they decided to add lambda and all kind of redundant stuff much to my dislike.

PR #771, opened 2024-04-19
Feature #2141:  Add column on Dataset tab and menu for "Validated" runs

nafiz1001:
- [ ] maybe check for opportunities to reuse `validation_status` from dataset object in the frontend


backend/fms_core/serializers.py
```diff
@@ -608,6 +609,17 @@ def get_released_status_count(self, obj):
     
     def get_blocked_status_count(self, obj):
         return Readset.objects.filter(dataset=obj.id, release_status=ReleaseStatus.BLOCKED).count()
+
+    def get_validation_status(self, obj):
+        counts = Dataset.objects.filter(pk=obj.id) \
```
UlysseFG, line 614: This is clearly not the most effective way to do this ...

UlysseFG: I would move this logic to the model (as a property also in case we have more than one place that requires (maybe in services) this same information.

frontend/src/components/datasets/DatasetTable.tsx
```diff
@@ -45,6 +46,17 @@ const getTableColumns = () => {
             dataIndex: "lane",
             sorter: true,
         },
+        {
+            title: "Validated",
```
UlysseFG, line 50: "Validated" does not really match the content of the column ... Maybe "Validation Status"

frontend/src/components/datasets/DatasetTable.tsx
```diff
@@ -45,6 +46,17 @@ const getTableColumns = () => {
...
+            dataIndex: "validation_status",
+            render: (validation_status) => {
+                return (
+                    (validation_status === 1 && <Button style={{color: "#a0d911"}}><CheckOutlined onPointerEnterCapture={undefined} onPointerLeaveCapture={undefined}/>Passed</Button>)
```
UlysseFG, line 54: the onPointer... undefined callbacks look like defaults. Do they serve a purpose here ?

UlysseFG, line 54: Maybe an enumerated type in constants.tsx to make the status matching a bit more elegant ...

nafiz1001: they make my linter stop complaining :)

nafiz1001:
I messed up my commit history

frontend/src/components/datasets/DatasetTable.tsx
UlysseFG: in that case ... sorry mr linter ...

UlysseFG (approved):

PR #776, opened 2024-04-25
Fix post-sample-next-step

nafiz1001:
Fixes #769

### DIFF
--- a/backend/fms_core/viewsets/sample_next_step.py
+++ b/backend/fms_core/viewsets/sample_next_step.py
@@ -204,12 +204,12 @@ class SampleNextStepViewSet(viewsets.ModelViewSet, TemplateActionsMixin, Templat
 
         if self.request.method == 'POST':
             sample__id__in = None
-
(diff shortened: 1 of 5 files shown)


backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -367,8 +378,9 @@ def labwork_step_info(self, request, *args, **kwargs):
             }
           }
         """
-        step_id = request.GET.get('step__id__in')
-        grouping_column = request.GET.get('group_by')
+        params = QueryDict(request.META.get('QUERY_STRING'))
+        step_id = params.get('step__id__in')
+        grouping_column = params.get('group_by')
```
nafiz1001, line 383: Should I be worried about the type of `step_id` and `grouping_column` 🤔

nafiz1001: Should `step_id` be `int` here 🤔

PR #781, opened 2024-05-01
[Side Project I Swear] Convert api.js to api.ts

nafiz1001:


frontend/src/components/samples/details/SampleDetailsLineage.tsx
```diff
@@ -86,10 +86,10 @@ function SampleDetailsLineage({sample, handleSampleClick, handleProcessClick} :
   const [nodesToEdges, setNodesToEdges] = useState<{ [key: string]: SampleLineageGraphProcessMeasurement }> ({})
   
   useEffect(() => {
-    if (sample.id === undefined) {
-      return;
-    }
     (async () => {
```
UlysseFG, line 89: is there any point in having an async callback inside an empty not async callback ?

frontend/src/utils/api.ts
```diff
@@ -436,7 +456,18 @@ function createAPIError(response) {
...
+interface FMSResponse<T = any> extends Response {
+  isJSON: boolean
+  data: T
+  filename?: string
+}
+interface JsonResponse<T = any> extends FMSResponse<T> { isJSON: true }
+interface ArrayBufferResponse extends FMSResponse<ArrayBuffer> { isJSON: false }
+interface StringResponse extends FMSResponse<string> { isJSON: false }
+interface EmptyResponse extends FMSResponse<Record<string, never>> { isJSON: false }
+type ResponseWithData<T = any> = JsonResponse<T> | ArrayBufferResponse | StringResponse | EmptyResponse
```
UlysseFG, line 468: Not sure i see the point of a type of responsewithdata that includes empty responses ... Maybe call it AnythingGoesResponse ?

nafiz1001: It's an anomylous response because that kind of response is only returned when reading from the response body fails. I guess I could call it AttachDataErrorResponse 🤷

nafiz1001: So far I have never seen that kind of response, and most frontend codes assume successful body read.

nafiz1001: Hmm... maybe I should throw an exception instead of a body every fails to be read or use `Promise.reject`.

frontend/src/components/samples/details/SampleDetailsLineage.tsx
nafiz1001: Well inside `useEffect` is the only appropriate place for executing async function (in a component), and `useEffect` cannot take an async callback as argument, so that pattern was kinda necessary unless we give that async callback a name.

UlysseFG (approved):

frontend/src/utils/api.ts
```diff
@@ -295,42 +296,49 @@ const api = {
...
   }
 };
 
 
 export default api;
 
-export function withToken(token, fn) {
-  return (...args) => fn(...args)(undefined, () => ({ auth: { tokens: { access: token } } }))
+type WithTokenFn<R extends ResponseWithData<any>> = (...args: any[]) => (_: undefined, getState: () => AuthTokensAccess) => Promise<R>
+export function withToken<R extends ResponseWithData<any>>(token: string | undefined, fn: WithTokenFn<R>) {
```
UlysseFG, line 326: won't this logoff someone if you call it without token ? Why would you call withToken withoutToken ??

UlysseFG: I see that you could already call withToken with an undefined token, but it might actually be a good thing to prevent it and to get some linter error ?

PR #787, opened 2024-05-02
Remove unneeded planning template from frontend template list.

UlysseFG:


frontend/src/modules/labworkSteps/actions.ts
```diff
@@ -47,7 +47,16 @@ export function initSamplesAtStep(stepID: FMSId) {
...
+      const templatesResponse = await dispatch(api.sampleNextStep.prefill.templates(protocol.id)).then(
+        (response) => {
+          let filteredResponse = { ...response }
+          if (!step.needs_planning){
+            filteredResponse.data = response.data.filter((template) => {return !template.description.includes("planning")})
+          }
+          return response
+        }
+      )
+
```
nafiz1001, line 59: Um... what's the user of `filteredResponse`?

nafiz1001: I believe you can just use `response.data` directly

UlysseFG: err looks like i pushed the wrong version ... filteredresponse is supposed to be returned. Not sure how this ended up being pushed. Let me fix that... response.data would not fix the issue...

nafiz1001 (approved):

PR #789, opened 2024-05-06
Move the sample loading after a filter change to its own useEffect.

UlysseFG:


nafiz1001 (approved):
The change makes sense, and the code works. So I approve 👍🏼

PR #798, opened 2024-05-21
Fix Placement

nafiz1001:


frontend/src/modules/placement/reducers.ts
```diff
@@ -442,30 +411,42 @@ function placementDestinationLocations(state: PlacementState, sources: Draft<Cel
...
                 const a = sources[indexA]
                 const b = sources[indexB]
-                const sampleA = a.sample
-                const sampleB = b.sample
-                if (sampleA === null) {
-                    throw new Error(`Cell at coordinates ${atCellLocations(a)} has no sample`)
-                }
-                if (sampleB === null) {
-                    throw new Error(`Cell at coordinates ${atCellLocations(b)} has no sample`)
+                const offsetsA = a.coordinates ? coordinatesToOffsets(getContainer(state, a).spec, a.coordinates) : []
```
UlysseFG, line 417: Might be worth extracting that offset comparison as a function seeing as it is recurring...

frontend/src/modules/placement/reducers.ts
```diff
@@ -442,30 +411,42 @@ function placementDestinationLocations(state: PlacementState, sources: Draft<Cel
...
+            }, {})
+
+            for (const sourceIndex in sources) {
+                const relativeOffset = relativeOffsetByIndices[sourceIndex]
+                const [startingRow, startingCol] = destinationStartingOffsets                
+
+                const finalOffsets = [startingRow, startingCol]
+
+                if (placementOptions.direction === PlacementDirections.ROW) {
+                    finalOffsets[1] += relativeOffset
```
UlysseFG, line 433: Row=0, Col=1 ? Might clarify with Constant or enum (...or type ?!)

UlysseFG (approved):
You can take my comments in considerations, but they are optional ...

PR #802, opened 2024-05-27
Tweak Placement

nafiz1001:
https://206.12.92.46/issues/2171

- ~[] React Error Boundary~
- ~[ ] Numbering cells during placement~
- [x] Sort samples also by project, especially for tubes without parent
- [x] Show container origin name only on the destination side
- [x] hide coordinates column for tubes without parent
- ~[ ] Show tube name on source side for tubes without parent?~

### DIFF
--- a/backend/fms_core/viewsets/sample_next_step.py
+++ b/backend/fms_core/viewsets/sample_next_step.py
@@ -399,14 +399,34 @@ class SampleNextStepViewSet(viewsets.ModelViewSet, TemplateActionsMixin, Templat
 
         grouped_step_samples = self.filter_queryset(self.get_queryset())
         # Get all samples on the steps with the grouping field
-        grouped_step_samples = grouped_step_samples.filter(step__id__exact=step_id).values_list("sample_id", grouping_column, "ordering_container_barcode", "ordering_container_coordinates")
+        grouped_step_samples = grouped_step_samples.filter(step__id__exact=step_id) \
+            .annotate(project_name=F("sample_next_step_by_study__study__project__name")) \
+            .annotate(sample_name=F("sample__name")) \
+            .values_list(
+                "sample_id",
+                "sample_name",
+                "project_name",
+                grouping_column,
+                "ordering_container_barcode",
+                "ordering_container_coordinates"
+            )
 
         groups = defaultdict(list)
         # Extract the locators from the entries
-        for sample_id, group_column, container_barcode, container_coordinates in grouped_step_samples.all():
-            groups[group_column].append({"sample_id": sample_id, "contextual_container_barcode": container_barcode, "contextual_coordinates": container_coordinates})
+        for sample_id, sample_name, project_name, group_column, container_barcode, container_coordinates in grouped_step_samples.all():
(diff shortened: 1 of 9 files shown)


nafiz1001:
I haven't tested yet. But I believe it should be ready.

backend/fms_core/viewsets/sample_next_step.py
```diff
@@ -399,14 +399,34 @@ def labwork_step_info(self, request: HttpRequest, *args, **kwargs):
...
         grouped_step_samples = self.filter_queryset(self.get_queryset())
         # Get all samples on the steps with the grouping field
-        grouped_step_samples = grouped_step_samples.filter(step__id__exact=step_id).values_list("sample_id", grouping_column, "ordering_container_barcode", "ordering_container_coordinates")
+        grouped_step_samples = grouped_step_samples.filter(step__id__exact=step_id) \
+            .annotate(project_name=F("sample_next_step_by_study__study__project__name")) \
+            .annotate(sample_name=F("sample__name")) \
+            .values_list(
+                "sample_id",
+                "sample_name",
+                "project_name",
```
UlysseFG, line 408: This is not slowing the query too much on the samples tab ?

nafiz1001: I tried dozens of container, and it took a few seconds to load of the containers.
But yeah it would be nice to optimize this when we go big.

nafiz1001:
Numbered order ftw

frontend/src/utils/functions.ts
```diff
@@ -56,4 +57,37 @@ export function compareArray(a: readonly number[], b: readonly number[]): number
...
+        const bOffsets = coordinatesToOffsets(spec, b.coordinates)
+        const arrayComparison = compareArray(aOffsets.reverse(), bOffsets.reverse())
+        if (arrayComparison > 0) orderB -= MAX/4
+        else if (arrayComparison < 0) orderA -= MAX/4
+    }
+
+    if (a.name > b.name) orderB -= MAX/8
+    else if (a.name < b.name) orderA -= MAX/8
+
+    // if (!a.parentContainerName) orderA -= MAX/16
```
UlysseFG, line 83: remove those if you don't use them ...

UlysseFG (approved):

frontend/src/components/placementVisuals/PlacementSamplesTable.tsx
```diff
@@ -151,19 +121,43 @@ const PlacementSamplesTable = ({ container: containerName, showContainerColumn }
...
+
     return (
         <Table<PlacementSample>
-            dataSource={samples}
+            dataSource={samples.filter(sample => !sample.placed)}
             columns={columns.filter(column => (
+                // normally for destination side only
                 column.key !== 'parentContainerName' || showContainerColumn
             ) && (
+                // only show coordinates if in a parent container
```
UlysseFG, line 159: visibility of columns may be more readable as a "visible" column attribute.

UlysseFG (approved):

PR #804, opened 2024-06-17
Remove the unique counter that was added to the pool name to make the barcode unique.

UlysseFG:


nafiz1001 (approved):
What was the reason again the lab didn't want unique name for their pool?

UlysseFG:
> What was the reason again the lab didn't want unique name for their pool?

this was confusing them... the counter was added after their own pool counter ... This is possibly going to be changed back when they start getting collisions.

PR #808, opened 2024-07-15
Changes to transfer template and template definition.

UlysseFG:


backend/fms_core/tests/test_template_importers/sample_pooling_planning_test.py
nafiz1001, line 1: Ideally, the second index should be a constant but it's too late I guess

backend/fms_core/template_importer/row_handlers/transfer/transfer.py
nafiz1001, line 1: We need better variable names, but I'll assume it's good ;)

backend/fms_core/tests/test_template_importers/transfer_test.py
nafiz1001, line 1: We need better variable names, but I'll assume it's good if it passes ;)

nafiz1001 (approved):

backend/fms_core/tests/test_template_importers/sample_pooling_planning_test.py
UlysseFG: too late ? which index ? TWO = 2 ?

nafiz1001: More like
```python
self.assertEqual(csv_content[1][ROBOT_SRC_BARCODE], "Src1")
```

PR #809, opened 2024-07-18
Placement Tweaks Tweaks

nafiz1001:
- [x] show selection count under tables
- [x] show 'Tubes without parent' in table and in cell popup
- [x] show project name for existing destination container added
- [x] disable selection for samples already present on the destination side (by selecting through the tables)

### DIFF
--- a/frontend/src/components/placementVisuals/AddPlacementContainer.tsx
+++ b/frontend/src/components/placementVisuals/AddPlacementContainer.tsx
@@ -8,6 +8,7 @@ import Input from "antd/lib/input/Input"
 import api from "../../utils/api"
 import { MAX_CONTAINER_BARCODE_LENGTH, MAX_CONTAINER_NAME_LENGTH, barcodeRules, nameRules } from "../../constants"
 import { FMSContainer, FMSId, FMSSample } from "../../models/fms_api_models"
+import { get as getProject } from "../../modules/projects/actions"
 
 export interface DestinationContainer {
     container_barcode: string
@@ -59,13 +60,17 @@ const AddPlacementContainer = ({ onConfirm, existingContainers }: AddPlacementCo
         if (container.samples.length > 0) {
             const loadedSamples: FMSSample[] = (await dispatch(api.samples.list({ id__in: container.samples.join(','), limit: 100000 }))).data.results
             const projectIDs = new Set(loadedSamples.map(sample => sample.project))
-            await dispatch(api.projects.list({ id__in: Array.from(projectIDs).join(',') }))
+            for (const projectID of projectIDs) {
+                if (projectID) {
+                    dispatch(getProject(projectID))
+                }
+            }
             loadedSamples.forEach(sample => {
                 newDestination[sample.id] = {
                     id: sample.id,
-                    coordinates: sample.coordinate ? coordinates[sample.coordinate].name : '...',
+                    coordinates: sample.coordinate ? coordinates[sample.coordinate]?.name ?? '...' : '',
                     name: sample.name,
-                    project: sample.project ? projects[sample.project].name : '...'
+                    project: sample.project ? projects[sample.project]?.name ?? '...' : ''
                 }
             })
         }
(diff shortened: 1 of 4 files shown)


frontend/src/components/placementVisuals/AddPlacementContainer.tsx
```diff
@@ -58,12 +59,29 @@ const AddPlacementContainer = ({ onConfirm, existingContainers }: AddPlacementCo
...
+            const coordinates = selectCoordinatesByID(store.getState())
+            const projects = selectProjectsByID(store.getState())
+
             loadedSamples.forEach(sample => {
-                newDestination[sample.id] = { id: sample.id, coordinates: (sample.coordinate && coordinates[sample.coordinate].name) as string, name: sample.name }
+                newDestination[sample.id] = {
+                    id: sample.id,
+                    coordinates: sample.coordinate ? coordinates[sample.coordinate]?.name ?? '...' : 'N/A',
+                    name: sample.name,
+                    project: sample.project ? projects[sample.project]?.name ?? '...' : 'N/A'
```
UlysseFG, line 79: not a fan of N/A ... empty or undefined would be good... same goes for coordinates ...

frontend/src/components/placementVisuals/PlacementSamplesTable.tsx
```diff
@@ -121,7 +121,21 @@ const PlacementSamplesTable = ({ container: containerName, showContainerColumn }
...
+        })
+    }), [selectedRowKeys, onChange, onSelect, isDestination, containerName])
+
+    const paginationProps: NonNullable<TableProps<PlacementSample>['pagination']> = useMemo(() => ({
+        showSizeChanger: true,
+        showTotal(total, range) {
+            return <>
+                <>{`${range[0]}-${range[1]} of ${total} items.`}</>
+                <>{' '}</>
+                <>{`${selectedRowKeys.length} selected.`}</>
```
UlysseFG, line 135: It looks like it takes precious space a button of the pagination (button to go to the next page) wrap around to the next line on my screen. Not sure this indicator is worth breaking the harmony of the display ...

frontend/src/components/placementVisuals/AddPlacementContainer.tsx
```diff
@@ -58,12 +59,29 @@ const AddPlacementContainer = ({ onConfirm, existingContainers }: AddPlacementCo
...
+            const coordinates = selectCoordinatesByID(store.getState())
+            const projects = selectProjectsByID(store.getState())
+
             loadedSamples.forEach(sample => {
-                newDestination[sample.id] = { id: sample.id, coordinates: (sample.coordinate && coordinates[sample.coordinate].name) as string, name: sample.name }
+                newDestination[sample.id] = {
+                    id: sample.id,
+                    coordinates: sample.coordinate ? coordinates[sample.coordinate]?.name ?? '...' : 'undefined',
+                    name: sample.name,
+                    project: sample.project ? projects[sample.project]?.name ?? '...' : 'undefined'
```
UlysseFG, line 79: ... ?! the string 'undefined'  ... what about leaving it empty ? and not 'empty' ... or at least ''.

nafiz1001: Fair enough. I just want the user to know that there is **definitely** no project associated with the sample.

UlysseFG: ok well it's not that important. we should have some standard way to say there is none ... i think some places we use "-". Not sure what would be best ...The value of the project name can be Any typical character combination so even "NA" and "undefined" are valid project names ... maybe a special icon we could use for that ?

frontend/src/components/placementVisuals/PlacementSamplesTable.tsx
nafiz1001: So shall I undo this?

UlysseFG: Up to you ... I don't see a great way to replace this to let people know they selected something if they are not on the first page...

UlysseFG (approved):

PR #812, opened 2024-07-25
added filter to retrieve latest released readsets by dataset

samouzegar10:
changes required to have the notification service on triton to work

### DIFF
--- a/backend/fms_core/filters.py
+++ b/backend/fms_core/filters.py
@@ -1,5 +1,6 @@
-from django.db.models import Q
-
+from django.db.models import Q, Max
+from django.utils import timezone
+import  datetime
 from .models import (Container,
                      DerivedBySample,
                      Index,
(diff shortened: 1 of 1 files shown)


backend/fms_core/filters.py
```diff
@@ -175,10 +176,17 @@ class Meta:
 
 class DatasetFilter(GenericFilter):
     release_flag = django_filters.NumberFilter(method="release_flag_filter")
+    latest_release_update = django_filters.CharFilter(method="latest_release_update_filter")
 
     def release_flag_filter(self, queryset, name, value):
         return queryset.filter(release_flag=value)
 
+    def latest_release_update_filter(self,queryset, name, value):
```
UlysseFG, line 184: your filter is not generic enough... Maybe make a filter released_within_hours and capture the value passed to the filter so you can filter from X hours to the current time instead of a fixed time ?

backend/fms_core/filters.py
```diff
@@ -1,5 +1,6 @@
-from django.db.models import Q
-
+from django.db.models import Q, Max
+from django.utils import timezone
+import  datetime
```
UlysseFG, line 3: extra space

backend/fms_core/filters.py
```diff
@@ -175,10 +177,16 @@ class Meta:
...
     release_flag = django_filters.NumberFilter(method="release_flag_filter")
+    latest_release_update = django_filters.CharFilter(method="latest_release_update_filter")
 
     def release_flag_filter(self, queryset, name, value):
         return queryset.filter(release_flag=value)
 
+    def latest_release_update_filter(self, queryset, name, value):
+        return queryset.annotate(
+            latest_release_update=Max("readsets__release_status_timestamp")
+        ).filter(latest_release_update__gte=value)
```
UlysseFG, line 188: if the time format of value is not the same as the one on freezeman this could cause issues but it is acceptable on our side ... Might be safer for the API to keep asking for a given period of time from the moment of the request ...

UlysseFG (approved):

nafiz1001 (approved):

PR #814, opened 2024-07-29
Fix Last Update

nafiz1001:


nafiz1001:
image

frontend/webpack.config.js
```diff
@@ -60,7 +60,7 @@ module.exports = (env, argv) => ({
     new webpack.DefinePlugin({
       GIT_COMMITHASH: JSON.stringify(gitRevisionPlugin.commithash()),
       GIT_BRANCH: JSON.stringify(gitRevisionPlugin.branch()),
-      GIT_LASTUPDATE: JSON.stringify(child_process.execSync('git log -1 --format=%cI').toString().trim()),
+      GIT_LASTUPDATE: JSON.stringify(child_process.execSync('git log -1 --format=%cD').toString().trim()),
```
UlysseFG, line 63: i like much more the iso format ... %ci          
You could also cut off after the date to keep it simple.

nafiz1001:
How about this?
image

UlysseFG:
Only worry is that the time might not be from the right timezone.

PR #816, opened 2024-07-31
Link Sample To Study

nafiz1001:


backend/fms_core/viewsets/project.py
```diff
@@ -59,4 +62,26 @@ def summary(self, _request):
...
+        })
+    
+    @action(detail=False, methods=["post"])
+    def add_samples_to_study(self, request, pk=None):
+        sample_ids = request.data.get("sample_ids")
+        project_id = request.data.get("project_id")
+        study_letter = request.data.get("study_letter")
+        step_order = request.data.get("step_order", None)
+
+        samples =  make_generator(Sample.objects.filter(id__in=sample_ids))
```
UlysseFG, line 74: don't think you need make_generator here as much as you like it ...

backend/fms_core/viewsets/project.py
```diff
@@ -59,4 +62,26 @@ def summary(self, _request):
...
+        project = Project.objects.get(id=project_id)
+
+        errors = defaultdict(list)
+        for sample in samples:
+            _errors, _ = add_sample_to_study(sample, project, study_letter, step_order)
+            for key, error in _errors.items():
+                if error:
```
UlysseFG, line 81: You set falsy errors in add_sample_to_study?

frontend/src/components/projects/LinkSamplesToStudy.tsx
```diff
@@ -0,0 +1,114 @@
...
+                        {studies.map(study =>
+                            <Select.Option
+                                key={study.id}
+                                value={study.letter}
+                            >
+                                {study.letter}
+                            </Select.Option>)
+                        }
+                    </Select>
+                    <Select
```
UlysseFG, line 97: Leave some space between the selection fields.

frontend/src/components/projects/LinkSamplesToStudy.tsx
```diff
@@ -0,0 +1,114 @@
...
+                }
+            }}
+            onCancel={() => {
+                if (handleCancel) {
+                    handleCancel()
+                }
+            }}
+        >
+            <Form>
+                <Form.Item label={"Study Letter"}>
```
UlysseFG, line 84: Maybe just call it "Study" the letter part is obvious from the dropdown content.

frontend/src/components/projects/LinkSamplesToStudy.tsx
UlysseFG: Do you have to include both form fields in the same form.item ? The second field would benefit from a label too to make it clear that it defines the target step.

frontend/src/components/projects/LinkSamplesToStudy.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    }
+                }
+            }}
+            onCancel={() => {
+                if (handleCancel) {
+                    handleCancel()
+                }
+            }}
+        >
+            <Form>
```
UlysseFG, line 83: Include a small sentence to indicate how many samples will be linked.

backend/fms_core/viewsets/project.py
nafiz1001: empty list

nafiz1001: that might be a bug

UlysseFG: Mmm you may get key = [] sometime when we call error["patate"] = do_that(). it would still extend correctly i think though.

UlysseFG:
Selection of more than a page at a time is not possible ... i think there was a complication to using the small check box in the header to do a select all but at least when you change page it should remember you other pages selections ...

UlysseFG:
Another issue that might arise is the simplicity of the error message ... if you select 100 samples you have no idea which one failed and for what reason ...

UlysseFG:
It may be useful to have a filtered column with the container name to facilitate selection using filters...

nafiz1001:
> Selection of more than a page at a time is not possible ... i think there was a complication to using the small check box in the header to do a select all but at least when you change page it should remember you other pages selections ...

I was able to achieve something like that in the PlacementSamplesTable. It was a bit hacky.

UlysseFG:
> I was able to achieve something like that in the PlacementSamplesTable. It was a bit hacky.

It may be time to reuse your hack ...

backend/fms_core/viewsets/project.py
UlysseFG: i think you might need to call ".all()" after your ".filter()" to iterate through the returned objects.

backend/fms_core/viewsets/project.py
nafiz1001: Ah now I remember why. It's so I add add any new entries in the `errors` when the error list is empty.

nafiz1001:
I'm thinking about renaming `selectAll` with `defaultSelection` and `selectedItems` with `exceptions`.

backend/fms_core/services/project.py
```diff
@@ -39,4 +43,38 @@ def create_project(name=None, principal_investigator=None, requestor_name=None,
     except ValidationError as e:
         errors.append(str(e))
 
-    return (project, errors, warnings)
\ No newline at end of file
+    return (project, errors, warnings)
+
+def add_sample_to_study(sample: Sample, project: Project, study_letter: str, step_order: int | None = None):
```
UlysseFG, line 48: Include a comment for the automated documentation and add tests for the service.

backend/fms_core/viewsets/project.py
```diff
@@ -59,4 +62,30 @@ def summary(self, _request):
...
+    @action(detail=False, methods=["post"])
+    def add_samples_to_study(self, request, pk=None):
+        sample_ids = request.data.get("sample_ids")
+        select_all = request.data.get("select_all", False)
+        project_id = request.data.get("project_id")
+        study_letter = request.data.get("study_letter")
+        step_order = request.data.get("step_order", None)
+
+        samples = (Sample.objects.filter(derived_by_samples__project=project_id, id__in=sample_ids) if not select_all
+                   else Sample.objects.filter(derived_by_samples__project                   =project_id).exclude(id__in=sample_ids)).all()
```
UlysseFG, line 76: weird spacing in here ... not sure if the line break make much sense other than to accommodate line length limits...

backend/fms_core/services/project.py
nafiz1001: Automated documentation... I totally forgot we had that, even though I worked on that...

frontend/src/components/projects/LinkSamplesToStudy.tsx
```diff
@@ -0,0 +1,125 @@
...
+                                id: NOTIFICATION_ID,
+                                title: "Samples linked to study",
+                                description: `Successfully linked samples to study ${study.letter} at step "${stepOrder.step_name}"`
+                            }))
+                            if (handleSuccess) {
+                                handleSuccess()
+                            }
+                        },
+                        () => {
+                            dispatch(notifyError({
```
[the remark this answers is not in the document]
UlysseFG: How to do it is a more delicate issue... i would personally keep the simple notification for failure but add to the response a "failure log" that can be downloaded by the frontend. Not sure if this is the typical way to go but i am not sure a temporary model would help since it might close before the user have a time to check the content. This is just an idea we can go another way but we need to have in mind the user has little information on the validation done in the backend to queue up a sample.

frontend/src/components/projects/ProjectsAssociatedSamples.tsx
```diff
@@ -64,14 +139,21 @@ export const ProjectsAssociatedSamples = ({ projectID: currentProjectID } : Proj
         SAMPLE_COLUMN_DEFINITIONS.VOLUME,
         SAMPLE_COLUMN_DEFINITIONS.CREATION_DATE,
         SAMPLE_COLUMN_DEFINITIONS.DEPLETED,
+        {
+            columnID: 'STUDY-STEP',
+            title: 'Study - Step',
```
UlysseFG, line 144: "Study step" would be better this is the current study step being displayed. No need to hyphenation i think.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -133,18 +135,76 @@ function PagedItemsTable<T extends object>({
...
+	// key. This allows us to use data objects that don't have an explicit 'id' property.
+	const UNKOWN_DATA_ROW_KEY = 'unknown-data-row-key'
+	const getRowKeyForDataObject = useCallback((data: T) => {
+		for (const key in tableDataState.objectMap) {
+			if (tableDataState.objectMap[key] === data) {
+				return key
+			}
 		}
-	}
+		return UNKOWN_DATA_ROW_KEY
```
UlysseFG, line 148: Is it possible to have a table row without key ? Is it the best behaviour to create a false key that could mimic as a real key. From my understanding, it could lead to a selection silently missing items ... The missing keys are ignored when listing the selection. If it is not possible, this code should not be there, if it is possible it should not behave that way because we want this to be obvious there is something fishy.

frontend/src/components/projects/ProjectsAssociatedSamples.tsx
```diff
@@ -40,22 +42,95 @@ const useLastProtocols = (sampleIDs: readonly Sample['id'][]) => {
...
+            refresh(sampleIDs)
+        }
+    }, [refresh, sampleIDs])
+
+    const StudySteps = useCallback(({ sampleID }: { sampleID: Sample['id'] }) => {
+        if (sampleID in studyStepsBySampleID) {
+            return <>{
+                studyStepsBySampleID[sampleID]?.map((studyStep) =>
+                    studiesByID[studyStep.study] && orderByStepOrder[studyStep.step_order]
+                        ? `${studiesByID[studyStep.study]}-${orderByStepOrder[studyStep.step_order]}`
```
UlysseFG, line 106: Would be nice to have a tooltip when mousing over each study step pair that would give the step name.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -169,27 +229,17 @@ function PagedItemsTable<T extends object>({
...
 	return (
 		<>
 			{columns && (
 				<>
-					{usingFilters && pagedItems.filters && (
-						<FiltersBar filters={pagedItems.filters} clearFilters={clearFiltersCallback}></FiltersBar>
-					)}
-					<Table
+					<div>
+						{topBarExtra}
```
UlysseFG, line 237: It is a bit weird not to have the Link to Study button not vertically aligned with the Clear Filter button.

frontend/src/components/projects/ProjectsAssociatedSamples.tsx
nafiz1001: I have feeling that would look worse if a sample is assigned to multiple studies, considering that the column is just a comma-separated set of study-step.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
nafiz1001: > if it is possible it should not behave that way because we want this to be obvious there is something fishy
Throw exception?

nafiz1001:
```ts
	const onSelectSingle = useCallback((record: T) => {
		const key = getRowKeyForDataObject(record)
		if (key !== UNKOWN_DATA_ROW_KEY) {
			let newSelectedItems = selectedItems
			if (selectedItems.includes(key)) {
				newSelectedItems = selectedItems.filter((id) => id !== key)
			} else {
				newSelectedItems = [...selectedItems, key]
			}
			setSelectedItems(newSelectedItems)
			if (selection) {
				selection.onSelectionChanged(newSelectedItems, selectAll)
			}
		}
	}, [getRowKeyForDataObject, selectAll, selectedItems, selection])

```

frontend/src/components/projects/ProjectsAssociatedSamples.tsx
UlysseFG: Ok ... it looks weird any way for multiple steps ... the list format is not totally satisfying but i am not sure how we would replace that.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
UlysseFG: throw notification i would say in case there is no valid key i would notify the user the selection failed for some unknown reason ... To me this looks like coding for bad coding ... I don't see why we would have a table with no key and use this function to look for a key ... I personally would just remove the default key and if the key is undefined fail more or less gracefully but obviously so we can fix that case.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
nafiz1001: I don't like frontend development.

backend/fms_core/services/project.py
```diff
@@ -39,4 +42,57 @@ def create_project(name=None, principal_investigator=None, requestor_name=None,
...
+            _, errors['queue_sample_to_study_workflow'], warnings['queue_sample_to_study_workflow'] = queue_sample_to_study_workflow(sample, study_obj, step_order)
+        else:
+            return f"Sample [{sample.name}] is already queued in study [{study_letter}] of project [{project.name}] at step [{step_order}]."
+
+
+    error = add_sample_to_study_(sample, project, study_letter, step_order)
+    if error:
+        errors['add_sample_to_study'].append(error)
+
+    dict_remove_falsy_entries(errors)
```
UlysseFG, line 95: Not sure this belongs inside the service ... i would move these to the viewset maybe ... This looks like extra work to provide a user with a cleaned up list of errors and warnings not for internal usage.

backend/fms_core/services/project.py
```diff
@@ -39,4 +42,57 @@ def create_project(name=None, principal_investigator=None, requestor_name=None,
...
+        step_order: Step order of a step in the study. Defaults to None.
+
+    Returns:
+        Tuple of errors and warnings
+    """
+
+    errors = defaultdict(list)
+    warnings = defaultdict(list)
+
+    def add_sample_to_study_(sample: Sample, project: Project, study_letter: str, step_order: int | None = None):
```
UlysseFG, line 64: Why make an internal call here to a function is everything in that service ? It looks silly and makes no sense to me ...

UlysseFG: It seems to also break the previous way of dealing with the errors in the template (it could previously report more than one error by row ...). I think you treat this like javascript while this is not the function inside does not have access to the same errors as the external one ... the error management you do here is likely not working in python ...

backend/fms_core/tests/test_services/project_add_samples_to_study_test.py
```diff
@@ -0,0 +1,95 @@
...
+        errors, warnings = add_sample_to_study(self.sample_project1, self.project2, self.study_project2_A.letter)
+        self.assertEqual(errors.get("add_sample_to_study"), [f"Sample [{self.sample_project1.name}] is not linked to project [{self.project2.name}]."])
+        self.assertEqual(dict(warnings), {})
+    
+    def test_sample_cannot_queue_to_the_same_step_twice(self):
+        errors, warnings = add_sample_to_study(self.sample_project1, self.project1, self.study_project1_A.letter, self.qc_step)
+        errors, warnings = add_sample_to_study(self.sample_project1, self.project1, self.study_project1_A.letter, self.qc_step)
+        self.assertEqual(errors.get("add_sample_to_study"), [f"Sample [{self.sample_project1.name}] is already queued in study [{self.study_project1_A.letter}] of project [{self.project1.name}] at step [{self.qc_step}]."])
+        self.assertEqual(dict(warnings), {})
+
```
UlysseFG, line 95: Add a test to verify "is_sample_queued_in_study" errors are carried over ...

UlysseFG (approved):

PR #817, opened 2024-07-31
Change novaseq x smallest container name in the template choices.

UlysseFG:


nafiz1001 (approved):
> the best code is no code at all.

\- Jeff Atwood

PR #818, opened 2024-07-31
Allow REPEAT_STEP on workflow samples.

UlysseFG:
- [x] Fix fetch function to get all completed in a single call... Currently break limit and offset.
- [x] Add test to verify the repeat is done correctly.


backend/fms_core/services/sample_next_step.py
```diff
@@ -256,7 +256,7 @@ def has_sample_completed_study(sample_obj: Sample, study_obj: Study) -> Tuple[Un
             errors.append(f"No step found for the given order.")
 
         # If the sample has completed the workflow, the StepHistory for the last step order in the study should
-        # have a workflow action of NEXT_STEP
+        # have a workflow action of NEXT_STEP or REPEAT_STEP
         if StepHistory.objects.filter(process_measurement__lineage__child=sample_obj, # for step with child
                                       study=study_obj, 
                                       step_order=step_order,
```
nafiz1001, line 262: Where's the filter for `REPEAT_STEP`?

backend/fms_core/tests/test_services/sample_next_step_test.py
```diff
@@ -292,6 +292,61 @@ def test_move_sample_to_next_step(self):
...
+        letter_B = "B"
+        start = 3
+        end = 7
+        study_B = Study.objects.create(letter=letter_B,
+                                       project=self.project,
+                                       workflow=self.workflow_pcr_free,
+                                       start=start,
+                                       end=end)
+
+        letter_C = "C"
```
nafiz1001, line 319: Do you actually need a `letter_C` variable, especially when you already have a `study_C` variable?

backend/fms_core/tests/test_services/sample_next_step_test.py
```diff
@@ -292,6 +292,61 @@ def test_move_sample_to_next_step(self):
...
+        protocol = Protocol.objects.get(name="Normalization")
+        process = Process.objects.create(protocol=protocol)
+        process_measurement = ProcessMeasurement.objects.create(process=process,
+                                                                source_sample=sample_in,
+                                                                execution_date=datetime.date(2021, 1, 10),
+                                                                volume_used=10)
+
+        old_sample_to_study_workflow_1, _, _ = queue_sample_to_study_workflow(sample_in, study_B)
+        old_sample_to_study_workflow_2, _, _ = queue_sample_to_study_workflow(sample_in, study_C)
+        new_sample_next_steps, errors, warnings = move_sample_to_next_step(step, sample_in, process_measurement, WorkflowAction.REPEAT_STEP, keep_current=True)
```
nafiz1001, line 337: Still kinda weird that a sample sample can be in two different studies and be moved to the next at the same time in two different studies

frontend/src/modules/studySamples/services.ts
```diff
@@ -109,8 +109,9 @@ export async function fetchSamplesAndLibraries(sampleList: number[]) {
 	return availableSamples
 }
 
-export async function fetchCompletedSamples(studyID: FMSId, stepOrderID: FMSId, workflow_action: WorkflowActionType, limit: number, offset: number) {
-	const stepHistoryResponse = await store.dispatch(api.stepHistory.getCompletedSamplesForStudy(studyID, {step_order__id__in: stepOrderID, limit, offset, workflow_action}))
+export async function fetchCompletedSamples(studyID: FMSId, stepOrderID: FMSId, workflow_actions: WorkflowActionType[], limit: number, offset: number) {
+  const qsWorkflowActions = workflow_actions.join(" ")
+	const stepHistoryResponse = await store.dispatch(api.stepHistory.getCompletedSamplesForStudy(studyID, {step_order__id__in: stepOrderID, limit, offset, workflow_action: qsWorkflowActions}))
```
nafiz1001, line 114: If you want to be extra cool, you can make the type more precise for getCompletedSamplesForStudy ;)
```ts
    getCompletedSamplesForStudy: (studyId, options) => get<JsonResponse<FMSPagedResultsReponse<any>>>('/step-histories/', {...options, study__id__in: studyId}),
```

nafiz1001:
```

File: frontend/src/modules/studySamples/services.ts (hunk 4 above)
Comment (UlysseFG) [reply frontend/src/modules/studySamples/services.ts:114]: nope i am not a cool type guy

File: backend/fms_core/tests/test_services/sample_next_step_test.py (hunk 3 above)
Comment (UlysseFG) [reply backend/fms_core/tests/test_services/sample_next_step_test.py:337]: you could want to do more than one thing with a sample ...

File: backend/fms_core/tests/test_services/sample_next_step_test.py (hunk 2 above)
Comment (UlysseFG) [reply backend/fms_core/tests/test_services/sample_next_step_test.py:319]: It is so that we can have on hand a not hard-coded way to reference the study in tests (which has a letter as key). It is not particularly useful but it seem a good idea nevertheless.

File: backend/fms_core/services/sample_next_step.py (hunk 1 above)
Comment (UlysseFG) [reply backend/fms_core/services/sample_next_step.py:262]: I think it went the way of the dodo ...

File: backend/fms_core/services/sample_next_step.py [hunk 5]
```diff
@@ -256,7 +256,7 @@ def has_sample_completed_study(sample_obj: Sample, study_obj: Study) -> Tuple[Un
             errors.append(f"No step found for the given order.")
 
         # If the sample has completed the workflow, the StepHistory for the last step order in the study should
-        # have a workflow action of NEXT_STEP
+        # have a workflow action of NEXT_STEP or REPEAT_STEP
         if StepHistory.objects.filter(process_measurement__lineage__child=sample_obj, # for step with child
                                       study=study_obj,
```
nafiz1001, line 261: Maybe we should reuse parts of the querysets? Like extract out querysets representing steps without children and with children, and then filter by workflow_action?

nafiz1001 (approved):

backend/fms_core/services/sample_next_step.py
UlysseFG: not sure django optimizing is not really that obvious since most of the work building the query is done by the framework... We would have to benchmark the different options ... (i would personally instead possibly try to retrieve all step_history then parse them to make sure none if there is any is next_step or repeat_step ... but as far as i know most query will never actually execute the third and fourth option since they will be true after the first one ...

nafiz1001 (approved):
Quite not what I asked for, but LGTM 👍🏼

PR #820, opened 2024-08-05
Sample submission uses generic individuals to carry sex and genome information for human and mouse.

UlysseFG:
- [x] Add default reference genome to taxons.
- [x] Initialize default reference genomes for humans (9606) and mouse (10090).
- [x] Creation of is_generic individuals using the prefix "GENERIC_" during sample submission template ingestion when missing name.
- [x] Rename existing "GENERIC_" individuals to "GEN_" and move "GENERIC_" name to alias.
- [x] Update sample submission test to test generic individuals.


backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -16,6 +16,9 @@
 from fms_core.services.index import get_index
 
 class SampleRowHandler(GenericRowHandler):
+    HUMAN_TAXON_ID = 9606
+    MOUSE_TAXON_ID = 10090
```
nafiz1001, line 20: Erm... You sure it's good in the long run to hardcode those?

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -59,10 +62,18 @@ def process_row_inner(self, sample, library, container, project, parent_containe
                                individual["reference_genome"],
                                mother_obj,
                                father_obj])
+        can_use_generic_individual = (any([taxon_obj is not None and taxon_obj.ncbi_id == self.HUMAN_TAXON_ID,
+                                           taxon_obj is not None and taxon_obj.ncbi_id == self.MOUSE_TAXON_ID,
+                                           reference_genome_obj is not None and reference_genome_obj.taxon.ncbi_id == self.HUMAN_TAXON_ID,
+                                           reference_genome_obj is not None and reference_genome_obj.taxon.ncbi_id == self.MOUSE_TAXON_ID])
```
nafiz1001, line 68: This one is especially problematic

nafiz1001: Can we somehow have can_use_generic_individual as a field/column for taxon or reference genome?

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
UlysseFG: Nah i am pretty sure it's not great i just have a hard time finding a good way around ... it all comes down to harcoding human and mouse in some form ... I could add another attribute to the taxon to indicate they are handle differently but it would be just a more ugly (DB hardcoding ?) solution in my eyes unless we can make in generic somehow ...

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
UlysseFG: That one is not problematic ... this is related to user input ... it is tied to the harcoded value but the harcoded value must not be can_use_generic_individual ... the fact there is a taxon_obj is due to user setting a value in a template column ...

backend/fms_core/tests/test_models/individual_test.py
```diff
@@ -55,3 +57,21 @@ def test_pedigree(self):
...
+        self.assertTrue(generic_individual.is_generic)
+        individual = Individual.objects.create(**create_individual(individual_name="GEN_joeblo", is_generic=False))
+        self.assertEqual(Individual.objects.count(), self.generic_individual_count + 2)
+        self.assertEqual(str(individual), "GEN_joeblo")
+        self.assertFalse(individual.is_generic)
+
+    def test_creating_generic_individual_without_flag(self):
+        with self.assertRaises(ValidationError):
+            try:
+                generic_individual = Individual.objects.create(**create_individual(individual_name="GENERIC_joeblo"))
```
nafiz1001, line 74: Unnecessary variable assignment to `generic_individual`?

backend/fms_core/models/individual.py
```diff
@@ -42,6 +44,7 @@ class Individual(TrackedModel):
     cohort = models.CharField(max_length=200, blank=True, help_text="Name to group some individuals in a specific study.")
     alias = models.CharField(blank=True, null=True, max_length=200, help_text="Original individual name used by external client.")
     reference_genome = models.ForeignKey(ReferenceGenome, null=True, blank=True, on_delete=models.PROTECT, related_name="individuals", help_text="Reference genome used to analyze samples.")
+    is_generic = models.BooleanField(default=False, help_text="Generic individual used to replace undefined individuals that share characteristics.")
```
nafiz1001, line 48: Is it actually useful to have a `is_generic` field instead of just looking up individual by name prefix?

backend/fms_core/models/individual.py
```diff
@@ -92,6 +95,12 @@ def add_error(field: str, error: str):
         if self.mother_id is not None and self.mother.sex == self.SEX_MALE:
             add_error("mother", "Mother cannot be of male sex.")
 
+        if self.is_generic and self.name[:len(self.GENERIC_INDIVIDUAL_PREFIX)] != self.GENERIC_INDIVIDUAL_PREFIX:
+            add_error("is_generic", f"'{self.GENERIC_INDIVIDUAL_PREFIX}' name prefix not used for generic individual.")
+
+        if not self.is_generic and self.name[:len(self.GENERIC_INDIVIDUAL_PREFIX)] == self.GENERIC_INDIVIDUAL_PREFIX:
+            add_error("is_generic", f"'{self.GENERIC_INDIVIDUAL_PREFIX}' name prefix cannot be used for non generic individuals.")
+
```
nafiz1001, line 103: Maybe we can reuse certain expressions such as
```python
has_generic_individual_prefix = self.name[:len(self.GENERIC_INDIVIDUAL_PREFIX)] == self.GENERIC_INDIVIDUAL_PREFIX
```

nafiz1001:
Maybe instead of renaming existing GENERIC_ individuals to GEN_ we change uppercase GENERIC to lowercase generic 🤔

backend/fms_core/tests/test_models/individual_test.py
UlysseFG: yeah ... i can remove it ...

backend/fms_core/models/individual.py
UlysseFG: it's much faster much. but also does not required business logic ... we can anywhere (frontend for exemple) check if the individual is generic without having to carry over the prefix format and test for it...

backend/fms_core/models/individual.py
UlysseFG: I could store that somewhere but i would rather this stays there and people rely on the attribute ... it's a maybe i'll think about it.

UlysseFG:
> Maybe instead of renaming existing GENERIC_ individuals to GEN_ we change uppercase GENERIC to lowercase generic 🤔

This is a safety measure to remove possible existing individual with that prefix to remove possible confusion ... if we change the prefix we still need to rename the existing individuals (at the moment i expect none but ...).

backend/fms_core/tests/test_template_importers/sample_submission_test.py
```diff
@@ -83,7 +83,7 @@ def test_import(self):
             derived_sample_id = DerivedBySample.objects.filter(sample_id=sample_obj.id).first().derived_sample_id
             biosample = DerivedSample.objects.get(id=derived_sample_id).biosample
             if biosample.individual.is_generic:
-                self.assertEqual(biosample.individual.name, generic_individual)
+                self.assertIn(generic_individual_prefix, biosample.individual.name)
```
nafiz1001: we should instead be asserting that the individual name starts with prefix, not that it just contains the prefix.

backend/fms_core/tests/test_template_importers/sample_submission_test.py
```diff
@@ -79,8 +82,11 @@ def test_import(self):
             sample_obj = Sample.objects.get(name=sample['name'])
             derived_sample_id = DerivedBySample.objects.filter(sample_id=sample_obj.id).first().derived_sample_id
             biosample = DerivedSample.objects.get(id=derived_sample_id).biosample
-            self.assertEqual(biosample.individual.name, individual_name)
-            self.assertEqual(biosample.individual.alias, individual_alias)
+            if biosample.individual.is_generic:
+                self.assertIn(generic_individual_prefix, biosample.individual.name)
```
nafiz1001, line 86: assertIn no good

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -59,27 +60,41 @@ def process_row_inner(self, sample, library, container, project, parent_containe
                                individual["reference_genome"],
                                mother_obj,
                                father_obj])
+        can_use_generic_individual = (any([taxon_obj is not None,
+                                           reference_genome_obj is not None])
+                                      and not any([individual["pedigree"],
+                                                   mother_obj,
+                                                   father_obj]))
```
nafiz1001, line 67: A comment explaining the reason for this condition would be nice.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -59,27 +60,41 @@ def process_row_inner(self, sample, library, container, project, parent_containe
...
+                self.errors['individual'].append(f"Individual cohort requires an individual name or taxon to be provided to be saved.")
             if individual["alias"]:
-                self.errors['individual'].append(f"Individual alias requires an individual name to be provided to be saved.")
-            if individual["reference_genome"]:
-                self.errors['individual'].append(f"Individual reference genome requires an individual name to be provided to be saved.")
+                self.errors['individual'].append(f"Individual alias requires an individual name or taxon to be provided to be saved.")
             if mother_obj:
                 self.errors['individual'].append(f"Individual mother requires an individual name to be provided to be saved.")
             if father_obj:
                 self.errors['individual'].append(f"Individual father requires an individual name to be provided to be saved.")
```
nafiz1001, line 83: There's great potential for refactoring this.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -59,27 +60,41 @@ def process_row_inner(self, sample, library, container, project, parent_containe
...
         # When the individual name is not provided any field that is stored on the individual need to raise an error.
         self.errors['individual'] = []
         self.warnings['individual'] = []
-        if not individual["name"] and need_individual:
-            if individual["taxon"]:
-                self.errors['individual'].append(f"Individual taxon requires an individual name to be provided to be saved.")
+        if not individual["name"] and need_individual and not can_use_generic_individual:
             if individual["sex"]:
-                self.errors['individual'].append(f"Individual sex requires an individual name to be provided to be saved.")
+                self.errors['individual'].append(f"Individual sex requires an individual name or taxon to be provided to be saved.")
```
nafiz1001, line 73: Why 'or taxon' when you haven't aren't changing for taxon? This could get in the way of refactoring :(

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -59,27 +60,41 @@ def process_row_inner(self, sample, library, container, project, parent_containe
...
+                                      and not any([individual["pedigree"],
+                                                   mother_obj,
+                                                   father_obj]))
         # When the individual name is not provided any field that is stored on the individual need to raise an error.
         self.errors['individual'] = []
         self.warnings['individual'] = []
-        if not individual["name"] and need_individual:
-            if individual["taxon"]:
-                self.errors['individual'].append(f"Individual taxon requires an individual name to be provided to be saved.")
+        if not individual["name"] and need_individual and not can_use_generic_individual:
```
nafiz1001, line 71: This if condition is kinda hard to wrap my mind around. I think it should be in a variable.

nafiz1001 (approved):
My brain keeps melting each time I try to review and understand each line of codes. So I approve.

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
UlysseFG: There is not much to be gained by refactoring this ... The taxon here is a best guess on what was missing on the using submission ... There is also possibly the pedigree and parents messing things up but it get complex fast explaining that here ... i rely on the other error messages to carry that meaning ... If you only get that message you are likely missing a taxon... if you also have a pedigree error you should give a name to your individual ... (one error requires name or taxon the other a name). Not sure there is a simple way to get this across to the user.

PR #824, opened 2024-08-12
Prevent edition to empty of the external id to store an empty string.

UlysseFG:


frontend/src/components/projects/ProjectEditContent.tsx
nafiz1001, line 1: I have a strong urge to type `newValues`

frontend/src/components/projects/ProjectEditContent.tsx
```diff
@@ -214,6 +214,9 @@ function serialize(values) {
   else
     newValues.targeted_end_date = null
 
+  if (!newValues.external_id || newValues.external_id.length == 0)
+    newValues.external_id = null
+
```
nafiz1001, line 219: so external_id either a an empty string, number 0 or an empty array?

UlysseFG: > so external_id is either a an empty string, a number 0, a null or an empty array?
    newValues.external_id = null
The form uses empty string as a default to an empty string instead of null. The backend accepts an empty string as a valid blank value. The ordering make a mess of things ...

backend/fms_core/models/project.py
```diff
@@ -24,7 +24,7 @@ class Project(TrackedModel):
     status = models.CharField(choices=((type, type) for type in PROJECT_STATUS_CHOICES), max_length=20, default="Open",
                               help_text="The status of the project.")
 
-    external_id = models.CharField(blank=True, null=True,  max_length=200, help_text="Identifier to connect to an external system.")
+    external_id = models.CharField(blank=True, null=True, max_length=200, help_text="Identifier to connect to an external system.")
```
nafiz1001, line 27: Shouldn't you enforce against empty string in the model?

nafiz1001:
https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null
> Avoid using [null](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) and [TextField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.TextField). If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

nafiz1001 (approved):
Although (in addition to my changes) it looks good, Django's notice kinda makes me hesitant to merge this.

frontend/src/components/projects/ProjectEditContent.tsx
samouzegar10, line 1: i agree with nafiz's comments, this looks to me a like a form validation that can be taken care of on the front end.

PR #825, opened 2024-08-15
Prevent tooltip clipping by enlarging bars when low sample count.

UlysseFG:


nafiz1001:
The title is misleading. The tooltip is always enlarged, not when sample count is low.

nafiz1001:
There. The title is more accurate now 🙂

UlysseFG:
Tried allowEscapeViewbox ... as far as i could see it was not helping because the graph was not large enough. I'll test you fix but i am doubtful (also it much cleaner to have a full page graph than the limited one.

nafiz1001 (approved):
I'll just go with fat bars.

nafiz1001 (approved):
So far looks good with 64 samples and 3 samples.

PR #827, opened 2024-08-17
Helpspot 31169: Fix Placement Bug

nafiz1001:


frontend/src/components/placementVisuals/PlacementSamplesTable.tsx
```diff
@@ -175,13 +178,7 @@ const PlacementSamplesTable = ({ container: containerName, showContainerColumn }
...
     return (
         <Table<PlacementSample>
-            dataSource={samples.filter(sample => !sample.placed).map(
-            (sample) => 
-                ({
-                    ...sample,
-                    parentContainerName: sample.parentContainerName ?? 'Tubes without parent' 
-                })
-            )}
+            dataSource={samples}
```
nafiz1001, line 181: I already filter at line 46

samouzegar10 (commented):
in terms of general code readability, it seems good, some concerns in the change of MAX/32 -> MAX/16, not sure what this would mean in terms of the app itself but as long as the behavior of the app is good and since this is an urgent hotfix, seems good to me

PR #831, opened 2024-08-29
Addition of validated_by and released_by to readset model.

UlysseFG:


backend/fms_core/migrations/0064_v4_11_0.py
```diff
@@ -58,6 +60,40 @@ def initialize_default_reference_genome(apps, schema_editor):
...
+            validated_by_id = None
+            released_by_id = None
+            for readset_validated_version in Version.objects.filter(content_type__model="readset", object_id=readset_id).order_by('-id'):
+                data = json.loads(readset_validated_version.serialized_data)
+                validation_status_timestamp = data[0]["fields"].get("validation_status_timestamp", None)
+                release_status_timestamp = data[0]["fields"].get("release_status_timestamp", None)
+                if validated_by_id is None and validation_status_timestamp is not None:
+                    validated_by_id = data[0]["fields"]["updated_by"]
+                if released_by_id is None and release_status_timestamp is not None:
+                    released_by_id = data[0]["fields"]["updated_by"]
```
nafiz1001, line 88: Maybe you can go in the opposite direction (`order_by('id')`) so you can exit this loop sooner?

backend/fms_core/models/individual.py
```diff
@@ -44,7 +44,7 @@ class Individual(TrackedModel):
     cohort = models.CharField(max_length=200, blank=True, help_text="Name to group some individuals in a specific study.")
     alias = models.CharField(blank=True, null=True, max_length=200, help_text="Original individual name used by external client.")
     reference_genome = models.ForeignKey(ReferenceGenome, null=True, blank=True, on_delete=models.PROTECT, related_name="individuals", help_text="Reference genome used to analyze samples.")
-    is_generic = models.BooleanField(default=False, help_text="Generic individual used to replace undefined individuals that share characteristics.")
+    is_generic = models.BooleanField(default=False, help_text="Flag indicating a generic individual used to replace undefined individuals that share characteristics.")
```
nafiz1001, line 47: Off-topic PR 😱

backend/fms_core/migrations/0064_v4_11_0.py
UlysseFG: object id are given in order of creation ... latest ones are most recent ... i want the latest one not the earliest. It is a bit lazy since it does not validate if both information are required during the loop but i don't expect millions of versions...

backend/fms_core/models/individual.py
UlysseFG: This is a stealth fix to a model change i did merge some time ago.

UlysseFG: Looks like stealth is not my forte...

nafiz1001 (approved):
You've addressed both of my comments, so you might be worthy of my approval.

backend/fms_core/models/dataset.py
```diff
@@ -31,17 +31,29 @@ def validation_status(self) -> Optional[ValidationStatus]:
...
+    def validated_by(self):
+        readset = self.readsets.first()
+        if readset.validated_by:
+            return readset.validated_by.username
+
+    @property
+    def released_by(self):
+        readset = self.readsets.first()
+        if readset.released_by:
+            return readset.released_by.username
```
nafiz1001, line 44: Risky thing if we have existing datasets that are partially released

backend/fms_core/filters.py
```diff
@@ -189,6 +190,12 @@ def latest_release_update_filter(self, queryset, name, value):
             latest_release_update=Max("readsets__release_status_timestamp")
         ).filter(latest_release_update__gt=value)
 
+    def latest_validation_status_update_filter(self, queryset, name, value):
+        return queryset.annotate(
+            latest_validation_status_update=Max("readsets__validation_status_timestamp")
+        ).filter(latest_validation_status_update__gt=value)
+
+
```
nafiz1001, line 198: Wasn't this code from sebastian's PR?

nafiz1001: Yes. Yes it is.

backend/fms_core/models/dataset.py
UlysseFG: released_by and validated_by are used to indicate progression in the processing of the dataset more than indicate the detail portrait ... These essentially only there to report who did the release of the dataset content (be it release or blocking).

backend/fms_core/filters.py
UlysseFG: yeah i spotted some cross polination in your PR sebastian be careful where you spread it ... you can play in my PR but make sure not to add unrelated stuff of your own (i keep that right to myself ;).

samouzegar10: the other pr was really just the endpoint and in order to test ulysses changes out, i decided to move the code here, 
the options proposed:
1. leave here and delete my pr (since this is the only change that it has)
2. i can remove this one line and we push the other pr and this one separately

any option goes for me. i would always go for simplicity but i leave to the commenters to decide

backend/fms_core/models/dataset.py
samouzegar10: in this case i propose to change all of these following properties, found in the model/dataset.py 


    @property
    def validation_status(self) -> Optional[ValidationStatus]:
        readset = self.readsets.first()
        if readset:
            return readset.validation_status

    @property
    def validated_by(self):
        readset = self.readsets.first()
        if readset.validated_by:
            return readset.validated_by.username

    @property
    def released_by(self):
        readset = self.readsets.first()
        if readset.released_by:
            return readset.released_by.username
            
since i used the already existing validation status property logic

backend/fms_core/filters.py
samouzegar10: i closed the other pr and kept the validation related code in this PR

backend/fms_core/filters.py
```diff
@@ -180,6 +180,7 @@ class Meta:
 class DatasetFilter(GenericFilter):
     release_flag = django_filters.NumberFilter(method="release_flag_filter")
     latest_release_update = django_filters.CharFilter(method="latest_release_update_filter")
+    latest_validation_status_update = django_filters.CharFilter(method="latest_validation_status_update_filter")
```
UlysseFG, line 183: use the same naming pattern as the release one.

backend/fms_core/serializers.py
```diff
@@ -590,44 +589,60 @@ class Meta:
         fields = "__all__"
 
 class DatasetSerializer(serializers.ModelSerializer):
-    files = serializers.SerializerMethodField()
+    # files = serializers.SerializerMethodField()
```
UlysseFG, line 592: Why are files commented out ?

backend/fms_core/serializers.py
```diff
@@ -590,32 +589,48 @@ class Meta:
...
-    
+
     def get_latest_release_update(self, obj):
         return Readset.objects.filter(dataset=obj.id).aggregate(Max("release_status_timestamp"))["release_status_timestamp__max"]
-    
+
     def get_readset_count(self, obj):
         return Readset.objects.filter(dataset=obj.id).count()
 
+    def get_validation_status_update_timestamp(self, obj):
```
UlysseFG, line 622: Keep the same naming pattern as release. Both are so similar i would prefer them to have a naming that reflect that.

backend/fms_core/serializers.py
```diff
@@ -590,32 +589,48 @@ class Meta:
...
+    latest_release_update = serializers.SerializerMethodField()
+    released_by = serializers.SerializerMethodField()
+    validation_status = serializers.SerializerMethodField()
+    validation_status_update_timestamp = serializers.SerializerMethodField()
+    validated_by = serializers.SerializerMethodField()
 
     class Meta:
         model = Dataset
-        fields = ("id", "external_project_id", "run_name", "lane", "files", "released_status_count", "blocked_status_count", "latest_release_update", "validation_status", "project_name", "metric_report_url", "readset_count", "archived_comments")
+        fields = "__all__"
```
UlysseFG, line 605: If you serialize more fields make sure they are added to the model in the frontend. Here released_by, validated_by, to_be_replaced(validation_status_update_timestamp) and experiment_run are not in fms_api_models.ts

backend/fms_core/filters.py
```diff
@@ -191,6 +191,7 @@ def latest_release_update_filter(self, queryset, name, value):
         ).filter(latest_release_update__gt=value)
 
     def latest_validation_update_filter(self, queryset, name, value):
+        print('GET IN THERE YOU!!')
```
UlysseFG, line 194: yeah ?

PR #834, opened 2024-09-10
Use name in taxon column

nafiz1001:


UlysseFG:
This will need to be rebased ... Also the taxon value in the field should be straight-up the taxon name stored in the db maybe with some basic transform like lower case or capitalize .

PR #839, opened 2024-09-16
added debounce on sortBy

samouzegar10:


samouzegar10:
the current fix aborts any ongoing api call. 
aborting the last ongoing is functional.
in samples page, the same initial call happens twice at the beginning for some reason.

frontend/src/components/filters/filterComponents/DebouncedInput.jsx
nafiz1001, line 1: This your opportunity to convert this to typescript 😉

frontend/src/modules/samples/actions.ts
nafiz1001, line 1: This seems like an unnecessary change, but I'd let it pass since it's very small.

frontend/src/utils/api.ts
nafiz1001, line 1: This is not your code but I find it strange that we delete an entry in `ongoingRequests` when we want to abort. I feel like it should be the opposite, but I'd verify beforehand.

```ts
        return request.then(res => {
            if (options.abort) {
                delete ongoingRequests[baseRoute]
            }
            return res
})
```

samouzegar10: when you say, it should be the opposite, can you elaborate?

nafiz1001: Oh wait. I see now that the callback only gets executed after the request has been resolved. So this is just cleaning up the entry for later request.

samouzegar10: indeed, firstly the abortController does its thing (assign the signal, right now theres an empty message and then it executes abort(), this ongoingRequests[baseRoute] is a "global" variable used to keep track.

frontend/src/components/filters/filterComponents/DebouncedInput.jsx
samouzegar10: the change was very minimal and the app seems to be working as if nothing changed

nafiz1001: I didn't mean just change file. I meant also adding adjusting types here and there. Just enough to eliminate any `any`s and type errors.

samouzegar10: ah, you had a certain view of what to do. 
want to have a conversation about it tomorrow :) ?

nafiz1001: Sure

frontend/src/modules/samples/actions.ts
```diff
@@ -39,8 +39,10 @@ export const update = (id, sample) => async (dispatch, getState) => {
 export const list = (options) => async (dispatch: AppDispatch, getState) => {
     const params = { limit: 100000, ...options }
     return await dispatch(networkAction(LIST,
-        api.samples.list(params),
-        { meta: params }
+        api.samples.list(params, { abort: true }),
+        {
+            meta: params,
+        }
```
nafiz1001, line 45: Now that I think about it. This might be risky because there might be different React components that depends on sample list API call. So if multiple components fire this API at once, one of them will be aborted which is no good. But I don't know any page on top of my head that could cause this.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -251,7 +247,7 @@ function PagedItemsTable<T extends object>({
 						columns={columns}
 						rowKey={getRowKeyForDataObject}
 						scroll={{x: 300}}
-						onChange={sortByCallback}
+						onChange={useDebounce(sortByCallback)}
```
nafiz1001, line 250: Be careful about conditional rendering. `useDebounce` is a hook which must always be executed for each re-rendering.

frontend/src/modules/samples/actions.ts
samouzegar10: it would have been risky, although, its pin pointed down to the samples list content page (table) even if lets say, another page would use it, the call itself will be safe until he makes a 2nd call on the same api call. its pretty safe as is :) im testing locally to see if anything changes in the app and i havent seen any changes in behavior

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
samouzegar10: sure no problem! this is sound and clear

samouzegar10: no major changes either, all is working as it should.

frontend/src/utils/api.ts
```diff
@@ -362,7 +362,7 @@ function apiFetch<R extends ResponseWithData<any>>(method: HTTPMethod, route: st
             const controller = new AbortController()
             signal = controller.signal
             if (ongoingRequests[baseRoute]) {
-                ongoingRequests[baseRoute].abort("")
+                ongoingRequests[baseRoute].abort("Aborting previous unfinished api call.")
```
nafiz1001, line 365: It might be more helpful to include the API route that failed such as the `baseRoute`.

samouzegar10: thats a good idea :)

UlysseFG:
I think i like more the useDebounce version.

I do get this notification message though sometime when changing filter or changing sort fast :

action.error.message is undefined

./src/store.ts/notificationError/</<@http://localhost:9000/main.js?8199501887c1e93c9bb4:45001:11
./node_modules/redux-logger/dist/redux-logger.js/</S/</</<@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:215478:7667
./node_modules/redux-thunk/es/index.js/createThunkMiddleware/middleware/</<@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:216596:16
dispatch@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:217320:28
./src/utils/actions.ts/networkAction/</<@http://localhost:9000/main.js?8199501887c1e93c9bb4:45078:15

nafiz1001:
> I think i like more the useDebounce version.
> 
> I do get this notification message though sometime when changing filter or changing sort fast :
> 
> action.error.message is undefined
> 
> ./src/store.ts/notificationError/</<@http://localhost:9000/main.js?8199501887c1e93c9bb4:45001:11 ./node_modules/redux-logger/dist/redux-logger.js/</S/</</<@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:215478:7667 ./node_modules/redux-thunk/es/index.js/createThunkMiddleware/middleware/</<@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:216596:16 dispatch@http://localhost:9000/vendors-node_modules_ant-design_icons_es_icons_ArrowLeftOutlined_js-node_modules_ant-design_i-ed462e.js?8199501887c1e93c9bb4:217320:28 ./src/utils/actions.ts/networkAction/</<@http://localhost:9000/main.js?8199501887c1e93c9bb4:45078:15

I sometimes come across this sort of error. For me it's often because I called `dispatch` on something that's neither a redux thunk nor an object that describes a redux action.

nafiz1001:
I haven't checked the no `useDebounce` option yet.

nafiz1001:
> I haven't checked the no `useDebounce` option yet.

debounce is totally worth it even locally

PR #840, opened 2024-09-16
Fix Dataset Release Page

nafiz1001:


backend/fms_core/viewsets/dataset.py
```diff
@@ -58,42 +58,22 @@ def add_run_processing(self, request, *args, **kwargs):
...
-                included_readset.released_by = request.user
-                included_readset.save()
-
-            # set release flag of exceptions to the opposite flag
-            excluded_readsets = Readset.objects.filter(id__in=exceptions)
-            opposite_status = [ReleaseStatus.AVAILABLE, ReleaseStatus.BLOCKED, ReleaseStatus.RELEASED][release_status]
-            for excluded_readset in excluded_readsets:
-                excluded_readset.release_status = opposite_status
-                excluded_readset.release_status_timestamp = timezone.now()
-                excluded_readset.released_by = request.user
```
UlysseFG, line 90: This is just gone ...

nafiz1001: Is there a problem with that?

nafiz1001: It's right here ;)
```python
        releease_status_timestamp = timezone.now()
        for readset in readsets:
            readset.release_status = readset_updates[str(readset.id)]
            readset.release_status_timestamp = releease_status_timestamp
            readset.save()
```

backend/fms_core/viewsets/dataset.py
```diff
@@ -58,42 +58,22 @@ def add_run_processing(self, request, *args, **kwargs):
         else:
             return Response(self.get_serializer(datasets.values(), many=True).data)
 
-    @transaction.atomic
```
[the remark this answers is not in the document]
nafiz1001: I assumed it was unnecessary, but I can put it back

backend/fms_core/viewsets/dataset.py
UlysseFG: released_by is not longer set...

backend/fms_core/viewsets/dataset.py
```diff
@@ -62,38 +62,23 @@ def add_run_processing(self, request, *args, **kwargs):
...
-            # set release flag of exceptions to the opposite flag
-            excluded_readsets = Readset.objects.filter(id__in=exceptions)
-            opposite_status = [ReleaseStatus.AVAILABLE, ReleaseStatus.BLOCKED, ReleaseStatus.RELEASED][release_status]
-            for excluded_readset in excluded_readsets:
-                excluded_readset.release_status = opposite_status
-                excluded_readset.release_status_timestamp = timezone.now()
-                excluded_readset.released_by = request.user
-                excluded_readset.save()
-        except Exception as err:
+            releease_status_timestamp = timezone.now()
```
UlysseFG, line 72: can you fix the name my ocd is hurting ...

frontend/src/models/fms_api_models.ts
```diff
@@ -105,13 +105,18 @@ export interface FMSArchivedComment extends FMSTrackedModel {
...
+    BLOCKED = 2,
+}
 export interface FMSReadset extends FMSTrackedModel {
     id: FMSId                          // Unique ID of object in database
     name: string                       // External name that identifies the readset if the run did not come from Freezeman
     sample_name: string                // Name that identifies the sample if the run did not come from Freezeman
     derived_sample: FMSId              // Derived sample matching the readset
     sample_source: FMSId               // Last non pool sample (if any, else last pool) before experiment
-    release_status: number              // The file's release status (AVAILABLE = 0, RELEASED = 1,BLOCKED = 2)
+    release_status: ReleaseStatus      // The file's release status (AVAILABLE = 0, RELEASED = 1,BLOCKED = 2)
```
UlysseFG, line 119: i think the comment is not longer needed ;).

backend/fms_core/viewsets/dataset.py
nafiz1001: I'm surprised it let me release readsets despite not setting released_by.

frontend/src/models/fms_api_models.ts
nafiz1001: `validation_status` is next

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -1,105 +1,279 @@
...
+            const newReadsetReleaseStates = readsets.data.results.reduce<typeof readsetReleaseStates>((acc, readset) => {
+                acc[readset.id] = { old: readset.release_status, new: undefined }
+                return acc
+            }, {})
+            setReadsetReleaseStates(newReadsetReleaseStates)
+        })
+    }, [datasetID, dispatch])
+
+    const setAllReleaseStatus = useCallback((newReleaseStatus: ReleaseStatus.RELEASED | ReleaseStatus.BLOCKED) => {
+        setReadsetReleaseStates(produce((prev) => {
```
UlysseFG, line 207: Don't you need readsetReleaseStates as first param for produce ?

nafiz1001: That's just a variable name. It can be any name. But yeah I agree that `readsetReleaseStates` would be a better name.

UlysseFG: The question is not about the name it is just that i was under the impression the function produce took 2 params the first being the immutable state.

backend/fms_core/viewsets/dataset.py
nafiz1001: but i think it's there already?
```python
        if self.release_status_timestamp is not None and self.released_by is None:
            add_error("released_by", f"The user that changed the release_status need to be recorded.")
        if self.release_status_timestamp is None and self.released_by is not None:
            add_error("released_by", f"Readset release status was never set. Cannot set released_by.")
        if self.validation_status_timestamp is not None and self.validated_by is None:
            add_error("validated_by", f"The user that changed the validation_status need to be recorded.")
        if self.validation_status_timestamp is None and self.validated_by is not None:
            add_error("validated_by", f"Readset validation status was never set. Cannot set validated_by.")
```

backend/fms_core/viewsets/dataset.py
```diff
@@ -62,38 +62,24 @@ def add_run_processing(self, request, *args, **kwargs):
...
+                readset.release_status_timestamp = release_status_timestamp
+                readset.released_by = request.user
+                readset.save()
+        except Exception as e:
             transaction.set_rollback(True)
-            return HttpResponseServerError(err)
+            return HttpResponseServerError(f"Error updating release status: {e}")
 
-        return Response(self.get_serializer(Dataset.objects.get(pk=pk)).data)
+        return Response(status=204)
```
samouzegar10, line 82: you removed the return of the data that was updated? what is the reasoning?

backend/fms_core/viewsets/dataset.py
UlysseFG: I did not test it ... as far as i know this might never have worked ...

UlysseFG: Mmm i think you might be bypassing validation by using the manager save on the manager instead of the readset object... Try adding .all() to your queryset and removeing the released_by... it should return an error.

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -1,105 +1,279 @@
...
+const RELEASE_STATUS_STRING = {
+    [ReleaseStatus.RELEASED]: 'Released',
+    [ReleaseStatus.BLOCKED]: 'Blocked',
+    [ReleaseStatus.AVAILABLE]: 'Available',
+} as const
+const OPPOSITE_STATUS = {
+    [ReleaseStatus.AVAILABLE]: ReleaseStatus.RELEASED,
+    [ReleaseStatus.RELEASED]: ReleaseStatus.BLOCKED,
+    [ReleaseStatus.BLOCKED]: ReleaseStatus.RELEASED,
+} as const
```
samouzegar10, line 36: i dont understand the "as const" at the end of these const defined variables.

frontend/src/components/readsets/ReadsetsListContent.tsx
samouzegar10, line 1: ouff, im looking into this file and i understand what ulysse meant in terms of complexification of the code.

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -1,105 +1,279 @@
...
     // and has to flush the state.
-    const currentDatasetID = dataset ? +dataset.id : undefined
     useEffect(() => {
-        if (currentDatasetID) {
-            dispatch(ReadsetTableActions.resetPagedItems())
-            dispatch(ReadsetTableActions.setFixedFilter(createFixedFilter(FILTER_TYPE.INPUT_OBJECT_ID, 'dataset__id', String(currentDatasetID))))
-            dispatch(ReadsetTableActions.listPage(1))
+        dispatch(ReadsetTableActions.resetPagedItems())
+        dispatch(ReadsetTableActions.setFixedFilter(createFixedFilter(FILTER_TYPE.INPUT_OBJECT_ID, 'dataset__id', dataset.id.toString())))
+        dispatch(ReadsetTableActions.listPage(1))
```
samouzegar10, line 60: careful with this piece of code, i believe this function launches a second api call to fetch the current page list

backend/fms_core/viewsets/dataset.py
UlysseFG: The error messages are not caught in the front end (the operation fails silently in the console).

UlysseFG:
This seem to be working well. I am worried that there is no way even for an admin to reset the status to available. This mean mistakes by PMs will force curations. I would prefer this not being the case. It would be great for admins to have a button to reset all readset in a dataset to available as a way to prevent time consuming curations.

UlysseFG:
One last thing ... Maybe it might be nice to have a small note to mention the user need to release or block all readsets to save the change. Not sure if people will wonder why they can't save if they just click individual readsets status.

frontend/src/components/readsets/ReadsetsListContent.tsx
samouzegar10, line 1: Also, i believe i understand the complexity of the task itself, although from it looks like, you took a backend approach to a more or less complicated frontend issue

samouzegar10, line 1: i wrote some comments on my side, we can discuss them tomorrow if you want

backend/fms_core/viewsets/dataset.py
nafiz1001: I just felt it was pointless to respond with a dataset if the frontend is never using it anyway.
I should look into catching the error.

frontend/src/components/readsets/ReadsetsListContent.tsx
nafiz1001: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

frontend/src/components/readsets/ReadsetsListContent.tsx
nafiz1001: Readset release management is complex 🥹

backend/fms_core/viewsets/dataset.py
UlysseFG: Well you should at least tell that there was an error ... if you remove the line to set released_by it does fail; it just does not show anything on the frontend side.

samouzegar10:
i left some comments in [redmine](https://redmine.c3g-app.sd4h.ca/issues/2275?issue_count=15&issue_position=14&next_issue_id=2147&prev_issue_id=2282)

essentially my comments can be summed down to the following: 
- the solution to this complicated problem can be solvable with a single line.
- to keep in mind that any code overhaul can be taxing not at the current file level but at the app level, keeping the code readable at a glance and homogeneous to the rest of its codebase is a big plus.

backend/fms_core/viewsets/dataset.py
samouzegar10: This was taken from a best practices for restful api's blog post: 
Updates & creation should return a resource representation
A PUT, POST or PATCH call may make modifications to fields of the underlying resource that weren't part of the provided parameters (for example: created_at or updated_at timestamps). To prevent an API consumer from having to hit the API again for an updated representation, have the API return the updated (or created) representation as part of the response.

In case of a POST that resulted in a creation, use a [HTTP 201 status code](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5) and include a [Location header](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30) that points to the URL of the new resource. Both of those in should be in addition to including the newly created resource representation as the body of the response.

frontend/src/components/readsets/ReadsetsListContent.tsx
samouzegar10: yea, you ended up creating a readonly enum by typing it const and asserting it.  thats what i gathered, asserting makes sense but creating an enum also does

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -1,105 +1,345 @@
...
+                onConfirm={async () => {
+                    await releaseStatusManager.updateReleaseStatus()
+                    await refreshDataset()
+                }}
+                disabled={!saveChangesEnabled}
+                icon={null}
+            >
+                <Button
+                style={{ margin: 6 }}
+                type={"primary"}
```
samouzegar10, line 193: the save changes doesn't seem to be working, the app updates to the chosen values but it automatically returns to its original value before the update

nafiz1001: The lack of positive feedback might be missing. On my side, on another computer, it works.

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -1,105 +1,361 @@
...
+                disabled={!blockAllEnabled}>
+                Block All
+            </Button>
+            {isAdmin && <Button
+                    style={{ margin: 6 }}
+                    onClick={() => {
+                        releaseStatusManager.setAllReleaseStatus(ReleaseStatus.AVAILABLE)
+                    }}
+                    disabled={!undoAllReleaseAndBlockEnabled}>
+                    Undo All Release/Block
```
UlysseFG, line 169: This is confusingly similar to the other undo button. I would rename this to Reset Released Status and move it to the right of the save change button (maybe even use a different colour scheme). Also it might be nice to have it disabled unless there is some released_status that are different from Available ...

UlysseFG (approved):

PR #842, opened 2024-09-17
Measured volume optional

nafiz1001:


backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -44,7 +44,7 @@ def process_row_inner(self, sample_container, measures, process, process_measure
         if initial_volume is None:
```
UlysseFG, line 44: Initial_volume has to be optional too.

nafiz1001: like this?
```python
        if initial_volume is None:
            initial_volume = source_sample_obj.volume
```

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -29,14 +29,17 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
             if sample_obj.is_library:
                 self.errors['source_sample'] = f"Source sample can't be a library or a pool of libraries."
 
+            measured_volume = sample_information['measured_volume']
+            if measured_volume is None:
+                measured_volume = sample_obj.volume
```
UlysseFG, line 34: Don't put false values in here.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -44,7 +44,7 @@ def process_row_inner(self, sample_container, measures, process, process_measure
...
             self.errors['initial_volume'] = 'Initial volume must be specified'
         if measured_volume is None:
-            self.errors['measured_volume'] = 'Measured volume must be specified'
+            measured_volume = source_sample_obj.volume
```
UlysseFG, line 47: don't use false values.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
nafiz1001: Isn't that what the lab wanted?

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -29,14 +29,17 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
...
+
             # Update sample with sample_information
             new_volume = None
-            if all([sample_information['initial_volume'], sample_information['measured_volume'],
```
UlysseFG, line 34: Keep old condition and add a new one that covers omitted measured volume.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -29,14 +29,17 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
...
-                    process_measurement['volume_used']]):
-                delta_volume = sample_information['measured_volume'] - sample_information['initial_volume']
+            if all([sample_information['initial_volume'], process_measurement['volume_used']]):
+                delta_volume = measured_volume - sample_information['initial_volume']
                 new_volume = sample_obj.volume + delta_volume - process_measurement['volume_used']
             else:
-                self.errors['volume'] = 'Initial Volume, Measured Volume and Volume Used are required.'
+                self.errors['volume'] = 'Initial Volume and Volume Used are required.'
```
UlysseFG, line 42: Volume used is the only required now.

UlysseFG:
You can remove the red star from the current volume column ... When you save your template reposition the cursor to the top left so the user don't get a weird default position.

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
```diff
@@ -29,14 +29,21 @@ def process_row_inner(self, sample, sample_information, process_measurement, pro
             if sample_obj.is_library:
                 self.errors['source_sample'] = f"Source sample can't be a library or a pool of libraries."
 
+            initial_volume = sample_information['initial_volume']
+            measured_volume = sample_information['measured_volume']
+
+            if initial_volume is None:
+                initial_volume = sample_obj.volume
```
UlysseFG, line 36: We need to validate that if initial_volume is none, measured_volume is none too. Otherwise we introduce false information the volume at the time of the submission might not be the same as at the time of the measurement.

backend/fms_core/utils.py
```diff
@@ -190,18 +190,29 @@ def make_timestamped_filename(file_name: str) -> str:
     time.tzset()
     return f"{name}_{time.strftime('%Y-%m-%d_%H-%M-%S')}{extension}"
 
-def serialize_warnings(warnings):
+
+WarningType = NewType('WarningType', dict[str, tuple[str] | str | list[str] | list[tuple[str, list]]])
```
UlysseFG, line 194: Really ?

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -40,11 +40,12 @@ def process_row_inner(self, sample_container, measures, process, process_measure
         volume_used = process_measurement['volume_used']
         sample_volume = source_sample_obj.volume
         final_volume = None
-        
+
         if initial_volume is None:
```
UlysseFG, line 44: We need to validate that if initial_volume is none, measured_volume is none too. Otherwise we introduce false information the volume at the time of the submission might not be the same as at the time of the measurement.

backend/fms_core/utils.py
nafiz1001: Yes :)

backend/fms_core/template_importer/row_handlers/sample_qc/sample.py
nafiz1001: It's already done? unless `sample_obj.volume` can be None for some reason...

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
nafiz1001: It's already done? Unless `source_sample_obj.volume` can be None for some reason...

UlysseFG: Nope if initial_volume is None before you set it, the original measured_volume can still be anything ... You are falling into your own trap already. We need to distinguish the submitted values and the intention behind them from the values in the variables. I want to make sure that if current volume is not set on the template and a measured volume is set, we don't use an internal value that might be different from the internal value at the time of the operation in the lab (we send back an error requesting a current value). To do that i want to prevent users to submit an empty current volume with a not empty measured volume (so we get a valid delta).

nafiz1001: Ah oops. I missed the part where you said "if initial_volume is none, measured_volume is none too". Basically make sure either both is None, or only measured is None.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -40,24 +40,27 @@ def process_row_inner(self, sample_container, measures, process, process_measure
...
-            self.errors['initial_volume'] = 'Initial volume must be specified'
+            if measured_volume is not None:
+                self.errors['initial_volume'].append('Current Volume must be specified if Measured Volume is specified')
```
UlysseFG, line 46: initial_volume => current_volume as error key to match the column name in the template.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
```diff
@@ -40,24 +40,27 @@ def process_row_inner(self, sample_container, measures, process, process_measure
...
         if not any(self.errors.values()):
             if initial_volume < 0:
-                self.errors['initial_volume'] = f"Initial volume ({initial_volume}) must be a positive value."
+                self.errors['initial_volume'].append(f"Initial volume ({initial_volume}) must be a positive value.")
 
             if measured_volume < 0:
                 self.errors['measured_volume'] = f'Measured volume ({measured_volume}) must be a positive value.'
             elif measured_volume > initial_volume:
-                self.warnings['measured_volume'] = ("Measured volume {0} is greater than initial volume {1}", [measured_volume, initial_volume])
+                self.warnings['measured_volume'].append(("Measured volume {0} is greater than initial volume {1}", [measured_volume, initial_volume]))
```
UlysseFG, line 63: Include dot at the end of sentence.

backend/fms_core/template_importer/row_handlers/library_qc/library_qc.py
UlysseFG, line 46: Include dot at the end of sentence.

UlysseFG (approved):

PR #844, opened 2024-09-18
In Project study display, make clear removed samples completed their last step.

UlysseFG:
This is to address the worry that the sample did not complete the previous step when removing the sample from the study at a later step. It made it look like the sample were removed before the previous step which is not the case, This is mainly to alleviate that fear.


nafiz1001 (approved):
LGTM

PR #847, opened 2024-09-23
base migration file to add for rm2283

samouzegar10:


nafiz1001:
Unfortunately, I merged one of my PRs first, so you'll have to do a rebase on the migration script.

backend/fms_core/migrations/0065_v4_12_0.py
nafiz1001, line 1: I feel like this is missing something. I think the generic QC step should be removed and samples in existing steps should be moved to their corresponding QC.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+                                                                      step=step,
+                                                                      created_by_id=admin_user.id,
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        step = Step.objects.get(name="Sample QC")
+        dnaStep = Step.objects.get(name="Sample QC (DNA)")
+        if step:
+            step = Step.objects.get(name="Sample QC")
+            rnaStep = Step.objects.get(name="Sample QC (RNA)")
+            sns = SampleNextStep.objects.filter(step__id=step.id).update(step=rnaStep)
```
UlysseFG, line 57: Do not use update on the querysteps to save modifications to instances of models... Those do not get saved by revision. To make sure the changes get stored in revision you have to use the save() on an instance of the model (an object).

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+            for specification in step_info["specifications"]:
+                step_specification = StepSpecification.objects.create(name=specification["display_name"],
+                                                                      sheet_name=specification["sheet_name"],
+                                                                      column_name=specification["column_name"],
+                                                                      value=specification["value"],
+                                                                      step=step,
+                                                                      created_by_id=admin_user.id,
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        step = Step.objects.get(name="Sample QC")
```
UlysseFG, line 52: maybe use oldStep instead to make sure there is no confusion.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+            {"display_name": "Sample QC Type", "sheet_name": "SampleQC", "column_name": "Sample Kind", "value": "DNA"}]
+        },
+        {"name": "Sample QC (RNA)", "protocol_name": "Sample Quality Control","expected_sample_type": "EXTRACTED_SAMPLE", "specifications": [
+            {"display_name": "Sample QC Type", "sheet_name": "SampleQC", "column_name": "Sample Kind", "value": "RNA"}]
+        }
+    ]
+
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment(f"Create the basic initial workflows.")
```
UlysseFG, line 29: You should update the revision message.

UlysseFG:
> Unfortunately, I merged one of my PRs first, so you'll have to do a rebase on master for the migration script.

Make sure you rebase ...

UlysseFG:
The template you have created contains test information ... those should not be saved in the template ... Make sure also to remove the old version.

UlysseFG:
There should be a change to the templates.py to support your change to the template (New column to be added to the header list).

UlysseFG:
The test template should be modified with the new column too. The test template is in /backend/fms_core/tests/valid_templates/.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        step = Step.objects.get(name="Sample QC")
+        dnaStep = Step.objects.get(name="Sample QC (DNA)")
+        if step:
+            step = Step.objects.get(name="Sample QC")
+            rnaStep = Step.objects.get(name="Sample QC (RNA)")
+            sns = SampleNextStep.objects.filter(step__id=step.id).update(step=rnaStep)
+            so = StepOrder.objects.filter(step__id=step.id).update(step=rnaStep)
+            if sns > 0:
```
UlysseFG, line 59: Odd condition to trigger the deletion of the old step ... once all the changes are complete you should be able to delete without condition. If you want to verify something to make sure there was no issue use an assert so it fails if the condition is not met ... here it would just silently skip the deletion.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+                                                                      sheet_name=specification["sheet_name"],
+                                                                      column_name=specification["column_name"],
+                                                                      value=specification["value"],
+                                                                      step=step,
+                                                                      created_by_id=admin_user.id,
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        step = Step.objects.get(name="Sample QC")
+        dnaStep = Step.objects.get(name="Sample QC (DNA)")
+        if step:
```
UlysseFG, line 54: There is an issue with the logic in this section ... You only use the RNA step in SampleNextStep and StepOrder. Those step are condition to the previous extraction step used previously in the workflows. so for each instance of sample next step you need to check in the list of linked studies the workflow step for extraction and verify if it is DNA or RNA. If there is a conflict between the studies you would need to create a new sampleNextStep and SampleNextStepByStudy for that other kind (i don't think there is any in the system so far but we cannot assume. Same goes for StepOrder except that in that case you can't have conflict you have just one workflow by StepOrder. You just need to use the same kind for the step connected as the one for the extraction step.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -0,0 +1,66 @@
...
+            step = Step.objects.get(name="Sample QC")
```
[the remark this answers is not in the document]
samouzegar10: my fault

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
-import reversion
+from django.conf import settings
 from django.db import migrations
 from django.contrib.auth.models import User
+import reversion
 
-ADMIN_USERNAME = 'biobankadmin'
+from fms_core.models import StepOrder, SampleNextStep, Step, Protocol, StepSpecification, SampleNextStepByStudy, Workflow, Study
```
UlysseFG, line 6: Do not import models into the migration this can lead to circular import which the migration system really do not like ... You should use the dynamic model loading used in most migrations (exemple:  PropertyType = apps.get_model("fms_core", "PropertyType").

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                  newStep = Step.objects.get(name="Sample QC (RNA)")
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+                  newStep = Step.objects.get(name="Sample QC (DNA)")
+                sampleNextStep.update(step=newStep)
+                stepOrder.update(step=newStep)
+                sampleNextStep.save()
+                reversion.add_to_revision(sampleNextStep)
+                stepOrder.save()
+                reversion.add_to_revision(stepOrder)
+            if self.assertNotEquals(sns, 0):
```
UlysseFG, line 57: If oldStep has no SampleNextStep you won't delete it ?

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                step_specification = StepSpecification.objects.create(display_name=specification["display_name"],
+                                                                      sheet_name=specification["sheet_name"],
+                                                                      column_name=specification["column_name"],
+                                                                      value=specification["value"],
+                                                                      step=step,
+                                                                      created_by_id=admin_user.id,
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        oldStep = Step.objects.get(name="Sample QC")
+        if oldStep:
```
UlysseFG, line 42: don't test for this if it does not exist we want that migration to fail ... something is wrong.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+            sns = SampleNextStep.objects.filter(step__id=oldStep.id)
+            for sampleNextStep in sns:
+                sampleNextStepByStudy = SampleNextStepByStudy.objects.get(sample_next_step__id=sampleNextStep.id)
+                stepOrder = StepOrder.objects.get(id=sampleNextStepByStudy.step_order.id)
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+                  newStep = Step.objects.get(name="Sample QC (RNA)")
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+                  newStep = Step.objects.get(name="Sample QC (DNA)")
+                sampleNextStep.update(step=newStep)
```
UlysseFG, line 51: I recall issues with update and reversion. Make sure that once the migration is executed a new version of the sampleNextStep is stored for each update(). If you want to save on tests you can use an assignment to the attribute of the object and save the object instead.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                stepOrder.update(step=newStep)
+                sampleNextStep.save()
```
UlysseFG, line 53: you already updated that object.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                reversion.add_to_revision(sampleNextStep)
+                stepOrder.save()
```
UlysseFG, line 55: Already updated...

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+            sns = SampleNextStep.objects.filter(step__id=oldStep.id)
+            for sampleNextStep in sns:
+                sampleNextStepByStudy = SampleNextStepByStudy.objects.get(sample_next_step__id=sampleNextStep.id)
+                stepOrder = StepOrder.objects.get(id=sampleNextStepByStudy.step_order.id)
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+                  newStep = Step.objects.get(name="Sample QC (RNA)")
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+                  newStep = Step.objects.get(name="Sample QC (DNA)")
+                sampleNextStep.update(step=newStep)
+                stepOrder.update(step=newStep)
```
UlysseFG, line 52: StepOrder is part of the workflow definition it should not change according to what sample are going through the workflow ...

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                                                                      created_by_id=admin_user.id,
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        oldStep = Step.objects.get(name="Sample QC")
+        if oldStep:
+            sns = SampleNextStep.objects.filter(step__id=oldStep.id)
+            for sampleNextStep in sns:
+                sampleNextStepByStudy = SampleNextStepByStudy.objects.get(sample_next_step__id=sampleNextStep.id)
+                stepOrder = StepOrder.objects.get(id=sampleNextStepByStudy.step_order.id)
+                if sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
```
UlysseFG, line 47: I think your approach of testing for the sample kind for deciding on which step to put them on. You should for safety sake use an elif{} and an else {error} approach to prevent a typo from skipping a whole branch of the migration without us knowing.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -1,43 +1,64 @@
...
+                  newStep = Step.objects.get(name="Sample QC (RNA)")
```
UlysseFG, line 48: I would suggest to make this migration faster to store the 2 steps in different variables when you create them earlier then just reuse the same variables down here this is potentially hundreds of DB hits for nothing.

backend/fms_core/templates.py
```diff
@@ -639,6 +639,7 @@
   # prefill_info : [("Template Sheet Name", "Template Column Header", "Queryset Name", "Sample Model Attribute/Property", "Extractor Function"), ...]
   "prefill info": [
       ("SampleQC", "Sample Name", "name", "name", None),
+      ("SampleQC", "Sample Kind", "kind", "kind", None),
```
UlysseFG, line 642: You should not have to add Sample kind here if the specification works correctly. kind is not a sample attribute so this likely does not work. You can remove this line and validate that the prefill in the workflow still works.

backend/fms_core/tests/valid_templates/Sample_QC_v4_12_0 (1).xlsx:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
+ZoneId=3
+HostUrl=https://drive.usercontent.google.com/download?id=1BXzA-ejOOP7XlEwrTPQKTZUSXsdWDPOV&export=download&authuser=1&confirm=t&uuid=a7e66226-781c-4014-a977-e69b632a44a8&at=AN_67v0iUmiJNObIfz8R0SmeS-tT:1727465983705
```
UlysseFG, line 3: mmm ? Not sure what is happening here.

backend/fms_core/viewsets/_constants.py
```diff
@@ -153,6 +153,7 @@
 _run_type_filterset_fields: FiltersetFields = {
     "id": PK_FILTERS,
     "name": CATEGORICAL_FILTERS_LOOSE,
+    **_prefix_keys("platform__", _platform_filterset_fields),
```
UlysseFG, line 156: i don't think this belong in this PR...

UlysseFG:
Looks like you merged something wrong in your branch. There is changes about the ExperimentRun list and api.ts that should not be part of this PR.

backend/fms_core/migrations/0065_v4_12_0.py
samouzegar10: no, the idea i had was to verify the count of sns in the db left but i made a mistake, i updated the file with a different condition but similar logic

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
+from django.conf import settings
```
UlysseFG, line 1: Garde cette migration dans l'autre fichier autant que possible ... les migrations ne sont pas atomiques entre elles ils vaut mieux ne pas trop séparer les migrations d'une version. Si tu avais à créer un second fichier de migration, le nom aurait le même numéro de version (c'est pour 4.12.0) mais le compteur augmente de 1.

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+            updatedSampleNextStep = SampleNextStep.objects.get(id=sampleNextStep.id)
+            sampleNextStepByStudy = SampleNextStepByStudy.objects.get(sample_next_step__id=sampleNextStep.id)
+            stepOrder = StepOrder.objects.get(id=sampleNextStepByStudy.step_order.id)
+            try:
+              if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+                updatedSampleNextStep.step = dnaStep
+                stepOrder.step = dnaStep
+              elif sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+                updatedSampleNextStep.step = rnaStep
+                stepOrder.step = rnaStep
```
UlysseFG, line 68: remove that from here if you are testing for step__id=oldStep.id later.

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+            except Exception:
+              errors["AssigningError"] = f"There was an error while assigning the new steps to the existing sample next steps objects {sampleNextStep}."
+            if errors:
+                raise ValidationError(errors)
+            stepOrder.save()
```
UlysseFG, line 73: remove ...

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+            stepOrder = StepOrder.objects.get(id=sampleNextStepByStudy.step_order.id)
+            try:
+              if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+                updatedSampleNextStep.step = dnaStep
+                stepOrder.step = dnaStep
+              elif sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+                updatedSampleNextStep.step = rnaStep
+                stepOrder.step = rnaStep
+            except Exception:
+              errors["AssigningError"] = f"There was an error while assigning the new steps to the existing sample next steps objects {sampleNextStep}."
```
UlysseFG, line 70: don't bother with errors... if migration fails during release we cancel release and we come back to the drawing board ...

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+            if errors:
+                raise ValidationError(errors)
+            stepOrder.save()
+            updatedSampleNextStep.save()
+            reversion.add_to_revision(updatedSampleNextStep)
+            reversion.add_to_revision(stepOrder)
```
UlysseFG, line 76: remove.

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+        # make sure there are no loose ends before deleting oldStep
+        so = StepOrder.objects.filter(step__id=oldStep.id)
+        for order in so:
+            stepOrder = StepOrder.objects.get(id=order.id)
+            if "DNA" in order.previous_step_order.all().first().step.name:
```
UlysseFG, line 81: this is risky. 3 letters could be part of normal words (BANDNAME !)... this likely work but it would be much safer to use the full step name.

backend/fms_core/migrations/0066_v4_12_1.py
nafiz1001: suddenly french

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+                stepOrder.step = dnaStep
+            elif "RNA" in order.previous_step_order.all().first().step.name:
+                stepOrder.step = rnaStep
+            stepOrder.save()
+        so = StepOrder.objects.filter(step__id=oldStep.id)
+        if len(so) == 0:
```
UlysseFG, line 87: Same issue here ... you skip the deletion but this should be an error (there is no reason for this to happen) so either don't test (if you have an else clause in the previous loop you don't need this) or include an else clause that fails so the migration gets rolled back.

backend/fms_core/migrations/0066_v4_12_1.py
```diff
@@ -0,0 +1,95 @@
...
+          oldStep.delete()
+          reversion.add_to_revision(oldStep)
```
UlysseFG, line 89: I think delete() include code to add the object to a revision (you might want to validate).

UlysseFG:
delete the old test template : backend/fms_core/tests/valid_templates/Sample_QC_v4_4_0.xlsx
if it is not old it should have the same version as the reference template.

UlysseFG:
Reference templates (backend/fms_core/static/submission_templates/Sample_QC_v4_12_0.xlsx) should not contain data ...

samouzegar10:
> Looks like you merged something wrong in your branch. There is changes about the ExperimentRun list and api.ts that should not be part of this PR.

this has been fixed

samouzegar10:
> Looks like you merged something wrong in your branch. There is changes about the ExperimentRun list and api.ts that should not be part of this PR.

this has been fixed

UlysseFG:
My comments about the templates are still valid and unresolved... There should not be data in the reference template and your test template should have the same version as the reference template. Old templates (test or reference) that are no longer valid are to be removed.

samouzegar10:
hello, i tested both dna and rna lab work step progressions. the tests we're all positively conclusive.

UlysseFG:
Is there a reason to delete the Sample pooling test template ?

UlysseFG:
Multiple (automated) tests are failing ... There is still some more work to be done here.
=> coverage run ./manage.py test -v 2

UlysseFG:
The sample QC template dropdown are broken.

samouzegar10:
> The sample QC template dropdown are broken.

which columns are broken, i have the template in front of me and the columns are working.


QC Flag, QC Quality Instruments, QC Quantity Instruments, Workflow Action and NA Quantity (ng) has the formula on all its corresponding rows.

could you clarify what you mean?

samouzegar10:
this pr is ready to be merged, this has been tested thoroughly and extensively. im confident on this since i did the entire testing loop 3 times. if i have the green light i will merge this pr

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,73 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
+    # load the new sample qc steps and other necessary assets
+    STEPS = [
+        {"name": "Sample QC (DNA)", "protocol_name": "Sample Quality Control","expected_sample_type": "EXTRACTED_SAMPLE",
+          "specifications": [
+            {"display_name": "SampleQcType", "sheet_name": "SampleQC", "column_name": "Sample Kind", "value": "DNA"}]
+        },
+        {"name": "Sample QC (RNA)", "protocol_name": "Sample Quality Control","expected_sample_type": "EXTRACTED_SAMPLE",
+          "specifications": [
+            {"display_name": "SampleQcType", "sheet_name": "SampleQC", "column_name": "Sample Kind", "value": "RNA"}]
+        }
```
nafiz1001, line 41: Optional: would be nice to follow the format:
```python
{
    "name": ...,
    "protocol_name": ...,
    ...
}
```

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,73 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
+        for sampleNextStep in sns:
+            updatedSampleNextStep = SampleNextStep.objects.get(id=sampleNextStep.id)
+            if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+              updatedSampleNextStep.step = dnaStep
+            elif sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+              updatedSampleNextStep.step = rnaStep
+            updatedSampleNextStep.save()
+            reversion.add_to_revision(updatedSampleNextStep)
+        # make sure there are no loose ends before deleting oldStep
+        so = StepOrder.objects.filter(step__id=oldStep.id)
```
nafiz1001, line 87: Optional: same applies here

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,73 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
+                                                                      updated_by_id=admin_user.id)
+                reversion.add_to_revision(step_specification)
+        oldStep = Step.objects.get(name="Sample QC")
+        dnaStep = Step.objects.get(name="Sample QC (DNA)")
+        rnaStep = Step.objects.get(name="Sample QC (RNA)")
+        # Start for data transfer process
+        sns = SampleNextStep.objects.filter(step__id=oldStep.id)
+        for sampleNextStep in sns:
+            updatedSampleNextStep = SampleNextStep.objects.get(id=sampleNextStep.id)
+            if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
```
nafiz1001, line 80: Optional: I'm personally biased towards obtaining DNA `SampleKind` directly and doing something like
```python
sampleNextStep.sample.derived_samples.first().sample_kind.id == dna_sample_kind.id
```
It feels more robust compared to string-based comparison.

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,73 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
+              updatedSampleNextStep.step = rnaStep
+            updatedSampleNextStep.save()
+            reversion.add_to_revision(updatedSampleNextStep)
+        # make sure there are no loose ends before deleting oldStep
+        so = StepOrder.objects.filter(step__id=oldStep.id)
+        for order in so:
+            if "Extraction (DNA)" in order.previous_step_order.first().step.name:
+                order.step = dnaStep
+            elif "Extraction (RNA)" in order.previous_step_order.first().step.name:
+                order.step = rnaStep
```
nafiz1001, line 92: Optional: similarly, I'm biased towards obtaining a `Step` model directly and comparing by id rather than name

nafiz1001: In fact, using `in` is more risky in general. It's preferable to use `==` or maybe even compare by step id if possible.

backend/fms_core/migrations/0065_v4_12_0.py
nafiz1001, line 1: I think you're supposed to also create a `StepOrder` to link steps to workflows. But I could be wrong...

nafiz1001: You might even need to create new workflows 😱 . Again, I could be wrong...

nafiz1001: False Alarm.

UlysseFG:
> > The sample QC template dropdown are broken.
> 
> which columns are broken, i have the template in front of me and the columns with dropdown in their cells are working.
> 
> QC Flag, QC Quality Instruments, QC Quantity Instruments, Workflow Action and NA Quantity (ng) has the formula on all its corresponding rows.
> 
> could you clarify which columns?

Well on my side all columns are broken ... there is no longer any dropdown working. It might have to do with saving the files with a version of Excel that do not support the LibreOffice format... We had that issue with Chris ... i will make the change to the template on my side and push it to your branch if you don't mind ...

samouzegar10:
> > > The sample QC template dropdown are broken.
> > 
> > 
> > which columns are broken, i have the template in front of me and the columns with dropdown in their cells are working.
> > QC Flag, QC Quality Instruments, QC Quantity Instruments, Workflow Action and NA Quantity (ng) has the formula on all its corresponding rows.
> > could you clarify which columns?
> 
> Well on my side all columns are broken ... there is no longer any dropdown working. It might have to do with saving the files with a version of Excel that do not support the LibreOffice format... We had that issue with Chris ... i will make the change to the template on my side and push it to your branch if you don't mind ...

hopefully excel will be a thing of the past sooner than later, i have no quarrels you doing it on your end.

backend/fms_core/migrations/0065_v4_12_0.py
samouzegar10: well the distinction here is important, the string comparison is not merely there for fun, we need to know what type sequencing the previous step was doing and comparing by id does not give us this answer

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,73 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
+        dnaStep = Step.objects.get(name="Sample QC (DNA)")
+        rnaStep = Step.objects.get(name="Sample QC (RNA)")
+        # Start for data transfer process
+        sns = SampleNextStep.objects.filter(step__id=oldStep.id)
+        for sampleNextStep in sns:
+            updatedSampleNextStep = SampleNextStep.objects.get(id=sampleNextStep.id)
+            if sampleNextStep.sample.derived_samples.first().sample_kind.name == "DNA":
+              updatedSampleNextStep.step = dnaStep
+            elif sampleNextStep.sample.derived_samples.first().sample_kind.name == "RNA":
+              updatedSampleNextStep.step = rnaStep
```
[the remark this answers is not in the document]
samouzegar10: in this instance, its important to know why the code has been implemented this way. we want to know what type of sequencing will occur with the previous step, dna or rna

nafiz1001 (approved):

PR #848, opened 2024-09-24
Remove concentration column for sample extraction

nafiz1001:


backend/fms_core/services/sample.py
```diff
@@ -359,8 +359,7 @@ def extract_sample(process: Process,
             if concentration_destination is not None:
                 sample_destination_data["concentration"] = concentration_destination
             elif sample_kind_destination.concentration_required:
-                # Force extracted sample to have a concentration if sample_kind is DNA.
-                errors.append(f"Concentration must be specified for an extracted sample if the sample_kind is DNA.")
+                sample_destination_data["concentration"] = 1
```
nafiz1001, line 362: Most likely wrong.

nafiz1001:
> ... but first we need to check on the model if that concentration_required attribute still make sense ... We could validate with the lab if they still need DNA samples to have a concentration at submission. Possibly at the level of sample submission but maybe we can remove that part from the extraction. The sample_kind .concentration_required may not make sense anymore ... it looks like this is not tied to the sample_kind anymore.

backend/fms_core/services/sample.py
UlysseFG: I think we'll get rid of the concentration_required attribute on sample_kind ... it was just for DNA and with this change it no longer hold the road. So remove the field on the sample_kind model and remove all tests that use that field. As far as i can see extraction is the only place it's used (a few easy test tweaks too).

UlysseFG: ... and yeah this is wrong...

UlysseFG:
When you change the header signature of a template you need to update the version of the file to the current version.

UlysseFG:
The change to the templates.py file header is missing.

UlysseFG:
You need to update the test template too.

samouzegar10:
since django is an ORM shouldnt there be a migration file to remove the now deprecated variable (concentration_required)?

nafiz1001:
> since django is an ORM shouldnt there be a migration file to remove the now deprecated variable (concentration_required) in the db model as well?

riiiiiiiiiiiiight

backend/fms_core/migrations/0067_v4_12_0.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.4 on 2024-10-16 17:30
```
UlysseFG, line 1: don't make a new migration file use the one that is not for index...

backend/fms_core/templates.py
```diff
@@ -658,7 +658,7 @@
           'headers': ['Extraction Type', 'Current Volume (uL)', 'Volume Used (uL)', 'Source Sample Name', 'Source Container Barcode', 'Source Container Coord',
```
UlysseFG, line 658: The Sample extraction template file name should be changed up here in file.

UlysseFG:
Update the templates version in the names (both the active template as well as updated test templates...)

nafiz1001:
> Update the templates version in the names (both the active template as well as updated test templates...)

i was hoping we could leave the test templates as they are 😩

UlysseFG (approved):

PR #850, opened 2024-09-26
Prevent crash when exporting sample without project (dev).

UlysseFG:


nafiz1001 (approved):
My eyeballs agree with the code

PR #851, opened 2024-09-27
fix PagedItemsTable

nafiz1001:


nafiz1001:
I'll let @sebastianballesteros do the fix.

PR #856, opened 2024-10-01
Add Marie-Michelle Simons to the list of lab tech

nafiz1001:
Add lab technician name to the list on Library Prep template


UlysseFG:
The new name does not appear in the dropdown. To have the formula automatically adjust you need to insert a cell within the boundary otherwise you need to manually adjust the cells range in the columns.

UlysseFG:
It would be better if the file would open on the actual template sheets instead of the index sheet. Just saying...

UlysseFG (approved):

PR #859, opened 2024-10-03
made the bottom of the table stick to the screen on all tables

samouzegar10:


nafiz1001:
It works surprisingly well with multiple tables because we're only showing one table at a time anyway.
But can you make the headers sticky too?

UlysseFG:
The study workflow tables on project page is wonky at the moment (the floating pagination sometime show sometime not sometime once only for multiple tables open ...).

UlysseFG:
I am not sure i like the look of the floating pagination... it hangs to the left of the page and has insufficient padding to the left. It also does not have the same size as the table which is strange ...

samouzegar10:
closing pr because antd upgrade is turning out to be a viable solution

PR #860, opened 2024-10-07
 Template Experiment Submission modification

nafiz1001:


UlysseFG:
- You need to make a migration to change the instruments names to match the ones in the template.
- You column you moved you have to insert them between the 2 column named not overwrite the other columns ... there is missing columns.
- You need to change the templates.py header definition if you change the order of the columns.

UlysseFG:
I really hate the instrument name change ...
I feel we should in fact strip all instruments serial ID from their name (while keeping them for the template dropdown) and modify the importer to strip it from the input ...
That way we can personalize the display to our taste in the app (adding the serial between parenthesis or adding the type or as a tooltip).
If you want to keep going with the current renaming, i suggest you apply the same rule to all instruments not just the one in the experiment run (that might lead to other breaks double check with the Protected and OnNetwork for hardcoded code). We'll likely have to keep fixing names forever ...

backend/fms_core/migrations/0065_v4_12_0.py
```diff
@@ -28,6 +28,23 @@ def set_measured_volume_properties_optional(apps, schema_editor):
...
             reversion.add_to_revision(property_type)
 
+def remove_serial_number_from_some_instruments(apps, schema_editor):
+    instrument_name_maps = {
+        "Rosalind Franklin": "Decomissioned-Rosalind Franklin",
+        "LH00375-Rosalind Franklin": "Rosalind Franklin",
+    }
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment("Prepend serial number to instrument names")
```
UlysseFG, line 38: Comment does not match job.

backend/fms_core/utils.py
```diff
@@ -73,6 +73,13 @@ def normalize_scientific_name(name: str) -> str:
     """
     return " ".join((a.title() if i == 0 else a.lower()) for i, a in enumerate(RE_WHITESPACE.split(name or "")))
 
+def str_remove_parenthesis(s: str | None) -> str | None:
```
UlysseFG, line 76: This is not a generic function you should put that function in the importer file where it is used ... i doubt we'll reuse this ...

UlysseFG:
Not quite certain why but the frontend instrument list no longer display instruments.

UlysseFG:
The template have different column order now ... I think it would be a good idea to have a new version for the templates (renaming the files and references to the files.)

UlysseFG (approved):
Should be good once rebased

PR #862, opened 2024-10-15
Fix Sample fetching in Labwork Step Overview Panel

nafiz1001:


samouzegar10:
i dont have much to say about this pr, you added a return and modified the main component. looks good to me

UlysseFG:
I still get forever loading tables... When loading a group with a lot of sample and then right away switching to another group i get forever loading second group. This seem to be related to the abort (i get an abort error in the console).

nafiz1001:
> I still get forever loading tables... When loading a group with a lot of sample and then right away switching to another group i get forever loading second group. This seem to be related to the abort (i get an abort error in the console).

(╯°□°)╯︵ ┻━┻

UlysseFG:
ᑐᙵᓱᒋᑦ

UlysseFG (approved):

PR #865, opened 2024-10-17
removed unused prefiller options on instrument type for the sample qc…

samouzegar10:
… dna/rna step


backend/fms_core/migrations/0067_v4_12_0.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 4.2.4 on 2024-10-17 17:30
```
UlysseFG, line 1: Do not create a new migration file ... use the existing one that is not for index...

backend/fms_core/models/instrument_type.py
```diff
@@ -24,7 +24,8 @@ class InstrumentType(TrackedModel):
     index_read_3_prime = models.CharField(choices=INDEX_READ_DIRECTIONS_CHOICES,
                                           max_length=10,
                                           help_text="Instrument specific read direction for the index part at the 3 prime end of the sequence.")
-
+    is_in_prefiller = models.BooleanField(default=True,
```
UlysseFG, line 27: Don't change the model ...

frontend/src/components/PrefillTemplateColumns.tsx
```diff
@@ -46,12 +47,13 @@ interface SelectInstrumentTypeProps extends SelectProps {
...
 function SelectInstrumentType({ type, ...props }: SelectInstrumentTypeProps) {
     const listInstrumentTypesCallback = useCallback(() => store.dispatch(api.instrumentTypes.list({ platform__name: type })), [type])
-    
+
     const [instrumentTypes, setInstrumentTypes] = useState<InstrumentType[]>([])
 
     useEffect(() => {
         listInstrumentTypesCallback().then((response) => {
-            setInstrumentTypes(response.data.results)
+            let results = response.data.results.filter((instrument: FMSInstrumentType) => instrument.is_in_prefiller)
```
UlysseFG, line 55: Add filtering code here. You need to check on the step make sure this is the sample QC step... then get only the right stuff though.

backend/fms_core/migrations/0067_v4_12_0.py
samouzegar10: removed

backend/fms_core/templates.py
```diff
@@ -631,7 +632,7 @@
   "user prefill info": {
       "QC Date (YYYY-MM-DD)": "date",
       "Volume Used (uL)": "number",
-      "Quality Instrument": "qc-instrument",
+      "Quality Instrument": SAMPLE_QC_QUALITY_INSTRUMENTS,
       "Quality Flag": VALID_QC_FLAG_CHOICES,
       "Quantity Instrument": "qc-instrument",
```
UlysseFG, line 637: We should replace that list too. I think we will just get this request later otherwise (using the sample QC template quantity instrument list as a reference).

backend/fms_core/template_importer/_constants.py
```diff
@@ -16,6 +16,7 @@
 VALID_QC_FLAG_CHOICES = [PASSED, FAILED]
 # QC Instruments
 LIBRARY_QC_QUALITY_INSTRUMENTS = ["Caliper LabChip", "TapeStation"]
+SAMPLE_QC_QUALITY_INSTRUMENTS = ["Agarose Gel", "BioAnalyzer", "Caliper LabChip", "NanoDrop", "TapeStation", "Tecan Absorbance"]
```
UlysseFG, line 19: I think the order in the template was from most used to least used ... We might want to keep the order from the template. The fact that she mention TapeStation as missing while it is not (merely need to scroll) make me think the order is important to them.

nafiz1001: I think we should teach the lab the right way of looking for instruments. Going by importance is going to be annoying when we have different labs using freezeman.

samouzegar10: obviously, this task in itself was pretty menial, we have to be careful to not customize too heavily on what the users are suggesting (as ulysse mentioned, the solution was to simply to scroll down). this type of "task" seems to be a recurrent one. im on board with nafiz but i dont think this is where we should make that kind of decision.

im going to keep the order on the template but i would like to remark that i agree with nafiz but this would open up an entire discussion that we could address when gathered at the office.

backend/fms_core/templates.py
```diff
@@ -631,9 +632,9 @@
   "user prefill info": {
       "QC Date (YYYY-MM-DD)": "date",
       "Volume Used (uL)": "number",
-      "Quality Instrument": "qc-instrument",
+      "Quality Instrument": SAMPLE_QC_QUALITY_INSTRUMENTS,
       "Quality Flag": VALID_QC_FLAG_CHOICES,
-      "Quantity Instrument": "qc-instrument",
+      "Quantity Instrument": SAMPLE_QC_QUALITY_INSTRUMENTS,
```
UlysseFG, line 637: The list for quantity and quality is not the same. Please use the Sample QC template as reference.

nafiz1001:
Part of me feels that a long term solution probably would have been to update the instrument models to narrow down the type of QC instruments. But I'd let Ulysse decide on that.

nafiz1001 (approved):

UlysseFG (approved):

UlysseFG:
> Part of me feels that a long term solution probably would have been to update the instrument models to narrow down the type of QC instruments. This way we don't need to update the code to include more sample QC instruments. But I'd let Ulysse decide on that.

I do agree partially but i spoke of it with Sebastian and personally i think the clean solution is too invested into something that is not much use. Once a clearer and broader use case emerge we could change the list for a better definition.

PR #866, opened 2024-10-17
Sample submission template changes

UlysseFG:
- [x] Replace Taxon ncbi ID by Taxon name.
- [x] Create parent container when doing sample submission when parent container fully defined (2 additional columns).
- [x] Remove Father, Mother and Pedigree columns.
- [x] Change the column order.
- [x] Update tests and test templates.


nafiz1001:
Maybe we should warn users if new containers are getting created, and new generic individual?

nafiz1001 (approved):

PR #867, opened 2024-10-22
Python-Generated Samplesheet

nafiz1001:


nafiz1001:
I'm actually not even sure if I'm done 🤷🏼‍♂️

backend/fms_core/services/samplesheet.py
```diff
@@ -60,22 +56,224 @@ def get_samplesheet(container_kind, placement):
...
-                    sheet.cell(row=SAMPLESHEET_FIRST_INPUT_ROW + i, column=COLUMN_INDEX_5PRIME).value = row_data[COLUMN_INDEX_5PRIME]
-                    i = i + 1
         else:
             errors.append(f"Container of kind {container_kind} cannot be used for an experiment run.")
     except Exception as err:
         errors.append(err)
 
+    row_data_by_lane.sort(key=lambda x: x[0])
+    bclconvert_data = [
+        BCLConvert_Datum(Lane=lane, Sample_ID=sample_alias, Index=sequences_3prime, Index2=sequences_5prime)
```
UlysseFG, line 69: how about using "*" as a separator and have Sample_ID be sample_alias + "*" + derived_sample_id ? Both here and for germline. The idea is to make the sample_ID unique while providing an easy separator for downstream processing to remove the part they don't want ...

backend/fms_core/services/samplesheet.py
```diff
@@ -54,28 +50,230 @@ def get_samplesheet(container_kind, placement):
...
+    Lane: str
+    Sample_ID: str
+    Index: str
+    Index2: str
+
+@dataclass
+class DragenGermline_Datum:
+    Sample_ID: str
+
+def generate_samplesheet_workbook(BCLConvert_Data: list[BCLConvert_Datum], DragenGermline_Data: list[DragenGermline_Datum]) -> Workbook:
```
UlysseFG, line 97: maybe use a `_` before your function name to indicate it's a local function.

UlysseFG:
Since you are no longer using the samplesheet template, you should remove it.

backend/fms_core/services/workbook_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+@dataclass
+class CellDescription:
+    value: str
+    validation: DataValidation | None
+    apply_cell: Callable[[Cell], None] | None
+
+CD = CellDescription
+
+def insert_cells(worksheet: Worksheet, topleft: tuple[int, int], descriptors: Iterable[Iterable[CellDescription]], order: Literal["row"] | Literal["col"]):
```
UlysseFG, line 15: is topleft the best name you can find for this ?

nafiz1001: `initial_cell_location`?

UlysseFG:
The sample ID is a bit weird but we'll adjust if there is an issue with the machine or with the run processing.

UlysseFG (approved):

PR #868, opened 2024-10-22
Sample pooling v4_12_0

nafiz1001:


UlysseFG:
Missing :
- Add "Illumina iSeq100" on the list column I

UlysseFG:
Info tab fields order is no longer correct.

nafiz1001:
> Missing :
> 
>     * Add "Illumina iSeq100" on the list column I

> Info tab fields order is no longer correct. You do not need to correct the test templates, just the template itself.

Should I make the same updates to the test templates?

nafiz1001:
I'll let Andres review the template changes

PR #871, opened 2024-10-28
Frontend reduce warnings

nafiz1001:


samouzegar10:
these improvements are nice, there might be a few things where i would maybe have an opinion on but overall, it looks good. good in my book

frontend/src/utils/menus.js
```diff
@@ -1,45 +1,53 @@
 import React from "react";
 import { Link } from "react-router-dom";
-import { Menu } from "antd";
 
-// renderMenuItem and matchingMenuKeys taken from chord_web
+/**
+ * @typedef {Required<import("antd").MenuProps>['items'][number]} MenuItem
+ * @typedef {Partial<{ key: string, icon: JSX.Element, text: string, url: string, onClick: () => void, children: BadMenuItem[], style: React.CSSProperties, disabled?: boolean }>} BadMenuItem
+ */
```
samouzegar10, line 7: this makes sense to me although i would suggest going for readability instead of overengineering, which seems to be a common outcome when you work on the frontend components. also formatting in vs code, it would be nice to not have the type definition in 1 line but in multiple lines since there are a lot of variables.

frontend/src/utils/menus.js
```diff
@@ -1,45 +1,53 @@
...
+	for (const i of menuItems) {
+		if (i.url && window.location.pathname.startsWith(i.url)) {
+			keys.push(i.key || i.url || "")
+		}
+		if (i.children) {
+			keys.push(...matchingMenuKeys(i.children))
+		}
+	}
+	return keys
+}
```
samouzegar10, line 53: i know this file was already using the window.location since its js but this could be converted to ts (which is something you are usually keeping an eye out for).

if you convert the file, you can use the following package to get the current url values required and, as you may already know, they can be deconstructed as follow:

`import { Route, Link, Routes, useLocation } from 'react-router-dom';
const location = useLocation();
const { hash, pathname, search } = location;
`

frontend/src/utils/menus.js
nafiz1001: I wasn't sure if multiple lines were possible in JSDoc. Turns out yes.

frontend/src/utils/menus.js
nafiz1001: I'm hesitant to make any big changes. The name `useLocation` implies that it's a hook, so `matchingMenuKeys` would need to be converted to a hook which leads to more code changes.

To make matters worse, there's recursion. Not sure if hooks behave well with recursion.

Maybe it can be avoided. But again, I'm trying to reduce changes and surprises while removing warnings.

samouzegar10: no worries at all, these are simply observations and not mandatory changes, in my experience pr comments are mostly suggestion unless there are some major breaking changes that would require, usually common sense is to be applied instead of the usual pragmatic approach on these smaller issues

PR #872, opened 2024-10-28
Last minute changes to sample submission template.

UlysseFG:
- [x] Library and Library in pool have only DNA and RNA in dropdown.
- [x] Coordinates are either by Column or Row order with an option cell.
- [x] Container name is the mandatory field and barcode is a formula/copy if left empty.


nafiz1001:
Row 10 (the first sample row) is missing the formula `=IF(M10="","",M10)` for barcode.

nafiz1001:
- [x] Library and Library in pool have only DNA and RNA in dropdown.
- [x] Coordinates are either by Column or Row order with an option cell.
- [ ] Container name is the mandatory field and barcode is a formula/copy if left empty.
  - Row 10 (the first sample row) is missing the formula =IF(M10="","",M10) for barcode.

nafiz1001 (approved):

PR #876, opened 2024-10-29
check Sample_ID length in service

nafiz1001:


backend/fms_core/services/samplesheet.py
```diff
@@ -78,7 +78,10 @@ def get_samplesheet(container_kind, placement):
     return out_stream.getvalue(), errors, warnings
 
 def _get_Sample_ID(sample_alias, derived_sample_id):
-    return f"{sample_alias}_{derived_sample_id}"
+    Sample_ID = f"{sample_alias}_{derived_sample_id}"
+    if len(Sample_ID) > 100:
+        raise Exception(f"Sample_ID '{Sample_ID}' is too long. It must be less than or equal to 100 characters.")
```
UlysseFG, line 83: Not sure it is the best way to deal with that ... i would rather cut-off the sample_alias part short of the extra length... it would be unique but the sample_name would be messy but if someone use 150+ characters for their sample name they deserve that ;)

UlysseFG: I mention that because if you let it as an exception the lab will have to create the samplesheet by themselves and cut the sample_ids manually ... this is worse than having sample longer than 100. I expect the solution to the issue is to truncate the name while keeping the derived_sample_id part for uniqueness.

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,19 @@
+def fit_string_with_ellipses(string: str, max_length: int, ellipses = "...") -> str:
```
UlysseFG, line 1: i would suggest ellipse instead of ellipses in function name and param (it sound like a list of ellipse while there is a single one). Max_length leave the impression there is a variability. I would use fitting_length or forced_length.

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    if max_length < 0:
+        raise Exception(f"The max_length '{max_length}' must be a positive integer")
```
UlysseFG, line 3: Use `.` at the end of error messages and warnings.

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    if max_length < len(ellipses):
```
UlysseFG, line 4: i think we are using the french term ... I would still use the singular across the code in any case.
ellipsis [noun] the deliberate omission of a word or words from a sentence that do not prevent the meaning from being understood.

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,19 @@
...
+        raise Exception(f"The max_length '{max_length}' is too short to fit the ellipses '{ellipses}'")
+
+    if len(string) > max_length:
+        if len(string) < len(ellipses):
+            raise Exception(f"The string '{string}' is too short to fit the ellipses '{ellipses}'")
```
UlysseFG, line 9: Use . at the end of error messages and warnings.

UlysseFG, line 9: this case is unreachable since string > max_length and max_length > ellipsis => string > ellipsis.

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+        left, right = string[:len(string) // 2], string[-len(string) // 2:]
+        while len(left) + len(ellipses) + len(right) > max_length:
```
UlysseFG, line 12: I doubt this is the most efficient way of doing this ... This is semi efficient for length 100 but its very lacking for 100000.
What about `left = string[:math.floor(max_length - len(ellipsis), 2)] right = string[:math.ceil(max_length - len(ellipsis), 2)]`

backend/fms_core/services/utils.py
```diff
@@ -0,0 +1,22 @@
+import math
+
+def fit_string_with_ellipsis_in_middle(string: str, max_length: int, ellipsis = "...") -> str:
```
UlysseFG, line 3: Sorry about that but i would strongly advise to create a couple of tests (at least one without removal and one with) for this service as well as leave a little docstring comment for documentation sake.

nafiz1001: I forgive you.

backend/fms_core/tests/test_utils.py
```diff
@@ -57,3 +59,34 @@ def test_str_cast_and_normalize(self):
     def test_get_normalized_str(self):
         self.assertEqual(get_normalized_str({"test": 5}, "test"), "5")
         self.assertEqual(get_normalized_str({"test": None}, "test", default="aa"), "aa")
+
+FIT_STRING = "Hello World!!!"
+FIT_ELLIPSIS = "..."
+
+@pytest.mark.parametrize("max_length, ellipsis, error", [
```
UlysseFG, line 66: Nice ...

UlysseFG (approved):

PR #877, opened 2024-10-30
cleaned up some breaking changes from the upgrade

samouzegar10:


UlysseFG (approved):

frontend/src/components/readsets/ReadsetMetricContent.tsx
```diff
@@ -0,0 +1,72 @@
+import { ExpandableConfig } from "antd/lib/table/interface"
```
UlysseFG, line 1: You can clean up the imports that are not used ...

frontend/src/components/readsets/ReadsetsListContent.tsx
```diff
@@ -389,61 +388,6 @@ function useColumns(filters: FilterSet, readsetTableCallbacks: PagedItemsActions
     return columns
 }
 
-function useExpandableMetricConfig(): ExpandableConfig<ObjectWithReadset> {
-    return useMemo(() => ({
```
UlysseFG, line 393: Any reason for getting rid of the useMemo ?

frontend/src/components/readsets/ReadsetMetricContent.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const columnTitle: ReactNode = <Typography.Text>Metrics</Typography.Text>
+    function checkIfDecimal(str: string) {
+      const num = parseFloat(str)
+      if (String(num).includes('.')) {
+          return num.toFixed(3)
+      } else {
+          return num
+      }
+    }
+    return {
```
UlysseFG, line 19: useMemo ?

samouzegar10: 1. I dont think useMemo is necessary in this instance since it renders on a optionally filed variable (metric) with very few amounts of data in each.
2. I believe useMemo is to be used adequately, one of the key principles in of useMemo is to memoize (remember) the data for the next render, in this instance, i didn't see any substantial benefit, it wouldn't offer any increase in performance when it comes down to this component.

frontend/src/components/readsets/ReadsetsListContent.tsx
samouzegar10: sorry for the copy paste,

1. I dont think useMemo is necessary in this instance since it renders on a optionally filed variable (metric) with very few amounts of data in each.
2. I believe useMemo is to be used adequately, the key principle of useMemo is to memoize (remember) the data for the next render, in this instance, i didn't see any substantial benefit, it wouldn't offer any increase in performance when it comes down to this component.

UlysseFG (approved):

PR #882, opened 2024-11-14
reports page placeholder/skeleton

samouzegar10:
hello all, this is the first commit to an incomplete component/feature.
there are some data that do not belong in the feature itself (the dummy data and the dataset for the table)


nafiz1001:
We have a report page in place currently. We might revisit this PR if we wish to incorporate graphs.

PR #883, opened 2024-11-15
added tumor to index sheet index at the sample_kind column

samouzegar10:


UlysseFG:
As you mentioned the Tumor does not appear in the dropdown ... You need to change the cells format (data validation) to include the new cell. Also, make sure to save the file with the top left cell of each sheet selected and the first sheet active.

UlysseFG:
Tumour shows up, but can you reset the cursors to better positions ? When you save the current cursor position get saved along the file. This looks bad when the template cursor position is not set to a default position.
That default position is typically the A:1 cell for each sheet.
For the first sheet there is an image there you might have to resize that cell to make it so the borders don't clip into the image.

PR #886, opened 2024-11-20
changed the formula for calculating the Na Quantity (ng) by using cur…

samouzegar10:
…rent volume (uL) if Measured Volume (uL)


UlysseFG:
Sample QC formula is not correct. A value appears as soon as we set the current volume. The value is also incorrect when the Measured volume is missing.

UlysseFG:
Library QC formula is also wrong. The NA quantity value does not match the values entered.

UlysseFG:
You seem to be missing parenthesis around the volume calculation ...

UlysseFG:
Can you, once the formula are fixed, fix the formating (reset the cursor to A:1 on each sheet and use the same font for column "NA Quantity (ng)" as the rest for the template (Calibri)).

PR #887, opened 2024-11-25
Fixed sample submission template index dropdown. Removed GRCh37.

UlysseFG:


nafiz1001 (approved):
I don't see GRCh37 in Reference Genome and I can select index in a dropdown.

PR #890, opened 2024-12-02
Multi-Column sort for PagedItemsTable

nafiz1001:


frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -117,23 +118,109 @@ function PagedItemsTable<T extends object>({
...
+	const sortByCallback = useCallback((columnKey: string) => {
+		const [existingFirst = undefined, existingSecond = undefined] = pagedItems.sortByList
+		const newFirst: Partial<SortBy> = { ...existingFirst }
+		const newSecond: Partial<SortBy> = { ...existingSecond }
+	
+		if (!shiftHeld.current) {
+			if (existingFirst?.key === columnKey) {
+				if (existingFirst.order !== 'descend') {
+					newFirst.order = 'descend'
+				} else {
```
UlysseFG, line 151: Lot's of ifs ... what about always setting the key and using an enum with a numeric order on a modulo 3 ?

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -117,23 +118,109 @@ function PagedItemsTable<T extends object>({
...
+					newSecond.key = undefined
+					newSecond.order = undefined
+				}
+			} else if (existingSecond?.key !== columnKey) {
+				newSecond.key = columnKey
+				newSecond.order = 'ascend'
+			}
+		}
+	
+		if (newFirst.key === undefined || newFirst.key === newSecond.key) {
```
UlysseFG, line 177: might be simpler to fix this at the end when calling the back end ... you can just ignore undefined keys then... Less risk of falling out of sync or having our brain explode trying to track where are the sortBy gone.

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -117,23 +118,109 @@ function PagedItemsTable<T extends object>({
...
+		}, [])
+		setSortByCallback(newSortByList)
+	},
+	[pagedItems.sortByList, setSortByCallback])
+	const finalColumns: NonNullable<TableProps<T>['columns']> = useMemo(() => {
+		return columns.map((column) => {
+			if (column.sorter) {
+				const sortByIndex = pagedItems.sortByList.findIndex((x) => x.key === column.key)
+				const sortBy = sortByIndex >= 0 ? pagedItems.sortByList[sortByIndex] : undefined
+				console.info({ sortByIndex, sortBy })
```
UlysseFG, line 199: spam machine...

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -117,23 +118,109 @@ function PagedItemsTable<T extends object>({
...
+		}
+	
+		if (newFirst.key === undefined || newFirst.key === newSecond.key) {
+			newFirst.key = newSecond.key
+			newFirst.order = newSecond.order
+			newSecond.key = undefined
+			newSecond.order = undefined
+		}
+
+		console.info({ newFirst, newSecond })
```
UlysseFG, line 184: Spammation

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -117,23 +118,109 @@ function PagedItemsTable<T extends object>({
 		[dispatch, setPageSizeCallback]
 	)
 
+	const shiftHeld = useRef(false)
```
UlysseFG, line 121: What about sortPriority [0, 1] instead of shiftHeld to use directly in the sortByList...

frontend/src/components/WorkflowSamplesTable/WorkflowSamplesTable.tsx
```diff
@@ -37,20 +38,20 @@ export interface WorkflowSamplesTableProps {
...
 			setFilter,
-			setFilterOptions)
-		return mergedColumns
+			setFilterOptions
+		)
 	}, [columns, filterDefinitions, filterKeys, filters, setFilter, setFilterOptions])
-
-
+	// eslint-disable-next-line @typescript-eslint/no-empty-function
+	const tableColumns = useCustomMultiColumnSorter(columnsWithFilters, sortByList, setSortByList ?? (() => {}))
```
UlysseFG, line 54: You did not felt like typing () = {} as UndefinedCallback ?

nafiz1001: I could create a variable called `UndefinedCallback`.

UlysseFG:
I am not sure how this should behave ... i have never used that kind of feature (it uses the keyboard!). I am not sure i like that if you have 2 sorters and you un-sort the 1st the 2nd become the 1st (and you lose it if you click on another column). Also shift click, becomes a click if you don't have a 1st sorter already ... This is weird. I imagine if the 1st sorter column it to the right of the table and i have to scroll to it ... i will need to scroll all the way then back to make the sorting instead of setting the 2nd sort first and scroll to the 1st sort column... I think the shift click should always set the 2nd sorter even if there is no 1st sorter. I will try to find a place with that feature... maybe we can rely on something similar elsewhere ...

nafiz1001:
- [ ] if primary sort removed, remove secondary.
- [ ] some table columns are missing sort icons. e.g. Samples (Samples tab chosen), Libraries (Libraries tab chosen).
- [ ] Tables retain their sort even after changing page.

UlysseFG:
It act pretty good i think. The only issue i saw is that when moving from page to page it retains the ordering on the table but the indicators no longer shows the ordering that was active to provide these results. Not really breaking ... Just realized it was part of your to do list ...

UlysseFG (approved):
Comment is optional ...

nafiz1001:
> It act pretty good i think. The only issue i saw is that when moving from page to page it retains the ordering on the table but the indicators no longer shows the ordering that was active to provide these results. Not really breaking ... Just realized it was part of your to do list ...

The one in my to do list is actually slightly different. The fact that indicators don't show is kinda problematic: how would the user know which columns to unfilter?

UlysseFG:
> > It act pretty good i think. The only issue i saw is that when moving from page to page it retains the ordering on the table but the indicators no longer shows the ordering that was active to provide these results. Not really breaking ... Just realized it was part of your to do list ...
> 
> The one in my to do list is actually slightly different. The fact that indicators don't show is kinda problematic: how would the user know which columns to unfilter?

They actually don't need as far as i could see clicking on another sort resets the sorting and set the newly selected sort as only active sort ...
It is not obvious that the order is kept and it seem not to be causing much issues ... So it's up to you ... it would be problematic for someone that would expect the same sort to be active and move to another page from there ... I haven't tested that ...

nafiz1001:
Just a little review of what I did for container.py ;)

PR #891, opened 2024-12-02
Report Page

nafiz1001:


UlysseFG:
First overview :
- Drawer is looking good. I would personally like an alternate way to open it other than the button at the top. Bento has a floaty button, there may be a way to have a tab style button on the side of the screen.
- When we change the report in the drawer,  it does not change the table if the column are different (for exemple i removed a group_by and the old group_by column was still there empty ...
- The line separator look like an old school html separator it clashes with the app look.
- You are missing some of the latest changes to my branch ... You have access to time_window_labels now instead of the date.
- Time windows would look better as tabs instead of a dropdown i think.
- There is a title that is weird ... looks like it contains the last selected group_by ... It is not helpful ... Maybe try to include some of the content of the drawer selections in a more "title-y" style.
- I will add a type to the headers soon so we can better display the numbers.

frontend/src/components/app/App.js
```diff
@@ -117,30 +129,18 @@ const MENU_ITEMS = [
...
   {
     url: "/datasets",
     icon: <FileZipOutlined />,
     text: "Datasets",
     key: "datasets",
   },
+  {
+    url: "/report",
+    icon: <FlagOutlined />,
+    text: "Report",
```
UlysseFG, line 141: use plural ("Reports") here ... we'll get more eventually...
We can already generate multiple reports ...

UlysseFG:
also ...
- there is no button to export the current report in excel...
- When the report contains no data there is no message nothing it looks like nothing is happening.

frontend/src/components/reports/Reports.tsx
```diff
@@ -0,0 +1,373 @@
...
+        timeWindowData
+            ? timeWindowData.map<RecordType>((data, index) => {
+                const record = { ...data, key: index.toString() }
+                for (const [key, value] of Object.entries(data)) {
+                    switch (fieldNameToHeader[key].data_type) {
+                        case "boolean":
+                            record[key] = value ? "Yes" : "No"
+                            break
+                        case "date":
+                            record[key] = value ? value : "N/A"
```
UlysseFG, line 317: I don't like the "N/A"s i would prefer empty columns.

PR #892, opened 2024-12-06
Fix Table Colum Appearances

nafiz1001:


UlysseFG:
Very nice upgrades ...
Couple of comments : 
-The Taxon table is "weird". Column are too large ... by a lot. Not sure it should even scroll horizontally.
- Some table are very compact ... but they occupy only half the height of the page so you still need to scroll a lot. Maybe we could have the height of the table be a prop so we could tailor it to each page available space ? This may require more work than this ticket is supposed to imply but it would hit the same target get more visible rows without scrolling.
- Experiment run could be compacted a bit by giving more space to the launch column and reducing a bit the arrow button.

frontend/src/components/samples/SampleTableColumns.tsx
```diff
@@ -49,12 +49,17 @@ export enum SampleColumnID {
 	SAMPLE_COUNT = 'SAMPLE_COUNT',
 }
 
+const SMALL_COLUMN_WIDTH = 90
```
UlysseFG, line 52: Maybe we need TINY around 40 ? This system is not very flexible but it simplify the management ... We could add a couple more steps to make it easier to find one that fits.

UlysseFG:
Last version i reviewed was better ...
we have now the same old issue with some table fields (names, project, index, ...) that wrap and make the rows harder to see all together...
I think the tables with issues are the ones you just prevented from extending horizontally past the size of the screen. This works well for experiment but project seem too tightly compacted, library have its index wrapped over many rows...
It might be hard to have a generic solution without tweaking here and there.

nafiz1001:
> Last version i reviewed was better ... we have now the same old issue with some table fields (names, project, index, ...) that wrap and make the rows harder to see all together... I think the tables with issues are the ones you just prevented from extending horizontally past the size of the screen. This works well for experiment but project seem too tightly compacted, library have its index wrapped over many rows... It might be hard to have a generic solution without tweaking here and there.

(╯°□°)╯︵ ┻━┻

frontend/src/components/libraries/LibrariesListContent.tsx
```diff
@@ -67,26 +67,28 @@ export default function LibariesListContent() {
...
-			[LibraryColumnID.INDEX_NAME]: 130,
-			[LibraryColumnID.VOLUME]: 130,
-			[LibraryColumnID.LIBRARY_SIZE]: 130,
-			[LibraryColumnID.CONCENTRATION_NM]: 130,
-			[LibraryColumnID.CONCENTRATION]: 160,
-			[LibraryColumnID.NA_QUANTITY]: 130,
-			[LibraryColumnID.QC_FLAG]: 130,
-			[LibraryColumnID.CREATION_DATE]: 130,
-			[LibraryColumnID.DEPLETED]: 130
+			[LibraryColumnID.INDEX_NAME]: LARGE_COLUMN_WIDTH*1.5,
```
UlysseFG, line 83: ugh

nafiz1001: 🤭

UlysseFG (approved):
Not perfect but will fix the current issue

PR #894, opened 2024-12-16
Add Infinium epic 8 breadchip

nafiz1001:


backend/fms_core/containers.py
```diff
@@ -5,6 +5,7 @@
 __all__ = [
     "ContainerSpec",
 
+    "CONTAINER_SPEC_INFINIUM_EPIC_8_BREADCHIP",
```
UlysseFG, line 8: Nope not bReadchip

backend/fms_core/containers.py
```diff
@@ -167,6 +168,14 @@ def __str__(self):
     is_run_container=True,
 )
 
+CONTAINER_SPEC_INFINIUM_EPIC_8_BREADCHIP = ContainerSpec(
+    container_kind_id="infinium epic 8 breadchip",
```
UlysseFG, line 172: beadchip ...

backend/fms_core/migrations/0067_alter_container_kind_alter_experimentrun_container_and_more.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.2.4 on 2024-12-16 15:55
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('fms_core', '0066_v4_12_0'),
```
UlysseFG, line 10: move this to the upcomming version migration file ...

UlysseFG:
Looking at the template makes me think we might have a few more things to fix ...
The template name should be renamed Experiment_infinium_v_4_13_0.xlsx (removed the 24 since this was a reference to the beadchip and updated the version).
The template also mention the run type that is specific to the other beadchip version ... we should migrate the name of the run type to something more generic like Inifinium and change the refence in the template.
We should also double check the template ingestion to make sure there is no issue using a 8 beadchip instead of a 24 (the pipeline will likely include 12 beadchips at once (96 samples).

backend/fms_core/containers.py
```diff
@@ -167,6 +168,14 @@ def __str__(self):
     is_run_container=True,
 )
 
+CONTAINER_SPEC_INFINIUM_EPIC_8_BEADCHIP = ContainerSpec(
+    container_kind_id="infinium epic 8 beadchip",
+    coordinate_spec=(alphas(8), ints(1, pad_to=2)), # 8 lanes
```
UlysseFG, line 173: This should be the other way around ... alphas(1), ints(8,

frontend/src/components/experimentRuns/ExperimentRunsDetachedFilters.ts
```diff
@@ -6,7 +6,7 @@ export enum RUN_TYPES {
     ILLUMINA = "Illumina",
     AXIOM = "Axiom",
     DNBSEQ = "DNBSEQ",
-    INFINIUM =  "Infinium Global Screening Array-24",
+    INFINIUM =  "Infinium",
```
UlysseFG, line 9: nice find ...

UlysseFG:
The coordinate dropdown in the template (you can let the test template alone) would really be best with appropriate coordinates...

backend/fms_core/migrations/0067_v4_13_0.py
```diff
@@ -1,7 +1,11 @@
 # Generated by Django 4.2.4 on 2024-11-15 18:29
 
 from django.db import migrations, models
+import django.db.models.deletion
 
+def rename_infinium_runtype(apps, schema_editor):
+    RunType = apps.get_model('fms_core', 'RunType')
+    RunType.objects.filter(name='Infinium Global Screening Array-24').update(name='Infinium')
```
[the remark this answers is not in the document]
nafiz1001: I don't think it's necessary ;)
`get` is for obtaining a single object i think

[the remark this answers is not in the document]
nafiz1001: ah but `.save` might be needed for custom saving.

UlysseFG (approved):

PR #895, opened 2024-12-16
Create endpoint to set experiment run end time.

UlysseFG:


UlysseFG:
> Hmm... So who's poking the endpoint? I'm guessing the techdev?

Normally it would be the monitor ; ). Tech Dev reported the lab needing a validation that files were copied out of the machine before removing them. This should be a good marker to point out the experiment run is complete. So i would say the monitor check if the run complete file is present then validate the files are all copied (there might be a file for that too) then the monitor reports the run experiment end_time ... that way we can report all clear on the experiment run table for the lab to do their clean up.

nafiz1001:
You're ok with timezone not being UTC?

nafiz1001 (approved):

UlysseFG:
> You're ok with timezone not being UTC?

Time is UTC as far as i know ...it's 5 hours later than the current time ...

UlysseFG:
The server is set as UTC.

nafiz1001:
> The server is set as UTC.

Wouldn't it be better to explicitly use UTC time?
When I tested locally, the time wasn't UTC, at least as it appears in DBeaver.

UlysseFG:
> > The server is set as UTC.
> 
> Wouldn't it be better to explicitly use UTC time? When I tested locally, the time wasn't UTC, at least as it appears in DBeaver.

This is pretty much as explicit as can be ... timezone.now() uses the settings.py definition.

PR #899, opened 2024-12-20
correct coordinate spec for infinium

nafiz1001:


UlysseFG (approved):

nafiz1001:
I should test with a template.

PR #900, opened 2024-12-20
Fix filter labwork

UlysseFG:


UlysseFG:
Looking at the ordering and the counts ... this is broken still...

PR #902, opened 2025-01-08
Add Annually as the default time window

UlysseFG:
- New option to have Annual time window which is now the default.
- Excel exported file column width should now be better fitted to the data length.


nafiz1001 (approved):
I got annual time window on web page and excel.

PR #904, opened 2025-01-09
QC flags can be edited on the edit sample form as long as they have been set once using templates.

UlysseFG:


frontend/src/components/samples/SampleEditContent.js
```diff
@@ -88,6 +89,8 @@ const SampleEditContent = ({ sample, isAdding}) => {
 
   const [form] = Form.useForm()
   const [formErrors, setFormErrors] = useState({})
+  const [disabledQualityFlag, setDisabledQualityFlag] = useState(true)
+  const [disabledQuantityFlag, setDisabledQuantityFlag] = useState(true)
```
nafiz1001, line 93: Do you really need to use useState 🤔 
Especially considering that `setDisabledQualityFlag` and `setDisabledQuantityFlag` seem to be called only once in the component, and they are only called with `sample` variable that's already passed as a prop to the component.

nafiz1001:
The colour contrast looks kinda awkward.
image

UlysseFG:
> The colour contrast looks kinda awkward. 
Maybe i can enlarge the icon to make it more green. not much to be done about green being green.

nafiz1001:
> > The colour contrast looks kinda awkward.
> 
> Maybe i can enlarge the icon to make it more green. not much to be done about green being green. (you're welcome for the edit - Nafiz)

Another option could be to just have a white check-mark (with transparent background) in a blue switch.

nafiz1001 (approved):

PR #905, opened 2025-01-14
tweaks

nafiz1001:


frontend/src/components/labwork/step/LabworkStepOverview.tsx
```diff
@@ -56,7 +56,7 @@ const LabworkStepOverview = ({step, refreshing, stepSamples, columns, filterDefi
 
   useEffect(() => {
     dispatch(getLabworkStepSummary(step.id, activeGrouping.key, {}))
-  }, [])  // Fetches the initial labwork step summary
+  }, [activeGrouping.key, dispatch, step.id])  // Fetches the initial labwork step summary
```
UlysseFG, line 59: i think this was meant to fire only once ... some other use effect triggers to refresh the grouping and step ... I don't recall where but at some place there was an infinite loop related to circular dependencies ...

frontend/src/components/labwork/step/LabworkStepOverviewPanel.tsx
```diff
@@ -39,7 +39,7 @@ const LabworkStepOverviewPanel = ({ stepID, refreshing, grouping, groupingValue,
 
 	const [isFetchingSamples, setIsFetchingSamples] = useState(false)
 	const [sampleAndLibraryList, setSampleAndLibraryList] = useState<SampleAndLibrary[]>([])
-  const displayedSamples = useAppSelector(selectLabworkStepsState).steps[stepID].displayedSamples
+  const displayedSamples = useAppSelector((state) => selectLabworkStepsState(state).steps[stepID].displayedSamples)
```
UlysseFG, line 42: isn't the useSelector used to prevent having to pass the state around all the time ?

nafiz1001: No. It's to access redux state. Generally you want your selector to be precise to minimize component rerendering since it uses the return value of the selector to check for changes in value.

nafiz1001: If you just have it return from selectLabworkStepsState, any changes under the entire LabworkStepsState will cause rerendering, including unrelated steps.

UlysseFG: oh i see it's like creating a new selector on the fly. Make you wonder what is gained using the selector in the first place though.

nafiz1001: To avoid rerendering. Maybe the cost of rerendering is higher than creating a new selector which is just a lambda?

PR #907, opened 2025-01-16
Allow json template to be read for template_importers.

UlysseFG:


backend/fms_core/template_importer/sheet_data.py
```diff
@@ -13,20 +13,24 @@
...
         self.headers = headers
+        self.shared_data = shared_data # This is additional information that is not assigned to a specific row. None for xls templates.
 
-        for i, row_list in enumerate(self.dataframe.values.tolist()):
-            if row_list[:len(self.headers)] == self.headers:
-                self.dataframe.columns = row_list
-                self.header_row_nb = i
-                break
+        if self.shared_data is not None: # This is not defined for xls templates
+            self.header_row_nb = -1 # No header in dataframe for json files
```
nafiz1001, line 27: So far it's not clear to me the use of `shared_data`.

backend/fms_core/tests/test_template_importers/json_templates_test.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_import(self):
+        
+        for i, importer in  enumerate(self.importers):
+            # Basic test for all templates - checks that template is valid
+            result = load_template(importer=importer, file=self.files[i])
+            self.assertEqual(result['valid'], True)
+            
+            # Custom tests for each template
+            match i:
+                case 0: # Sample QC
```
nafiz1001, line 58: Very interesting. I didn't expect you to use new Python language syntax ;)

backend/fms_core/template_importer/sheet_data.py
UlysseFG: This is to hold the information that would go above the header in the templates ... There is a few templates with values there already. In the case here we use it to tell appart json and xls templates.

backend/fms_core/tests/test_template_importers/json_templates_test.py
nafiz1001, line 1: Since you're testing against a JSON, I wonder if it would be more convenient if instead you accessed the values in the JSON directly (as an object) for making assertions like `self.assertEqual(sample.volume, self.json["Current Volume (uL)"] - self.json["Volume Used (uL)"])`

UlysseFG: There is not much to be gained by that ... some fields are calculated some stuff are generated from the DB ... This would complicate things without making it much more reliable. It would imply to import into the test some calculations that could evolve too which would duplicate the maintenance.

nafiz1001 (approved):

PR #910, opened 2025-01-21
Fix labwork for library transfer

nafiz1001:


frontend/src/components/libraries/LibraryTableColumns.tsx
```diff
@@ -75,6 +75,7 @@ export const LIBRARY_COLUMN_DEFINITIONS: { [key in LibraryColumnID]: LibraryColu
 				)
 			)
 		},
+		width: 200,
```
UlysseFG, line 78: I would put at least 300 here. Index names are very long ...

UlysseFG (approved):

PR #914, opened 2025-01-23
Tweak to the endpoint to filter with new annotation.

UlysseFG:


nafiz1001 (approved):
LGTM
hopefully it doesn't break something else ;')

PR #917, opened 2025-01-28
Fix date range picker

nafiz1001:


UlysseFG (approved):
Cannot replicate the issue anymore.

PR #918, opened 2025-01-29
Workflow Queueing

nafiz1001:


frontend/src/models/paged_items_factory.ts
```diff
@@ -161,6 +165,9 @@ export function createPagedItemsActions<Prefix extends string, M extends FMSTrac
             
             // The paged items reducer just needs the item ID's, not the actual
             // items that were retrieved, so extract the list of ID's from the data.
+            if (!reply.results) {
+                console.error(LIST_PAGE, params, reply)
+            }
```
nafiz1001, line 170: I should remove this.

backend/fms_core/services/sample.py
```diff
@@ -1048,4 +1049,24 @@ def can_remove_sample(sample: Sample) -> Tuple[bool, List[str], List[str]]:
         was_processed = ProcessMeasurement.objects.filter(source_sample_id=sample.id).exists()
         is_removable = not is_child and not is_parent and not was_processed
 
-    return is_removable, errors, warnings
\ No newline at end of file
+    return is_removable, errors, warnings
+
+T = TypeVar("T")
+def defaultSelection_exceptedIDs(queryset: QuerySet[T], default_selection: bool, excepted_ids: list[int]) -> QuerySet[T]:
```
UlysseFG, line 1055: This does not belong in sample service.

UlysseFG: We typically want in services stuff that can be both use from the API and from the other parts of the backend ... this is totally specific to a single endpoint.

backend/fms_core/services/sample.py
```diff
@@ -1048,4 +1049,24 @@ def can_remove_sample(sample: Sample) -> Tuple[bool, List[str], List[str]]:
...
+    """
+    Filter a queryset based on the default_selection and excepted_ids.
+    If default_selection is True, the queryset will exclude the excepted_ids.
```
UlysseFG, line 1058: This goes against what i would expect by the name. default selection would indicate a typical filter rather than an exclusion you seem to be applying your front end logic to the backend name scheme and it makes it harder to understand. This little function is barely useful as is if you have to read it to understand it you are removing all gain.
I strongly suggest you just add those lines to your viewset they need context and this function bring the worse part of frontend design to the backend (scattering the code all over the place).

backend/fms_core/viewsets/sample.py
```diff
@@ -416,19 +417,17 @@ def versions(self, _request, pk=None):
 
     @action(detail=False, methods=["post"])
     def add_samples_to_study(self, request, pk=None):
-        excepted_sample_ids = request.data.get("excepted_sample_ids")
         default_selection = request.data.get("default_selection", False)
```
UlysseFG, line 420: I would change the name to something more frontend agnostic ... The default being unclear in this case. Maybe call it "negative_selection" indicating that you remove the ids listed from the selection ?

backend/fms_core/viewsets/sample.py
```diff
@@ -416,19 +417,17 @@ def versions(self, _request, pk=None):
...
+        excepted_sample_ids = request.data.get("excepted_sample_ids", [])
         project_id = request.data.get("project_id")
         study_letter = request.data.get("study_letter")
         step_order = request.data.get("step_order", None)
 
         samples = self.filter_queryset(self.get_queryset())
+        samples = defaultSelection_exceptedIDs(samples, default_selection, excepted_sample_ids)
```
UlysseFG, line 427: The except option makes it that much harder to optimize the query ... Instead of limiting the query to X samples we have to work with N - X which likely forces a full scan of the table(and is slow as all hell). I am curious if only using positive selection (default_selection=false) would be faster. Might be worth investigating ...

backend/fms_core/viewsets/sample.py
```diff
@@ -444,4 +443,14 @@ def add_samples_to_study(self, request, pk=None):
         if errors:
             raise ValidationError(errors)
         else:
-            return Response(status=204)
\ No newline at end of file
+            return Response(status=204)
+
+    @action(detail=False, methods=["post"])
+    def sample_ids_by_default_selection_excepted_ids(self, request):
+        default_selection = request.data.get("default_selection", False)
```
UlysseFG, line 450: default selection is false by default ?

frontend/src/utils/api.ts
```diff
@@ -288,7 +293,7 @@ const api = {
     get: studyId => get<JsonResponse<FMSStudy>>(`/studies/${studyId}/`),
     add: study => post("/studies/", study),
     update: study => patch(`/studies/${study.id}/`, study),
-    list: (options, abort?: boolean) => get<JsonResponse<FMSPagedResultsReponse<FMSStudy>>>('/studies', options, {abort}),
+    list: (options, abort?: boolean) => get<JsonResponse<FMSPagedResultsReponse<FMSStudy>>>('/studies/', options, {abort}),
```
UlysseFG, line 296: the evil "/" strikes again!

frontend/src/utils/api.ts
```diff
@@ -230,6 +230,9 @@ const api = {
       request: (options, template) => filteredpost(`/samples/prefill_template/`, {...options}, form({ template: template })),
     },
     search: q => get("/samples/search/", { q }),
+    sample_ids_by_default_selection_excepted_ids(defaultSelection: boolean, exceptedSampleIDs: Array<FMSSample['id']>, queryParams?: QueryParams) {
```
UlysseFG, line 233: why the different notation? if it's too long you can remove some typing ;).

backend/fms_core/viewsets/sample.py
```diff
@@ -444,4 +443,14 @@ def add_samples_to_study(self, request, pk=None):
         if errors:
             raise ValidationError(errors)
         else:
-            return Response(status=204)
\ No newline at end of file
+            return Response(status=204)
+
+    @action(detail=False, methods=["post"])
+    def sample_ids_by_default_selection_excepted_ids(self, request):
```
UlysseFG, line 449: that is really ugly you are building the backend around your frontend. I am not sure why you would need that ... It also likely destroy any benefit you gain from paging ... I will read on and edit my comment but the ugliness will remain (i don't want to forget to mention it ;).

frontend/src/utils/api.ts
nafiz1001: Just to add variety ;)

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -62,6 +63,35 @@ def destroy(self, request, pk=None):
...
+    @action(detail=False, methods=["post"])
+    def destroy_list(self, request):
+        sample_ids = request.data.get("sample_ids", [])
+        study = request.data.get("study", None)
+        stepOrder = request.data.get("step_order", None)
+        if not sample_ids:
+            return HttpResponseBadRequest("No sample IDs provided.")
+        if not study:
+            return HttpResponseBadRequest("No study ID provided.")
+        if not stepOrder:
```
UlysseFG, line 75: be careful testing against falsy values ... 0 is false and is often a valid value ... test against the default value as much as possible instead (typically "is not None", but also [] and {}), Here step order start at 1, but this is never a bad thing to test for a more specific value.

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -62,6 +63,35 @@ def destroy(self, request, pk=None):
...
+            return HttpResponseBadRequest("No step order provided.")
+        queryset = self.get_queryset().filter(sample_next_step__sample__id__in=sample_ids, study=study, step_order__order=stepOrder)
+        values_list = queryset.values_list("sample_next_step__sample", "study", "step_order__order")
+        errors = []
+        removed = {}
+        try:
+            with transaction.atomic():
+                for sample_id, study_id, order in values_list:
+                    sample = Sample.objects.get(id=sample_id)
```
UlysseFG, line 84: could be very DB intensive ... might be worth batching the sample and study objects queries (stored in a dict). The study is likely to be the same over and over ... PostGreSQL does some caching but it is better not to rely on the DB caching.

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -62,6 +63,35 @@ def destroy(self, request, pk=None):
...
+                    study = Study.objects.get(id=study_id)
+                    newremoved, newerrors, _ = dequeue_sample_from_specific_step_study_workflow_with_updated_last_step_history(sample, study, order)
+                    errors.extend(newerrors)
+                    removed[sample_id] = newremoved
+                if errors:
+                    raise IntegrityError(errors, removed)
+        except IntegrityError as err:
+            return ValidationError(err)
+        return Response(data={"details": removed}, status=status.HTTP_200_OK)
```
UlysseFG, line 93: i recommend adding a similar return list to the queueing endpoint this might prevent having to ask the backend for list of IDs.

frontend/src/components/app/App.js
```diff
@@ -57,6 +59,7 @@ import { useAuthInit } from "./useAuthInit";
 import { useRefreshHook } from "./useRefreshHook";
 import InstrumentsRoute from "../instruments/InstrumentsRoute";
 import { Reports } from "../reports/Reports";
+import { LabworkSamplesPage } from "../management/LabworkSamplesPage";
```
UlysseFG, line 62: labworksamplespage is not a good name for this page... WorkflowAssigmentPage would be more explanatory or WorkflowQueueManagerPage... ?

frontend/src/components/containers/ContainersDetailContent.tsx
```diff
@@ -74,6 +70,9 @@ const ContainersDetailContent = () => {
         extra={
           !isLoaded ? null :
             <Space>
+              <Button onClick={() => navigate(
+                `/lab-work/samples?${(container?.children.length ?? 0) > 0 ? 'container__location__barcode=' : 'container__barcode='}${container?.barcode}`
+              )}>Workflow</Button>
```
UlysseFG, line 75: Workflow may be a bit too little. Manage workflows ? Add/Remove Workflow ?

frontend/src/components/containers/ContainersDetailContent.tsx
```diff
@@ -74,6 +70,9 @@ const ContainersDetailContent = () => {
         extra={
           !isLoaded ? null :
             <Space>
+              <Button onClick={() => navigate(
+                `/lab-work/samples?${(container?.children.length ?? 0) > 0 ? 'container__location__barcode=' : 'container__barcode='}${container?.barcode}`
```
UlysseFG, line 74: It might be nice (optional) to find a way to disable the button for containers that are not candidates for workflow management (freezers, rooms, drawers, freezer racks, ...)

frontend/src/components/management/LabworkSamples.tsx
```diff
@@ -0,0 +1,457 @@
...
+import SamplesTableActions from '../../modules/samplesTable/actions'
+import serializeFilterParamsWithDescriptions from "../pagedItemsTable/serializeFilterParamsTS";
+
+const MAX_SELECTION = 960
+
+interface LabworkSamplesProps {
+    fixedFilter?: FilterSetting
+}
+
+export function LabworkSamples({ fixedFilter }: LabworkSamplesProps) {
```
UlysseFG, line 26: Name could be better...

frontend/src/components/management/LabworkSamples.tsx
```diff
@@ -0,0 +1,457 @@
...
+        id: Project['id']
+        name: Project['name']
+    }
+    const [commonProjects, setCommonProjects] = useState<CommonProject[]>([])
+
+    const [selectedProject, setSelectedProject] = useState<Project['id']>()
+    const [selectedStudyWorkflow, setSelectedStudyWorkflow] = useState<StudyWorkflow>()
+
+    const refreshActions = useCallback(async () => {
+        const sampleIDs = (await dispatch(api.samples.sample_ids_by_default_selection_and_excepted_ids(
```
UlysseFG, line 165: I am not convinced you need that endpoint ... We can ask for all samples matching a filter and exclude locally ... or we can add an exclude query param for the list endpoint. You are fetching the same sample from id the very next line ... The fetchSamples will get you the count too...

backend/fms_core/viewsets/sample.py
nafiz1001: When it's false, it's normal selection.

backend/fms_core/viewsets/sample.py
nafiz1001: It's a trade-off between CPU and memory.

backend/fms_core/viewsets/sample.py
nafiz1001: That endpoint is intended to be a compromise in case there are more pages that use PagedItemsTable with selection enabled.

frontend/src/components/management/LabworkSamples.tsx
nafiz1001: Fair enough... I can check the exclude query param path.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,454 @@
...
+        id: Project['id']
+        name: Project['name']
+    }
+    const [commonProjects, setCommonProjects] = useState<CommonProject[]>([])
+
+    const [selectedProject, setSelectedProject] = useState<Project['id']>()
+    const [selectedStudyWorkflow, setSelectedStudyWorkflow] = useState<StudyWorkflow>()
+
+    const refreshActions = useCallback(async () => {
+        const samples = await dispatch(fetchSamplesByDefaultSelectionAndExceptedIDs(defaultSelection, exceptedSampleIDs, filters))
```
UlysseFG, line 166: This has no use ... it needs to be replaced by a call to sample list ...

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,454 @@
...
+            return acc
+        }, {})
+
+        type StudyStepKey = `${Study['id']}-${Step['name']}`
+        const sampleNextSteps = (await dispatch(api.sampleNextStep.listSamples(sampleIDs))).data.results
+        const studyStepCount: Record<string, number> = {}
+        const sampleIDsByStudyStep: Record<StudyStepKey, Set<Sample['id']>> = {}
+        for (const sampleNextStep of sampleNextSteps) {
+            for (const study of sampleNextStep.studies) {
+                const key: StudyStepKey = `${study}-${sampleNextStep.step.name}`
```
UlysseFG, line 208: study step should not be identified by step name ... step name can be present multiple time in a workflow study. I don't see a logic where this would not be important ...You are likely to prevent queueing samples to steps they are not currently queued to.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,454 @@
...
+
+                studyStepCount[key] = (studyStepCount[key] ?? 0) + 1
```
UlysseFG, line 210: If the key is not unique this count will be wrong too.

backend/fms_core/viewsets/sample.py
nafiz1001: It's gone.

backend/fms_core/filters.py
```diff
@@ -67,6 +67,7 @@ class Meta:
         }
 
 class SampleFilter(GenericFilter):
+    id__not__in = django_filters.CharFilter(method="id_not_in_filter")
```
UlysseFG, line 70: I tend to think this might be moved to the Generic base class ... We should use it sparingly but it could be used elsewhere ... I tend to think an numeric filter would be faster since this is numeric field in the DB (it possibly has to cast each values to a char)

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,448 @@
...
+
+        setIsFetching("Loading projects...")
+        const countByProjectID = Object.values(projectsBySample).reduce<Record<Project['id'], number>>((acc, projects) => {
+            for (const project of projects) {
+                acc[project] = (acc[project] ?? 0) + 1
+            }
+            return acc
+        }, {})
+        const projectIDs = Object.keys(countByProjectID).map((id) => parseInt(id))
+        const commonProjects = (await fetchProjects(projectIDs)).map((project) => ({
```
UlysseFG, line 235: we should be using fetchX more often across the app ... This may help reduce calls to the backend...

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,448 @@
...
+            studyWorkflowsByProject[study.project_id] = studyWorkflows
+
+            for (const stepOrder of workflow.steps_order) {
+                if (stepOrder.order < study.start || stepOrder.order > study.end) continue
+                queueActions.push({
+                    study: study,
+                    workflow: workflow.name,
+                    step: stepOrder.step_name,
+                    stepOrder: stepOrder.order,
+                    alreadyQueued: [...sampleIDsByStudyStep[`${study.id}-${stepOrder.step_name}`] ?? []]
```
UlysseFG, line 293: You changed the key when creating the object this will likely never match anything.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,448 @@
...
+                        key: 'queue',
+                        label: `Queue Options (${projectQueueActions.length})`,
+                        children: <Flex wrap gap={"small"}>{projectQueueActions}</Flex>,
+                        classNames: { body: 'labwork-samples-actions' },
+                    }
+                ]}/>
+            )
+        }
+        return result
+    }, [commonProjects, defaultSelection, dequeueActions, dispatch, exceptedSampleIDs, filters, isFetching, projectByID, queueActions, refreshActions, sampleByID, selectedProject, selectedStudyWorkflow, studyWorkflowsByProject])
```
UlysseFG, line 441: not sure how valuable is the useMemo in this case ... each time basically anything change in the page you have to calculate it again ... Might be more useful to have more targeted ones. To me it looks like an umbrella against unintended re-renders. Might be more efficient to fix the unintended re-renders at the source (unnecessary useAppSelectors ?).

frontend/src/models/hooks.ts
```diff
@@ -0,0 +1,42 @@
+import { useCallback, useEffect } from "react"
+import { useSearchParams } from "react-router-dom"
+import { FilterDescription, FilterValue } from "./paged_items"
+import { PagedItemsActionsCallbacks } from "../components/pagedItemsTable/PagedItemsTable"
+
+export function useQueryParamsForPagedItems(
```
UlysseFG, line 6: Maybe this should be in the component folder for pagedItemsTable ?

frontend/src/hooks.tsx
```diff
@@ -0,0 +1,172 @@
...
+import api from './utils/api'
+import React from 'react'
+import { Popover, Spin, Tag } from 'antd'
+import { fetchStudies } from './modules/cache/cache'
+
+// Use throughout your app instead of plain `useDispatch` and `useSelector`
+export const useAppDispatch: () => AppDispatch = useDispatch
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
+
+export function useSampleAndLibraryList(sampleIDs: FMSId[], offset?: number, pageSize?: number) {
```
UlysseFG, line 17: you want to do away with the component structure for hooks ? These do not look like they are widely applicable in the app.

UlysseFG: Are you sure you want your hooks to return JSX elements? This seem to me that this may limit the reusability of your hooks.

UlysseFG:
General opinions ... The drawer content is very compact ... maybe too much. When you open a collapsible for queueing or dequeueing i think having a vertical array of options with similar size would be preferable. If you want to prevent the scrolling, perhaps having thinner buttons could alleviate that issue.

UlysseFG:
I am the already queued greying of steps seem to be broken atm. Possibly due to the comment i posted earlier.

UlysseFG:
Manage workflow button in container detail has not been updated with the correct URL after moving the component.

frontend/src/components/app/App.js
```diff
@@ -111,6 +114,19 @@ const MENU_ITEMS = [
...
   },
+  {
+    icon: <CarryOutOutlined />,
+    text: "Management",
+    key: "management",
+    children: [
+      {
+        url: "/management/workflow-samples",
+        icon: <RetweetOutlined />,
+        text: "Workflow Samples",
```
UlysseFG, line 125: Workflow assignment would be better here too.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -0,0 +1,448 @@
...
+                            }))
+                            await refreshActions()
+                        } catch (error) {
+                            dispatch(notifyError({
+                                id: `LabworkSamples_${action.study.id}_${action.stepOrder}`,
+                                title: "Error dequeuing samples from workflow",
+                                description: `Failed to dequeue samples from study ${action.study.letter} at step "${action.step}" for project ${project.name}".`
+                            }))
+                        }
+                    }} type="primary">{`${action.stepOrder}. ${action.step}`}</Button>
```
UlysseFG, line 381: I would prefer " - " instead of ". " after stepOrder.

frontend/src/components/management/WorkflowAssignment.tsx
nafiz1001: The sad thing about fetch is it doesn't handle filtering...

frontend/src/components/management/WorkflowAssignment.tsx
nafiz1001: It's more of a convenience of being in a function than useMemo ;)
It let's me take advantage of return to exit out with a component result sooner.

frontend/src/hooks.tsx
nafiz1001: I've decided to undo that unnecessary change and a few others not related to labwork samples

frontend/src/components/management/WorkflowAssignment.tsx
UlysseFG: eheh you might be the only one i know thinking a useMemo with 10 dependencies is easier to understand than an if ;). that fine you can keep it ;).

frontend/src/components/containers/ContainersDetailContent.tsx
```diff
@@ -74,6 +70,9 @@ const ContainersDetailContent = () => {
         extra={
           !isLoaded ? null :
             <Space>
+              <Button onClick={() => navigate(
```
UlysseFG, line 73: There goes another comment ... I meant the containers that are unlikely to be useful for mass queueing. A rack or a box are not sample holding but they can be used to queue samples in tubes. Queueing all sample in a freezer is unlikely to be a good option. Buut let this go i guess the empty list should be enough to pass the message ;)

UlysseFG:
- The tootip above the header of the name column is very annoying ... i keep clicking it by accident.
- Aligning the text to the left on the queueing / dequeueing button would be nice.
- adding spaces before and after the "-" would make button more readable.
- It's weird that when moving from the workflow management to the sample table it keeps the same filter ...
- Queueing orders the steps in the queueing order in the dequeue options part. It can end up listing step 8, 6, 7.
- Workflow assignment seem to be too long for the menu width. maybe we can use Assign Workflow instead...

UlysseFG (dismissed):

UlysseFG (commented):
Just had a problem with a sample that led to a white page ... I filtered by project BQC. ordered by sample name ascending selected the second sample (0002), clicked the queue /dequeue button then selected project BQC19 in the dropdown ... i got a white page and the following console error : Uncaught TypeError: studyWorkflowsByProject[selectedProject] is undefined.

UlysseFG (approved):

PR #919, opened 2025-01-29
Auto archived comments

UlysseFG:


nafiz1001:
So far I see that it creates comment for passing and failing automatically. However, the frontend does not automatically show the new (automated) comment; I had to reload the page.

nafiz1001:
For 'Dataset release status revoked.' I believe we should specify which dataset because that comment is only generated when you reset dataset individually.

backend/fms_core/services/archived_comment.py
```diff
@@ -0,0 +1,45 @@
...
+    Returns:
+        Tuple including the created archived comment instance as well as a list of errors and warnings.
+    """
+    archived_comment = None
+    errors = []
+    warnings = []
+
+    content_type_dataset = ContentType.objects.get_for_model(object_model)
+    try:
+        archived_comment = ArchivedComment.objects.create(content_type=content_type_dataset, object_id=object_id, comment=comment)
```
nafiz1001, line 40: It looks like the function is intended to be model agnostic, but dataset is still referenced here.

UlysseFG:
> So far I see that it creates comment for passing and failing and setting back to available automatically. However, the frontend does not automatically show the new (automated) comment; I had to reload the page.

Yeah i have to fix this on another PR ... This is an already existing issue.

UlysseFG:
> > For 'Dataset release status revoked.' I believe we should specify which dataset because that comment is only generated when you reset dataset individually. It also shows up in the lane validation comment, so it's not clear which dataset got revoked.
> 
> This is especially important for multi-project run.

It's only important for those since otherwise the association lane dataset is 1-to-1.

frontend/src/components/experimentRuns/ExperimentRunValidation.tsx
```diff
@@ -63,38 +63,32 @@ function ExperimentRunValidation({ experimentRunName }: ExperimentRunValidationP
...
 		(lane: LaneInfo) => {
 			setIsValidationInProgress(true)
-			dispatch(setRunLaneValidationStatus(lane, ValidationStatus.PASSED))
-				.finally(() => {
-          lane.datasets.map((dataset) => {dispatch(get(dataset.datasetID))})
-          setIsValidationInProgress(false)
-        }
-      )
+			dispatch(setRunLaneValidationStatus(lane, ValidationStatus.PASSED)).finally(() => {
+				Promise.allSettled(lane.datasets.map((dataset) => dispatch(get(dataset.datasetID)))).finally(() =>
```
UlysseFG, line 67: We don't need to wait for it ... it'll update on its own time ... if we wait for all async calls it kind of defeat the purpose...

nafiz1001: Fine.

nafiz1001 (approved):

PR #921, opened 2025-02-03
Collapsible validation time fix.

UlysseFG:


frontend/src/modules/experimentRunLanes/actions.ts
```diff
@@ -44,6 +46,28 @@ export function setRunLaneValidationStatus(lane: LaneInfo, status: ValidationSta
 	}
 }
 
+export function setRunLaneValidationTime(lane: LaneInfo) {
+	return async (dispatch, getState) => {
+    const validationTime = lane.datasets.reduce<string | undefined>((latest, dataset) => {
+      const currentDatasetTime = getState().datasets.itemsByID[dataset.datasetID].latest_validation_update
```
nafiz1001, line 52: I would rename `currentDatasetTime` to `latestValidationTime`.

frontend/src/modules/experimentRunLanes/reducers.ts
```diff
@@ -56,6 +57,18 @@ function reducers(state: Draft<ExperimentRunLanesState>, action: AnyAction): Exp
...
+			const experimentRunLanes = state.runs[experimentRunName] as ExperimentRunLanes
+			if (experimentRunLanes) {
+				const lane = experimentRunLanes.lanes.find(x => x.laneNumber === laneNumber)
+				if (lane) {
+					lane.validationTime = validationTime
+				}
+			}
+			break 
+		}
+
```
nafiz1001, line 71: My condolence.

nafiz1001: Thankfully it's using immer.

nafiz1001 (approved):
After my changes, I approve.

frontend/src/components/experimentRuns/ExperimentRunValidation.tsx
```diff
@@ -59,53 +59,36 @@ function ExperimentRunValidation({ experimentRunName }: ExperimentRunValidationP
 
 	}, [experimentRunName, experimentRunLanesState])
 
+	const updateLane = useCallback((lane: LaneInfo) => {
+		Promise.allSettled(lane.datasets.map((dataset) => dispatch(get(dataset.datasetID)))).finally(() => {
```
UlysseFG, line 63: Any clue why this is working without await ? I spent a good while without it having the promise returning right away ... From looking at the console you need the await otherwise you get spammed with calls to the backend ... it ends up falling back on its feet but i think it is a bit random.

UlysseFG: > From looking at the console you need the await otherwise you get spammed with calls to the backend ... it ends up falling back on its feet but i think it is a bit random.
Apparently it is not the await ... some useEffect within the collapsible gets all excited once we start refreshing the store...

nafiz1001: > Any clue why this is working without await ? I spent a good while without it having the promise returning right away ... From looking at the console you need the await otherwise you get spammed with calls to the backend ... it ends up falling back on its feet but i think it is a bit random.

It might be because you used curly braces which returns `void` by default unless you explicitly return a `Promise` value.

nafiz1001: `async (dataset) => {dispatch(get(dataset.datasetID))}`

Without an the await keyword, it just returns void.

frontend/src/components/experimentRuns/ExperimentRunValidation.tsx
```diff
@@ -162,14 +152,19 @@ function LanePanel({ lane, canValidate, canReset, isValidationInProgress, setPas
   const datasetsById = useAppSelector(selectDatasetsByID)
 	const [datasets, setDatasets] = useState<Dataset[]>([])
 
+  useEffect(() => {
+    const refreshedDatasets = lane.datasets.map((dataset) => datasetsById[dataset.datasetID])
+    setDatasets(refreshedDatasets as Dataset[])
```
nafiz1001, line 157: `useMemo` would be simpler

frontend/src/modules/experimentRunLanes/actions.ts
```diff
@@ -47,12 +47,12 @@ export function setRunLaneValidationStatus(lane: LaneInfo, status: ValidationSta
 }
 
 export function setRunLaneValidationTime(lane: LaneInfo) {
-	return async (dispatch: AppDispatch, getState: () => RootState) => {
-    const latestValidationTime = lane.datasets.reduce<string | undefined>((latest, dataset) => {
+  return async (dispatch: AppDispatch, getState: () => RootState) => {
+    const validationTime = lane.datasets.reduce<string | undefined>((latest, dataset) => {
```
nafiz1001, line 51: why rename `latestValidationTime` to `validationTime`.

UlysseFG: LatestValidationTime gives the impression there is more than one ... this is the only value available ... multiple time. Keeping it simple is part of making it clear in that case.

frontend/src/components/experimentRuns/ExperimentRunValidation.tsx
```diff
@@ -162,14 +152,19 @@ function LanePanel({ lane, canValidate, canReset, isValidationInProgress, setPas
...
+	}, [datasetsById])
+
```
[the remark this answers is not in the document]
UlysseFG: Maybe ... Not sure how as far as i can see it would me similar code with an extra wrapper ...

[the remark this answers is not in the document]
nafiz1001: It's slightly better because you don't need to wait for the next rerender to get the refreshed datasets.

nafiz1001 (approved):
Tested

PR #923, opened 2025-02-05
No partial release validation.

UlysseFG:


backend/fms_core/viewsets/dataset.py
```diff
@@ -84,6 +84,14 @@ def set_release_status(self, request, pk):
             transaction.set_rollback(True)
             return HttpResponseServerError(f"Error updating release status: {e}")
 
+        # Validate that all release status are set (released or blocked) at once.
+        readsets = list(Readset.objects.filter(dataset=pk).all())
+        readset_count = len(readsets)
+        unset_count = len(list(filter(lambda x : x.release_status==ReleaseStatus.AVAILABLE, readsets)))
```
nafiz1001, line 90: Prefer list comprehension over `filter` + `lambda`

nafiz1001: It even be more efficient to just use a `for` loop ;)

nafiz1001 (approved):

PR #924, opened 2025-02-05
Improve samples load time.

UlysseFG:


frontend/src/modules/libraries/actions.ts
```diff
@@ -15,10 +16,10 @@ export const get = id => async (dispatch, getState) => {
     return await dispatch(networkAction(GET, api.libraries.get(id), { meta: { id } }));
 };
 
-export const list = (options) => async (dispatch, getState) => {
+export const list = (options, abort = false) => async (dispatch: AppDispatch, getState) => {
```
nafiz1001, line 19: 👍🏼

backend/fms_core/viewsets/library.py
```diff
@@ -29,57 +29,7 @@
...
-    )
-    queryset = queryset.annotate(
-        first_derived_sample=Subquery(
-            DerivedBySample.objects
-            .filter(sample=OuterRef("pk"))
-            .values_list("derived_sample", flat=True)[:1]
-        )
-    )
-
+    queryset = Sample.objects.all()
```
nafiz1001, line 32: It would make sense to at least filter queryset so that you're left with libraries.

backend/fms_core/viewsets/_fetch_data.py
```diff
@@ -105,19 +101,13 @@ def fetch_data(self, ids: List[int] =[]) -> Tuple[List, int]:
...
             'updated_by',
             'updated_at',
             'deleted',
-            'count_derived_by_sample',
         )
 
-        count = self.queryset.count() # Get count after value to have rows merged but before paging to have complete count
-
         if self.fetch_limit is not None and self.fetch_offset is not None:
             self.queryset = self.queryset[self.fetch_offset:self.fetch_offset+self.fetch_limit] # page the queryset
```
nafiz1001, line 112: Why not split this into two if statement?

```python
self.queryset = self.queryset[self.fetch_offset:]
self.queryset = self.queryset[:self.fetch_limit]
```

nafiz1001 (approved):
I looked at the code and played around. So far I didn't come across any bugs.

backend/fms_core/viewsets/_fetch_data.py
UlysseFG: This is not equivalent...

nafiz1001: I know it's not equivalent. But, maybe the user would like to paginate with only either offset or limit?

UlysseFG: This is a old bit of code unrelated to the current issue ... we should talk this out and create an independent ticket for that. I think what you said makes sense but i do not know why we did it that way in the first place ...

nafiz1001:
Oops I approved twice.

PR #930, opened 2025-02-20
Separate serializing of range and date range filters.

UlysseFG:


UlysseFG:
> I wonder if combining them was intentional choice...

Yeah i think so they used to behave the same way by taking the default value returned by the filter component.

PR #933, opened 2025-02-26
New sample submission template.

UlysseFG:
- [x] Concentrations to be mandatory => only on the user side
- [x] Comments to be added above the column instead of having to hover over the cell => Maybe, i will see how this looks
- [x] Show error message when special characters are added or mandatory information is missing => Addition of red conditional background when bad characters are used. Don't think we can check for mandatory information without macro unless everything is red by default...
- [x] Not sure if “Tissue source” is needed and should be mandatory for nucleic acid => Make user empty cell be reported as TISSUE.
- [x] Location kind and container coordinates should be greyed out for 96-well plates => Adding conditional background.
- [x] Location kind and container coordinates should be mandatory for tubes. Not sure what “site” is. => Adding mandatory to those column for when they are not greyed out. Moving site to greyed out.
- [x] PoolSubmission example tab maybe should have a more detailed example ? => Clean up exemples.
- [x] Add ParseBio and Dovetail new indices.


nafiz1001 (approved):
I ran test and it passed :)

PR #934, opened 2025-02-26
Add "Start with" and "Exact" filter options

nafiz1001:


frontend/src/components/pagedItemsTable/serializeFilterParamsTS.ts
```diff
@@ -4,128 +4,147 @@
...
+    SortBy,
 } from '../../models/paged_items'
 import { QueryParams } from '../../utils/api'
 import dayjs from 'dayjs'
 
-export default function serializeFilterParamsWithDescriptions(filters: FilterSet) {
-  const params: NonNullable<QueryParams> = {}
-
-  function hasSpaces(string) {
+function hasSpaces(string: string) {
```
UlysseFG, line 21: There was a doubt about the type there ? ;)

nafiz1001: Typescript doubted it ;)

frontend/src/components/pagedItemsTable/serializeFilterParamsTS.ts
```diff
@@ -4,128 +4,147 @@
...
+    return params
+}
+
+export default function serializeFilterParamsWithDescriptions(filters: FilterSet) {
+    const params: NonNullable<QueryParams> = {}
+
+    Object.keys(filters).forEach((key) => {
+        const setting = filters[key]
+        if (setting.value && setting.description) {
+            const paramsForFilterKey = paramsForFilterKeyAndSetting(key, setting.value, setting.description, setting.options)
```
UlysseFG, line 129: Not sure why you made a function out of this ... will it ever be called anywhere but here ? Just forces us to scroll around the file to find the rest of the code ...

nafiz1001: I factored it out in case we want to obtain query parameters for one particular column filter-options. It's a left-over relic from my work in another branch.

UlysseFG (approved):

nafiz1001:
@UlysseFG I decided to use this branch to fix the query param filter on WorkflowAssignment. Now it should do exact filter, but only for simple text-based filters.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -29,33 +30,60 @@ export function WorkflowAssignment({ fixedFilter }: LabworkSamplesProps) {
     const wholeFilters = useMemo(() => ({ ...filters, ...fixedFilters }), [filters, fixedFilters])
 
     const samplesTableCallbacks = usePagedItemsActionsCallbacks(SamplesTableActions)
+    const [searchParams] = useSearchParams()
     useEffect(() => {
+        samplesTableCallbacks.clearFixedFiltersCallback()
```
UlysseFG, line 35: isn't the return callback dealing with that already ? or this is to prevent cross-effect from regular sample table fixed filters ?

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -101,8 +129,19 @@ export function WorkflowAssignment({ fixedFilter }: LabworkSamplesProps) {
         setOpen(false)
     }, [])
 
+    console.info(fixedFilters)
```
UlysseFG, line 132: !

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -101,8 +129,19 @@ export function WorkflowAssignment({ fixedFilter }: LabworkSamplesProps) {
...
     return (
         <>
+            {Object.keys(wholeFilters).length > 0 && (
+                <>
+                    <b>Fixed Filters:</b>
+                    <ul>
+                        {Object.values(fixedFilters).map((setting, index) => (
+                            <li key={setting.description?.key ?? index}>{`${setting.description?.label}: ${setting.value}`}</li>
```
UlysseFG, line 140: Not sure about the retro looking list of fixed filters... The filter is listed in the table header but cannot be cleared. if you remove it from the column the exact is stuck and the filter does not accept other values ... I think fixed filter were meant for components that cannot take other values for a given column ... (Exemple project column for ProjectAssociatedSamples). In our case one could decide to clear the filter and input something else ... It if more of a default filter than a fixed one.

frontend/src/components/management/WorkflowAssignment.tsx
nafiz1001: It's to prevent cross-effect from regular sample table fixed filters.

frontend/src/components/management/WorkflowAssignment.tsx
nafiz1001: Hmm... maybe I should hide the column instead?

UlysseFG:
When trying to filter recursively the location, it also automatically triggers the Starts_with ... it should instead set the exact. When looking for everything within a container you only want one hit. (it was starts_with before because there was no exact ...)

PR #939, opened 2025-03-03
Fix timeline spacing

nafiz1001:


nafiz1001:
This PR also affects the Versions timeline, let me know what you think.

UlysseFG (approved):
Looks good as far as i can see ...

PR #940, opened 2025-03-04
Multiple comments with scrollbar

nafiz1001:


UlysseFG:
It is less confusing and quicker to review than before but the scrolling comment within a scrolling collapsible is not ideal. For exemple, once you reach the bottom of the comments it scroll down the page instead which is annoying.

UlysseFG (approved):
It is better than before.

PR #947, opened 2025-03-13
Always allow Get Samplesheet

nafiz1001:


frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -42,15 +42,6 @@ function Placement({ stepID, sampleIDs }: PlacementProps) {
       await dispatch(fetchSamplesheet(activeDestinationContainer, cells))
     }, [dispatch, activeDestinationContainer, cells])
 
-    const isPlacementComplete = useMemo(() => {
```
UlysseFG, line 45: Don't get rid of that. We'll reactivate once we have the placement upgraded ...

frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -187,7 +178,7 @@ function Placement({ stepID, sampleIDs }: PlacementProps) {
                             <AddPlacementContainer onConfirm={onConfirmAddDestinationContainer} existingContainers={loadedContainers} />
                         </Col>
                         {usesSamplesheet && <Col span={3} >
-                            <Button onClick={handleGetSamplesheet} disabled={!isPlacementComplete}>Get Samplesheet</Button>
+                            <Button onClick={handleGetSamplesheet}>Get Samplesheet</Button>
```
UlysseFG, line 181: Just comment out the disabled... This is a temporary thing. We should put it back once it no longer cause issues ...

PR #951, opened 2025-03-21
Fix invalid index message

UlysseFG:


nafiz1001 (approved):
The changes appear to have fixed the problem.

PR #962, opened 2025-04-02
Validation before processing during template ingestion.

UlysseFG:


backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -20,6 +20,16 @@ class SampleRowHandler(GenericRowHandler):
     def __init__(self):
         super().__init__()
 
+    def validate_row_input(self, **kwargs):
+        super().validate_row_input(**kwargs)
+
+        sample = kwargs.get("sample", None)
```
nafiz1001, line 26: Why not assume that the expected parameters are present? `sample, library, container, project, parent_container, individual, sample_kind_objects_by_name, defined_pools`
```
    def process_row_inner(self, sample, library, container, project, parent_container, individual, sample_kind_objects_by_name, defined_pools):
```

UlysseFG: The validate row input signature would be different for a different template row ...

nafiz1001 (approved):

PR #963, opened 2025-04-03
Link Dataset to Project using FK instead of name mapping.

UlysseFG:


backend/fms_core/migrations/0069_v5_0_0.py
```diff
@@ -0,0 +1,76 @@
...
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+
+        reversion.set_comment("Insert fk of FMS project for each existing Dataset before removing project related fields.")
+        reversion.set_user(admin_user)
+
+        for dataset in Dataset.objects.all():
+            try:
+                project = Project.objects.get(name=dataset.project_name)
+            except:
+                project = Project.objects.get(name__startswith=dataset.project_name)
```
nafiz1001, line 24: Filter with `name__startswith` seems kinda risky.

UlysseFG: A bit in the sense it might not work but if we don't get exactly one match it is going to fail. The issue here is that someone had the great idea to rename a project, the good news is the renaming was done by appending stuff ... This is mainly a single project causing issue to the migration. I will have to run the migration on a prod backup the day before the release to make sure all still runs ...

backend/fms_core/schema_validators.py
```diff
@@ -107,20 +107,21 @@ def deconstruct(self):
...
                     "type": "object",
                     "properties": {
+                        "project_obj_id": {"type": "string"},
                         "external_project_id": {"type": "string"},
                         "project_name": {"type": "string"},
                         "sample_name": {"type": "string"},
                         "derived_sample_id": {"type": ["null", "number"]},
                     },
-                    "required": ["sample_name", "external_project_id", "project_name"]
+                    "required": ["project_obj_id", "sample_name", "external_project_id", "project_name"]
```
nafiz1001, line 124: I guess we are still handling `external_project_id`?

UlysseFG: It is still the unifying ID for project in the center ... if the json ever get ingested elsewhere it would still be useful ... maybe i could remove it as a requirement but it might be better to keep the pressure for compliance on that side.

nafiz1001:
Would be nice if there was a hyperlink on the project name.
image

UlysseFG:
> Would be nice if there was a hyperlink on the project name, and maybe even on the run name.

Agreed you can make a ticket for that should be pretty easy to do ...

nafiz1001 (approved):
Run validation report ingestion finally works.

PR #967, opened 2025-04-23
Resize the timelines left margin for larger labels.

UlysseFG:


nafiz1001 (approved):
I checked out sample timeline, dataset timeline and user timeline and they all looked ok or better.

PR #968, opened 2025-04-24
Placement Refactor Attempt #2

nafiz1001:


frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -295,23 +293,17 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
...
 
 	useEffect(() => {
 		if (stepSamples.selectedSamples.items.length === 0) {
-			// ensures there are no left over samples from containers that have no selection
-			// because of the abrupt disable of placement tab
-			const sourceContainers = selectSourceContainers(store.getState()).map((c) => c.name)
-			dispatch(flushPlacementContainers(sourceContainers))
-			dispatch(flushLabworkStepPlacementContainers(sourceContainers))
-
 			onTabChange(GROUPED_SAMPLES_TAB_KEY)
```
UlysseFG, line 304: This will trigger twice onTabChange(GROUPED_SAMPLES_TAB_KEY) on clearSelectedSamples. We should remove one of the instances. The one in clearSelectedSamples is likely the easiest to remove as it will consistantly trigger the one here.

frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -144,37 +135,12 @@ function Placement({ stepID, sampleIDs }: PlacementProps) {
...
-                }))
-            }
-        })
-    }, [activeSourceContainer?.name, destinationContainers, dispatch, sourceContainers])
-
-    const removeSelectedCells = useCallback(() => {
+    const undoPlacementsCallback = useCallback(() => {
         if (activeDestinationContainer) {
-            dispatch(undoSelectedSamples(activeDestinationContainer.name))
+            dispatch(undoPlacements(activeDestinationContainer.name))
```
UlysseFG, line 143: We pass the container for the placeAllSource we should do the same for the undoPlacements.

frontend/src/components/placementVisuals/Placement.tsx
```diff
@@ -188,7 +154,7 @@ function Placement({ stepID, sampleIDs }: PlacementProps) {
                         </Col>
                         {usesSamplesheet && <Col span={3} >
                             <Button
-                            onClick={handleGetSamplesheet}
+                                onClick={handleGetSamplesheet}
                             // disabled={!isPlacementComplete}
```
UlysseFG, line 158: With PacBio coming with smrt cells that can be empty i don't this this line is making a come back. You can likely remove the commented line for good.

frontend/src/components/placementVisuals/Placement.tsx
nafiz1001: Huh? But I am passing the container `activeDestinationContainer.name`

UlysseFG: It is just a question of consistency :) Up to you.

nafiz1001: Where's the inconsistency? 🤔

frontend/src/modules/placement/classes.ts
```diff
@@ -0,0 +1,874 @@
...
+            if (cell.getSamples(true).length > 0) {
+                this.placementState.error ??= `Sample ${cell.getSamples(true)[0].name} already exists in cell ${cell.coordinates}`
+            }
+        }
+    }
+
+    compareSamples(a: SamplePlacementIdentifier, b: SamplePlacementIdentifier) {
+        const MAX = 128
+
+        const A = this.#resolveSamplePlacementIdentifier(a)
```
UlysseFG, line 325: This here appear very inefficient ... you are allocating objects in a sorting loop this can likely be done better with passing the objects from outside the sorting call. A single item will be reallocated multiple time here. You also incur the risk of modifying your item between calls.

frontend/src/modules/placement/classes.ts
```diff
@@ -0,0 +1,874 @@
...
+                this.unSelectSample(sampleID)
+            }
+        } else {
+            for (const sampleID of sampleIDs) {
+                this.selectSample(sampleID)
+            }
+        }
+    }
+
+    compareSamples(a: Pick<SamplePlacementIdentifier, 'sample'>, b: Pick<SamplePlacementIdentifier, 'sample'>) {
```
UlysseFG, line 542: I am not convinced you need more than one definition of this compareSample algorithm... There is non negligible risk here you will change one and forget the other ones ... the logic is the same as far as i can see... If you want to use the object inheritance you have here you could have your base class method do the basic sorting logic and have the derived classes prepare the data to call it properly using a super call.

frontend/src/components/placementVisuals/PlacementSamplesTable.tsx
```diff
@@ -21,113 +19,147 @@ interface PlacementSample {
...
+        } else {
+            if (!sample.coordinates) return
+            dispatch(multiSelect({
+                forcedSelectedValue: selected,
+                context: {
+                    source: activeSourceContainer
+                },
+                type: 'samples-placements',
+                parentContainer: parentContainerID,
+                samples: [{ sample, cell: { fromContainer: parentContainerID, coordinates: sample.coordinates } }],
```
UlysseFG, line 146: Exposing a different format for samples here is risky and forcing you to track and test it all over the place ... you should have a uniform format (at least list sample in an object so you do not need to test it all the time) and test it within your placement objects as much as possible.

frontend/src/modules/placement/classes.ts
UlysseFG: Ignore the object inheritance part here ... Would be too much of a rework.

frontend/src/modules/placement/classes.ts
```diff
@@ -0,0 +1,874 @@
...
+    enter() {
+        if (this.fromContainer.sameContainerAs(this.context.sourceContainer)) {
+            return
+        }
+        const samples: SampleClass[] = []
+        if (this.context.sourceContainer instanceof TubesWithoutParentClass) {
+            samples.push(...this.context.sourceContainer.getSortedSamples(true))
+        } else {
+            // source containers should have no new placements
+            samples.push(...this.context.sourceContainer.getSortedPlacements(true).map((s => s.sample)))
```
UlysseFG, line 666: same getSortedSamples should exist on both class.

UlysseFG:
When you place samples you should not display a count before it reaches 2 ... 1s is kind of the blank default. You only need to see those once they reach 2.

UlysseFG (approved):

PR #969, opened 2025-04-24
Labwork manage workflow

nafiz1001:


frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -354,6 +354,11 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
 					/>
 				</>
 			}
+			<a href={`/management/workflow-assignment?${SampleColumnID.QUEUED_STEPS}=${step.name}`}>
```
UlysseFG, line 357: We are using <Link> nearly everywhere ... this would be good to stick to the same.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -83,7 +83,8 @@ export function WorkflowAssignment(props: LabworkSamplesProps) {
                 render: (_, { sample }) => {
                     if (!sample) return null
                     const sampleNextSteps = sampleNextStepsBySampleID[sample.id]
-                    if (!sampleNextSteps || sampleNextSteps.length === 0) return null
+                    if (!sampleNextSteps) return <Spin size={"small"} />
+                    if (sampleNextSteps.length === 0) return null
```
UlysseFG, line 87: Is this case actually happening ? sampleNextStepsBySampleID is likely to never return an empty list since it is built from queued samples ...

nafiz1001: Oh I see. It would return `undefined` if there is no sample next step.

UlysseFG (approved):

PR #971, opened 2025-04-28
Infinium cleanup

UlysseFG:


backend/fms_core/viewsets/_utils.py
```diff
@@ -378,8 +378,15 @@ def default_batch_prefilling(sample_id_list: List[int], template, batch_sheet_na
 
             return batch_rows_list
 
+        def get_step_specific_experiment_prefix(step_name: str):
+            match step_name:
+                case "Experiment Run Infinium":
+                    return ""
+                case _:
+                    return "CHANGE ME "
```
nafiz1001, line 386: Oh my, a match expression where you would normally use an if-else statement 🤭.

UlysseFG: Leaving some space for other experiment run to have their own pattern.

backend/fms_core/tests/test_template_importers/experiment_run_test.py
nafiz1001, line 1: In the future, we should cleanup the way variables are named here, or maybe use some elegant technique for testing bunch of values. One strategy I can think of is to use `dict` to track values instead of variables like `cp1_4`, `cp1_4_p1`, etc.

nafiz1001 (approved):

backend/fms_core/tests/test_template_importers/experiment_run_test.py
UlysseFG: This could be prettier.

PR #976, opened 2025-05-05
Fix Sequencer side value refused for MiSeq

UlysseFG:


nafiz1001:
The dropdown for "Sequencer Side" column ends a lot sooner than the other columns like "Experiment Container Kind".

UlysseFG:
> The dropdown for "Sequencer Side" column ends a lot sooner than the other columns like "Experiment Container Kind".

You mean the dropdown does not apply to as many rows ? Ok i'll check that.

nafiz1001 (approved):
Looks soooooo good to me.

PR #977, opened 2025-05-05
Index validation change to allow mismatched indices.

UlysseFG:


backend/fms_core/viewsets/index.py
```diff
@@ -89,7 +90,7 @@ def validate(self, _request):
             form_errors["length_3prime"].append(f"Validation length for index at 3 prime end cannot be negative.")
 
         threshold = _request.GET.get("threshold", "").strip()
-        threshold = int(threshold) if threshold else None
+        threshold = int(threshold) if blank_str_to_none(threshold) is not None else None
```
nafiz1001, line 93: Maybe you can do something like
```suggestion
        threshold = int(threshold) if threshold else None
```

PR #983, opened 2025-05-13
Options to From/To Step

nafiz1001:


frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -481,13 +481,13 @@ function WorkflowOptions({ defaultSelection, exceptedSampleIDs, filters }: Labwo
             <Collapse key={'actions'} accordion items={[
                 {
                     key: 'dequeue',
-                    label: `Dequeue Options (${projectDequeueActions.length})`,
+                    label: `(${projectDequeueActions.length}) Dequeue From Step`,
```
UlysseFG, line 484: I am not sure i like the counter reversed like that ... This could lead to weird text alignment if the counter for the queue and dequeue are much different in length (100 vs 1) i would prefer for the text to be aligned on the left with the counter on the right like before.

nafiz1001: My concern is that it might confuse the user if the counter appears after. But if it's not confusing to you, chances are it's fine for the user anyway.

UlysseFG: Well it certainly is not clearer to me if it is before ;)

UlysseFG (approved):

PR #984, opened 2025-05-21
New STOOL sample kind.

UlysseFG:


nafiz1001 (commented):
Maybe should be called feces ;)

nafiz1001 (approved):
💩

PR #989, opened 2025-05-28
Fix trailing zero handling in RangeFilterComponent

nafiz1001:


UlysseFG:
Pressing the reset button no longer clears the values displayed in the filter. Not sure this is a good tradeoff ...

nafiz1001:
> Pressing the reset button no longer clears the values displayed in the filter. Not sure this is a good tradeoff ...

Should be resolved in the new commit.

PR #990, opened 2025-05-28
Don't show alias in sample workflow management

nafiz1001:


UlysseFG:
Mmm... The PR does what the ticket asked, but the aim of this ticket was to try to flatten the rows which is not attained by these changes... Not sure these changes are worth it without additional work on the other columns. Project name lengths are too long; we should apply ellipse for those and the Queued Steps column is too small for the content.

PR #993, opened 2025-06-02
Sample Identity QC step

UlysseFG:
- [x] Create a template for the new step.
- [x] Create the new step and protocol using a migration.
- [x] Add the new step and protocol to new workflows that mirror the existing one with the extra step.
- [x] Create the backend viewset changes and Importer code to ingest the template.
- [x] Add tests for new model, services and importer.


backend/fms_core/template_importer/row_handlers/sample_identity_qc/sample.py
```diff
@@ -0,0 +1,61 @@
...
+            if sample_obj.is_pool:
+                self.errors['sample'] = f"Sample identity QC can't be performed on a pool of libraries."
+
+            # Update sample with sample_information
+            new_volume = None
+            volume_used = process_measurement.get('volume_used', None)
+            if volume_used is not None:
+                new_volume = sample_obj.volume - volume_used
+            else:
+                self.errors['volume'] = 'Volume Used is required.'
```
nafiz1001, line 25: Oh great, so `errors` here is just a string rather than array of string. 😪

nafiz1001 (approved):
Other than the comments I have written, I approve.

PR #998, opened 2025-06-13
 Add library ratio in sample submission template for pools and default to equal ratio.

nafiz1001:


nafiz1001:
I have not thoroughly tested but code-wise it should be ready.

backend/fms_core/services/sample.py
```diff
@@ -602,12 +601,29 @@ def pool_submitted_samples(samples_info,
         errors.append(f"Reception date is not valid.")
 
     if not errors:
-        # Calculate the total volume of the pool
-        pool_volume = sum(sample["volume"] for sample in samples_info)
+        equal_volume_ratio = True
```
UlysseFG, line 604: Don't do this here ... calculate the ratio outside and pass the actual ratio in ... Keep the business logic outside the services as much as possible ...

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
```diff
@@ -27,8 +27,8 @@ def validate_row_input(self, **kwargs):
         # make sure all required values are present. WIP.
         if sample["sample_kind"] is None:
             self.errors['sample_kind'] = [f"Sample Kind is a required field."]
-        if sample["volume"] is None:
-            self.errors['volume'] = [f"Volume (uL) is a required field."]
+        if sample["volume"] is None and sample["volume_ratio"] is None:
+            self.errors['volume'] = [f"'Either Volume (uL)' or 'Ratio Library In Pool' is required field."]
```
UlysseFG, line 31: You should be able to suggest which one is required with the template content (you can pull the sample type for this purpose if you need).

backend/fms_core/tests/test_services/sample_test.py
```diff
@@ -135,9 +135,10 @@ def setUp(self) -> None:
...
-            {"alias": "Alias2", "individual": self.test_individuals[1], "collection_site": "TestSite", "sample_kind": self.sample_kind_dna, "tissue_source": self.sample_kind_blood, "library": self.test_libraries[5], "project": None, "studies": [], "volume": Decimal(30), "experimental_group": None, "fragment_size": 100},
+            {"alias": "Alias1", "individual": self.test_individuals[0], "collection_site": "TestSite", "sample_kind": self.sample_kind_dna, "tissue_source": self.sample_kind_blood, "library": self.test_libraries[4], "project": None, "studies": [], "volume_ratio": Decimal(0.25), "experimental_group": None, "fragment_size": 100},
+            {"alias": "Alias2", "individual": self.test_individuals[1], "collection_site": "TestSite", "sample_kind": self.sample_kind_dna, "tissue_source": self.sample_kind_blood, "library": self.test_libraries[5], "project": None, "studies": [], "volume_ratio": Decimal(0.75), "experimental_group": None, "fragment_size": 100},
```
UlysseFG, line 139: Might be a good idea to add another service test for sum of ratios not equal to 1.

backend/fms_core/services/sample.py
nafiz1001: Isn't the service layer for business logic? Where else would the business logic take place? In the controller layer? model layer?

backend/fms_core/template_importer/row_handlers/sample_submission/sample.py
nafiz1001: DONE

backend/fms_core/services/sample.py
```diff
@@ -602,12 +601,29 @@ def pool_submitted_samples(samples_info,
...
+        for sample in samples_info:
+            if sample.get("volume_ratio", None) is not None:
+                equal_volume_ratio = False
+
         for sample in samples_info:
-            sample_volume = sample["volume"]
-            # Calculate the volume ratio of each sample in the pool
-            sample["volume_ratio"] = decimal_rounded_to_precision(sample_volume / pool_volume, 15)
+            volume_ratio = sample.get("volume_ratio", None)
+            if not equal_volume_ratio and volume_ratio is None:
```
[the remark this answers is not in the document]
UlysseFG: This comment is to be understood paired with the other one that request you take the ratio calculation out of the service. If you call the service without providing a ratio this should be an error. You have to calculate the equal ratio before calling the service. Do you actually mean prefill or "prefill" ? We will not be prefilling equal ratio since this is much more processing that we currently do for prefilling... There is not much need for it too. If we do something in that direction it's likely going to be a formula in excel...

backend/fms_core/tests/test_services/sample_test.py
nafiz1001: DONE

backend/fms.log.1
```diff
@@ -0,0 +1 @@
+[2025-07-09 13:02:54,705] [DEBUG] (0.004) COMMIT; args=None; alias=default
```
UlysseFG, line 1: Not sure we need logs to be part of the repo ...

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
UlysseFG, line 1: Try to keep your config local...

backend/fms_core/services/sample.py
```diff
@@ -546,9 +546,49 @@ def pool_samples(process: Process,
...
+        if sample.get("volume_ratio", None) is not None:
+            equal_volume_ratio = False
+
+    for sample in samples:
+        if not equal_volume_ratio and sample.get("volume_ratio", None) is None:
+            errors.append(f"Volume ratio for sample {sample['alias']} is missing but all samples must either define volume ratio or not simultaneously.")
+    if errors:
+        return errors
+
+    exp = Decimal(f'1E-{DerivedBySample.VOLUME_RATIO_DECIMAL_PLACES}')
```
UlysseFG, line 577: nice.

backend/fms.log.1
nafiz1001: REMOVED

.vscode/settings.json
nafiz1001: REMOVED

backend/fms_core/services/sample.py
```diff
@@ -546,9 +546,49 @@ def pool_samples(process: Process,
 
     return sample_destination, errors, warnings
 
+def test_and_update_volume_ratio(samples: list[dict]):
+    """
+    Tests the volume ratio of the samples in the pool and updates them if necessary.
+    If all samples have volume_ratio defined, it checks that they are all defined and equal.
+    If all samples do not have volume_ratio defined, it sets the volume_ratio to 1/n for each sample.
```
UlysseFG, line 553: if none of the samples have the volume_ratio defined, ...

backend/fms_core/services/sample.py
```diff
@@ -546,9 +546,49 @@ def pool_samples(process: Process,
...
+    if errors:
+        return errors
+
+    exp = Decimal(f'1E-{DerivedBySample.VOLUME_RATIO_DECIMAL_PLACES}')
+    if equal_volume_ratio:
+        for sample in samples:
+            sample["volume_ratio"] = (Decimal(1) / Decimal(len(samples))).quantize(exp)
+    else:
+        # Verify that the ratio add up to 1
+        total_volume_ratio = Decimal(sum(Decimal(sample["volume_ratio"]).quantize(exp) for sample in samples)).quantize(exp)
```
UlysseFG, line 583: You loop around samples a lot ... try to store the information required here during the initial loop. If there is missing ratios, the value is n * 1/n ... Also by doing the rounding so each sample ratio you increase your risk of having rounding errors i think. For example, you have 3 sample without ratio you allocate the value 1/3 to them the rounding will add 0.333...3 three times which will not add up to 1 at the 15 decimal precision, but if you add 1/3 at the full precision 3 times then round them up to the 15th decimal they will round up to 1.

backend/fms_core/services/sample.py
```diff
@@ -602,12 +640,7 @@ def pool_submitted_samples(samples_info,
         errors.append(f"Reception date is not valid.")
 
     if not errors:
-        # Calculate the total volume of the pool
-        pool_volume = sum(sample["volume"] for sample in samples_info)
-        for sample in samples_info:
-            sample_volume = sample["volume"]
-            # Calculate the volume ratio of each sample in the pool
-            sample["volume_ratio"] = decimal_rounded_to_precision(sample_volume / pool_volume, 15)
+        errors.extend(test_and_update_volume_ratio(samples_info))
```
UlysseFG, line 643: Don't use your non service function within a service too. This goes in the importer (or row handler)...

backend/fms_core/tests/test_services/sample_test.py
```diff
@@ -731,4 +734,47 @@ def test_pool_submitted_samples(self):
             else:
                 self.assertEqual(derived_sample.biosample.individual, self.test_individuals[1])
                 self.assertEqual(derived_by_sample_pool_.volume_ratio, Decimal("0.75"))
-                self.assertEqual(derived_sample.library, self.test_libraries[5])
\ No newline at end of file
+                self.assertEqual(derived_sample.library, self.test_libraries[5])
+
+    def test_test_and_update_volume_ratio(self):
+        samples: list[dict[str, Any]] = [{ "alias": "A" }, { "alias": "B" }, { "alias": "C" }, { "alias": "D" }]
```
UlysseFG, line 740: Test using a number of sample that do not split exactly to see how it handles the rounding error issue.

backend/fms_core/template_importer/row_handlers/sample_submission/pool.py
```diff
@@ -80,9 +82,35 @@ def process_row_inner(self, samples_info, pool, seq_instrument_type, reception_d
...
+                volume_ratio = sample.get("volume_ratio", None)
+                if not equal_volume_ratio:
+                    if volume_ratio is None:
+                        self.errors["volume_ratio"] = [f"All samples in the pool {pool['name']} must either define volume ratio or not at the same time."]
+                    else:
+                        total_volume_ratio += Decimal(volume_ratio)
+                else:
+                    self.warnings["volume_ratio"] = [f"Each sample in the pool {pool['name']} will have a volume ratio of {Decimal(1/len(samples_info)).quantize(exp)}."]
+                    total_volume_ratio = 1
+                    if volume_ratio:
```
UlysseFG, line 102: you are testing a condition that cannot happen unless you made a mistake in your code... This test and error should never be called... better remove them.

backend/fms_core/template_importer/row_handlers/sample_submission/pool.py
```diff
@@ -80,9 +82,35 @@ def process_row_inner(self, samples_info, pool, seq_instrument_type, reception_d
...
+                        self.errors["volume_ratio"] = [f"All samples in the pool {pool['name']} must either define volume ratio or not at the same time."]
+                    else:
+                        sample["volume_ratio"] = (Decimal(1) / Decimal(len(samples_info))).quantize(exp)
```
UlysseFG, line 105: move calculation out of the loop.

backend/fms_core/template_importer/row_handlers/sample_submission/pool.py
```diff
@@ -80,9 +82,45 @@ def process_row_inner(self, samples_info, pool, seq_instrument_type, reception_d
...
+                        total_volume_ratio += Decimal(volume_ratio).quantize(EXP)
+
+            if not self.errors.get("volume_ratio") and total_volume_ratio != 1:
+                if equal_volume_ratio:
+                    # correct by adjusting first volume ratio
+                    delta = Decimal(1).quantize(EXP) - total_volume_ratio
+                    samples_info[0]["volume_ratio"] += delta
+                    total_volume_ratio += delta
+                else:
+                    self.errors["volume_ratio"].append(f"Total volume ratio of the samples in the pool {pool['name']} must add up to exactly 1. ({float(total_volume_ratio)})")
```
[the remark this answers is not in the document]
nafiz1001: I think I casted to float because it looked prettier without any extra zeros.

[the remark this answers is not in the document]
UlysseFG: Yeah pretty is good. I am worried someone could take that value then use it to calculate the missing amount and find out it is off and not understand why it is not adding up to 1. Try using .normalize() instead ?

UlysseFG:
There is still some dropdown broken in the template. The Selection Target dropdown fails for SampleSubmission and UserSampleSubmission tabs. Index dropdown fails for SampleSubmission tab. When failing to submit all ratios, we get spammed with "This field cannot be null" messages that do not help identify the issue. You can possibly use the derived_by_sample model clean Add_error() function  to provide more information to the user.

UlysseFG:
Here are the latest issues ... "Sample Coord" column for tabs UserSampleSubmission and SampleSubmission have their Data Validation (drop down) broken by the cell shifts. The header column is not high enough the names that use 2 lines are clipped at the top. The column "Ratio Library In Pool" should not display 15 decimals all the time. The number general format is fine. The version of the template should be changed to the current version (both in the file name and inside the file) since there is an actual change in headers and general format.

nafiz1001:
> Here are the latest issues ... "Sample Coord" column for tabs UserSampleSubmission and SampleSubmission have their Data Validation (drop down) broken by the cell shifts. The header column is not high enough the names that use 2 lines are clipped at the top. The column "Ratio Library In Pool" should not display 15 decimals all the time. The number general format is fine. The version of the template should be changed to the current version (both in the file name and inside the file) since there is an actual change in headers and general format.

Not just the "Sample Coord". Even the "Location Coord"

UlysseFG (approved):

PR #1012, opened 2025-08-06
Run tests in parallel to speed them up.

UlysseFG:


nafiz1001 (approved):
with 9 max worker, it works maxfully.

PR #1018, opened 2025-08-12
Carry over selections in Labwork step to Workflow Management

nafiz1001:


frontend/src/components/labwork/step/LabworkStep.tsx
```diff
@@ -74,7 +75,7 @@ const LabworkStep = ({ protocol, step, stepSamples }: LabworkStepPageProps) => {
 				console.error('No templates are associated with step!')
 			}
 		}
-	}, [stepSamples, selectedTemplate])
+	}, [stepSamples, selectedTemplate, isAutomationStep])
```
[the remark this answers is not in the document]
nafiz1001: 😪 Fine.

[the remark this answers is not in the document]
nafiz1001: DONE

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
```diff
@@ -234,8 +235,8 @@ function PagedItemsTable<T extends object>({
 	// avoid dilema selectAll and selectedItems logic
 	useEffect(() => {
 		setDefaultSelection(false)
-		setExceptedItems([])
-		selection?.onSelectionChanged([], false)
+		setExceptedItems(selection?.initialExceptedItems ?? [])
```
UlysseFG, line 238: Not sure about this ... If the user changes the table filter after changing the selection it would basically reselect the initial sample selection. I think this is a dangerous approach the user might not see the change and drop sample they did not intend to drop ... It used to reset the selection. Maybe it should reset the initialExceptedItems ([])? I don't see where the reset can be done.

nafiz1001: Ideally navigation to the WorkflowManagement should only be done through `useNavigateToWorkflowAssignment` which sets initial selections before navigating. The `WorkflowAssignment` component is so far the only component that sets initial selection.

frontend/src/components/management/WorkflowAssignment.tsx
```diff
@@ -125,19 +127,20 @@ export function WorkflowAssignment(props: LabworkSamplesProps) {
                 acc[sample.sample.id] = sample
             }
             return acc
-        }, {} as Record<string, SampleAndLibrary>)
+        }, {} as Record<string, SampleAndLibraryAndIdentity>)
```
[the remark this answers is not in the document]
UlysseFG: Careful not to redefine any ;)

frontend/src/components/pagedItemsTable/PagedItemsTable.tsx
UlysseFG: A quick test verified my fears ... Once you are on the workflow management page, the behaviour difference between preselected samples and no preselection make it awkward ... For exemple, if i select samples on the labwork step then click on manage workflow then filter to have 0 selection i will still have X samples selected from the preselection. I would expect the component to combine the excepted items with the filter to only select the sample that are part of the filterset (or alternately do like it use to - reset selection to an empty list).

UlysseFG: Reseting selection might be the easier way to go the other option requires you to call the backend to count the samples that match both the filters and the preselection.

PR #1021, opened 2025-08-14
Show Identity QC on sample detail content

nafiz1001:


frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -157,6 +164,82 @@ const SampleDetailsContent = () => {
...
+
+  /**
+   * @typedef {import("../../../models/fms_api_models").FMSSampleIdentity} FMSSampleIdentity
+   * @type {[FMSSampleIdentity | undefined, (value: FMSSampleIdentity) => void]}
+   */
+  const sampleIdentityState = useState(undefined);
+  const [sampleIdentity, setSampleIdentity] = sampleIdentityState
+  useEffect(() => {
+    const biosampleId = sample?.biosample_id
+    console.info('sample?.biosample_id', biosampleId)
```
UlysseFG, line 176: debug info ...

backend/fms_core/viewsets/_constants.py
```diff
@@ -114,7 +114,7 @@
 
 _sample_identity_filterset_fields: FiltersetFields = {
     "id": PK_FILTERS,
-    "biosample_id": FK_FILTERS,
+    "biosample__id": FK_FILTERS,
```
UlysseFG, line 117: biosample_id is the name of the fk field biosample__id is the Id on the foreign table ... I am not sure if django does the optimisation itself but join is definitely slower than no join...

frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -157,6 +164,82 @@ const SampleDetailsContent = () => {
...
+    if (biosampleId) {
+      const identities = dispatch(api.sampleIdentity.list({ biosample__id: biosampleId.toString() }))
+      identities.then(({ data }) => {
+        console.info('Sample identities', data)
```
UlysseFG, line 180: more debug ...

PR #1028, opened 2025-09-11
Pacbio library qc template

UlysseFG:


nafiz1001:
> Worth creating a ticket to fix the modal display?

I think a rebase/merge would solve the problem.

PR #1033, opened 2025-09-15
Include a Experiment Container Name in the experiment run templates.

UlysseFG:


nafiz1001:
As you can probably guess, `backend/fms_core/static/submission_templates/Experiment_run_illumina_v5_3_0.xlsx` also has broken "Experiment Container Kind".

PR #1040, opened 2025-09-18
Prefill of transfer template using 'All' for Volume used

UlysseFG:


nafiz1001 (approved):
I didn't run it but it looks like it should work ;)

PR #1046, opened 2025-09-24
Changed the Pacbio container size and made some modifications following latest meeting.

UlysseFG:


backend/fms_core/static/submission_templates/Experiment_run_Pacbio_v5_3_0.xlsx
nafiz1001, line 1: Must be tricky to draw a rectangle on Gimp.

nafiz1001:
Sequencer Position column doesn't seem to have dropdown (I know this is unrelated to the PR).

UlysseFG:
> Sequencer Position column doesn't seem to have dropdown (I know this is unrelated to the PR).

The column uses a formula based on the coordinate to generate the position as used by the lab ... I could add a dropdown too, but i don't really want the lab to input anything here ... i want to be able to trust the coordinate they used more than that value which is there as a helper and to record the location of the sample in a format the lab might be more used to.

PR #1050, opened 2025-09-30
Preference of Justice

nafiz1001:
https://redmine.c3g-app.sd4h.ca/issues/2608

### DIFF
--- a/frontend/src/components/processMeasurements/ProcessMeasurementsListContent.js
+++ b/frontend/src/components/processMeasurements/ProcessMeasurementsListContent.js
@@ -1,4 +1,4 @@
-import React, {useRef} from "react";
+import React, {useEffect, useRef} from "react";
 import {connect} from "react-redux";
 import {Link} from "react-router-dom";
 import {Button, Tag} from "antd";
@@ -151,6 +151,9 @@ const ProcessMeasurementsListContent = ({
   clearFilters,
   setSortBy,
 }) => {
+  useEffect(() => {
+    listTable()
+  }, [listTable])
 
   const listExport = () =>
     withToken(token, api.processMeasurements.listExport)


backend/fms_core/models/freezeman_user.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+class FreezemanUser(models.Model):
+    user = models.ForeignKey(User, on_delete=models.PROTECT, related_name='freezeman_user')
```
UlysseFG, line 5: use onetoonefield instead.

backend/fms_core/models/freezeman_user.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from django.contrib.auth.models import User
+
+class FreezemanUser(models.Model):
```
UlysseFG, line 4: Derive from trackedModel...

backend/fms_core/models/freezeman_user.py
```diff
@@ -0,0 +1,12 @@
...
+    user = models.ForeignKey(User, on_delete=models.PROTECT, related_name='freezeman_user')
+
+    @property
+    def username(self):
+        return self.user.username
+
+    def __str__(self):
+        return "FreezemanUser{username=%}" % self.username
```
UlysseFG, line 12: Don't use the internal model name in the str descriptor ... this could be shown to the user ... since username is unique you can just display that. The rest of the str might clash with context specific uses.

backend/fms_core/models/user_profile.py
```diff
@@ -0,0 +1,28 @@
...
+    "type": "object",
+    "properties": {
+        "table.sample.page-limit": {"type": "number"}
+    }
+}
+
+class UserProfile(models.Model):
+    user = models.ForeignKey(FreezemanUser, on_delete=models.PROTECT, related_name="user_profile")
+    preferences = models.JSONField(
+        default={
```
UlysseFG, line 21: don't hardcode specific preferences in the default json. This is likely to change overtime and we won't know what is default and what is not ... Leave it null if it is not defined ... We can override this at a higher level (within a service or viewset) to return the frontend a default set of preferences that can be updated without migration.

backend/fms_core/models/user_profile.py
```diff
@@ -0,0 +1,28 @@
...
+            "table.sample.page-limit": 100
+        },
+        validators=[JsonSchemaValidator(PREFERENCES_SCHEMA)]
+    )
+
+    def __str__(self):
+        return "UserProfile{username=%}" % str(self.user)
```
UlysseFG, line 28: profile is a many to one relation this will not work ... Offer "Profile " + profile.name + "*" if personalized ?

backend/fms_core/serializers.py
```diff
@@ -116,7 +118,9 @@
     "MetricSerializer",
     "ArchivedCommentSerializer",
     "SampleIdentityMatchSerializer",
-    "SampleIdentitySerializer"
+    "SampleIdentitySerializer",
+    "FreezemanUserSerializer",
```
UlysseFG, line 122: don't make a FreezemanUserSerializer. This is a stub for connecting internal tables to django user table ...

backend/fms_core/serializers.py
```diff
@@ -1039,4 +1043,19 @@ def get_sex_concordance(self, instance: SampleIdentity):
...
+    class Meta:
+        model = FreezemanUser
+        fields = ["id", "username"]
+
+class UserProfileSerializer(serializers.ModelSerializer):
+    user = FreezemanUserSerializer()
+
+    class Meta:
+        model = UserProfile
+        fields = ["user", "preferences"]
```
UlysseFG, line 1061: include name and personalized flag when added.

backend/fms_core/models/profile.py
```diff
@@ -0,0 +1,43 @@
...
+
+from ._utils import add_error as _add_error
+
+class Profile(TrackedModel):
+    # TODO: somehow allow many user for each profile
+    parent = models.ForeignKey("self", on_delete=models.PROTECT, blank=True, null=True)
+    preferences = models.JSONField(blank=True, null=True)
+
+    @property
+    def username(self):
```
UlysseFG, line 16: No use for this here.

backend/fms_core/models/profile.py
```diff
@@ -0,0 +1,43 @@
...
+        return self.user.username
+
+    def is_personalized(self):
+        return bool(self.preferences)
```
UlysseFG, line 20: the parent is what tells you if it is personalized ...

backend/fms_core/models/profile.py
```diff
@@ -0,0 +1,43 @@
+from django.db import models
+from django.core.exceptions import ValidationError
+
+from .tracked_model import TrackedModel
+from fms_core.schema_validators import PREFERENCES_VALIDATOR
+from fms_core.models.freezeman_user import FreezemanUser
+
+from ._utils import add_error as _add_error
+
+class Profile(TrackedModel):
```
UlysseFG, line 10: Still needs a name (and a description ideally).

backend/fms_core/models/profile.py
```diff
@@ -0,0 +1,43 @@
...
+        def add_error(field: str, error: str):
+            _add_error(errors, field, ValidationError(error))
+
+        preferences = self.final_preferences()
+        for error in PREFERENCES_VALIDATOR.validator.iter_errors(preferences):
+            path = "".join(f'["{p}"]' for p in error.path)
+            add_error("preferences", f"{path}: {error.message}" if error.path else error.message)
+
+    def __repr__(self) -> str:
+        return super().__repr__() + f" (user={self.user})"
```
UlysseFG, line 43: use name instead of user.

backend/fms_core/serializers.py
```diff
@@ -1039,4 +1042,28 @@ def get_sex_concordance(self, instance: SampleIdentity):
     
     def get_identity_matches(self, instance: SampleIdentity):
         matches = SampleIdentityMatch.objects.filter(Q(tested=instance)).all()
-        return SampleIdentityMatchSerializer(matches, many=True).data
\ No newline at end of file
+        return SampleIdentityMatchSerializer(matches, many=True).data
+    
+class FreezemanUserSerializer(serializers.ModelSerializer):
```
UlysseFG, line 1047: I don't think you need a serializer for this guy.

backend/fms_core/migrations/0074_profile_freezemanuser.py
```diff
@@ -0,0 +1,100 @@
+# Generated by Django 4.2.4 on 2025-10-02 18:49
```
UlysseFG, line 1: Use the existing migrations ...

backend/fms_core/migrations/0074_profile_freezemanuser.py
```diff
@@ -0,0 +1,100 @@
...
+from django.contrib.auth import get_user_model
+
+import reversion
+
+ADMIN_USERNAME = 'biobankadmin'
+
+def create_default_profile(apps, schema_editor):
+    User = get_user_model()
+    Profile = apps.get_model('fms_core', 'Profile')
+    reversion.register(Profile)
```
UlysseFG, line 16: Use decorator on the model.

backend/fms_core/migrations/0074_profile_freezemanuser.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        migrations.CreateModel(
+            name='FreezemanUser',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_at', models.DateTimeField(auto_now_add=True, help_text='Date the instance was created.')),
+                ('updated_at', models.DateTimeField(auto_now=True, help_text='Date the instance was modified.')),
+                ('deleted', models.BooleanField(default=False, help_text='Whether this instance has been deleted.')),
+                ('created_by', models.ForeignKey(blank=True, on_delete=django.db.models.deletion.PROTECT, related_name='%(app_label)s_%(class)s_creation', to=settings.AUTH_USER_MODEL)),
+                ('profile', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='users', to='fms_core.profile')),
```
UlysseFG, line 90: maybe user freezeman_users as related_name to make sure there is no confusion when we write in code.

backend/fms_core/models/profile.py
```diff
@@ -0,0 +1,39 @@
...
+        preferences = self.final_preferences()
+        for error in PREFERENCES_VALIDATOR.validator.iter_errors(preferences):
+            path = "".join(f'["{p}"]' for p in error.path)
+            add_error("preferences", f"{path}: {error.message}" if error.path else error.message)
+        
+        if errors:
+            raise ValidationError(errors)
+
+    def __repr__(self) -> str:
+        return super().__repr__() + f" (name={repr(self.name)})"
```
UlysseFG, line 39: Mmm why use the repr of the str name ? Looks like you want this to look more complex than it needs to be.

backend/fms_core/serializers.py
```diff
@@ -1039,4 +1042,15 @@ def get_sex_concordance(self, instance: SampleIdentity):
     
     def get_identity_matches(self, instance: SampleIdentity):
         matches = SampleIdentityMatch.objects.filter(Q(tested=instance)).all()
-        return SampleIdentityMatchSerializer(matches, many=True).data
\ No newline at end of file
+        return SampleIdentityMatchSerializer(matches, many=True).data
+
+class ProfileSerializer(serializers.ModelSerializer):
+    name = serializers.CharField(read_only=True)
```
UlysseFG, line 1048: model fields are there by default. I don't think we need to worry about read only here.

backend/fms_core/serializers.py
```diff
@@ -1039,4 +1042,15 @@ def get_sex_concordance(self, instance: SampleIdentity):
...
+    preferences = serializers.JSONField(read_only=True)
```
UlysseFG, line 1049: Did you want a methodfield ?

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,54 @@
+from collections import defaultdict
+from fms_core.models import Profile, FreezemanUser
+from fms_core.schema_validators import PREFERENCES_VALIDATOR
+
+def get_default_profile() -> Profile:
+    return Profile.objects.get(name="Default")
+
+def get_profile(user_id: int) -> Profile:
+    from fms_core.services.freezeman_user import get_or_create_freezeman_user
+    fm_user = get_or_create_freezeman_user(user_id)
```
UlysseFG, line 10: don't use get_or_create everywhere... you need to have cases where you would need to create the object if it does not exists (as a normal behaviour). Here if the Freezeman user does not exist we want it to crash and we want to investigate (it is not normal that the user does not exists).

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,54 @@
+from collections import defaultdict
+from fms_core.models import Profile, FreezemanUser
+from fms_core.schema_validators import PREFERENCES_VALIDATOR
+
+def get_default_profile() -> Profile:
+    return Profile.objects.get(name="Default")
+
+def get_profile(user_id: int) -> Profile:
+    from fms_core.services.freezeman_user import get_or_create_freezeman_user
```
UlysseFG, line 9: don't import inside your services unless there is a real reason ...

backend/fms_core/serializers.py
nafiz1001: Very likely yes...

backend/fms_core/services/profile.py
UlysseFG, line 10: I am not sure we want this by freezeman_user_id ... The frontend should be unaware of that id so will not be providing it ... there is not much added value to use that id here over the django user id.

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,48 @@
...
+    # parent should have all preference settings
+    for k, v in parent_profile.preferences.items():
+        if new_preferences.get(k, None) == v:
+            del new_preferences[k]
+
+    if not new_preferences and fm_user.profile.parent is None:
+        # don't update user preferences if the user hasn't personalized yet anyway which means they stuck with default
+        return fm_user.profile, errors, warnings
+
+    if fm_user.profile.parent is None:
```
UlysseFG, line 37: where is the else ? What happens when a profile has a parent ? this is where the actual update_preferences happens.

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,48 @@
...
+        fm_user.profile = Profile.objects.create(
+            name=f"{fm_user.username}",
+            parent=fm_user.profile,
+            preferences=new_preferences
+        )
+
+    fm_user.profile.save()
```
UlysseFG, line 44: create() saves the object automatically ... You would need to do some **updating** to need to save ;).

backend/fms_core/viewsets/profile.py
```diff
@@ -0,0 +1,44 @@
...
+
+class ProfileViewSet(viewsets.ModelViewSet):
+    queryset = Profile.objects.all()
+    serializer_class = ProfileSerializer
+    permission_classes = [IsAuthenticated]
+
+    http_method_names = ['get', 'post', 'head', 'options', 'trace']
+
+    def retrieve(self, _request, pk, *args, **kwargs):
+        return Response(bad_request(_request, "Fetch individual profile via /profile/?user_id=USER_ID"))
```
UlysseFG, line 23: Mmm ? You might want to return profiles from time to time ...

backend/fms_core/models/profile.py
nafiz1001: `__repr__` is what's presented when debugging. I figured overloading `__str__` might not be useful for us.

backend/fms_core/viewsets/profile.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def list(self, _request, *args, **kwargs):
+        params = QueryDict(self.request.META.get('QUERY_STRING'))
+        user_id = params.get("user_id")
+        if not user_id:
+            raise ValidationError({"user_id": "user_id parameter is required"})
+        profile = get_profile(int(user_id))
+        return Response(self.get_serializer(profile).data)
+    
+    def update(self, request, *args, **kwargs):
```
UlysseFG, line 33: update method signature should include a pk (from the route)

UlysseFG: https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions

backend/fms_core/viewsets/profile.py
```diff
@@ -0,0 +1,44 @@
...
+from fms_core.models import FreezemanUser, Profile
+from fms_core.serializers import ProfileSerializer
+
+
+class ProfileViewSet(viewsets.ModelViewSet):
+    queryset = Profile.objects.all()
+    serializer_class = ProfileSerializer
+    permission_classes = [IsAuthenticated]
+
+    http_method_names = ['get', 'post', 'head', 'options', 'trace']
```
UlysseFG, line 21: educational content ?

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,48 @@
...
+        errors.append(f"{path}: {error.message}" if error.path else error.message)
+    if errors:
+        return fm_user.profile, errors, warnings
+
+    parent_profile = fm_user.profile.parent if fm_user.profile.parent else fm_user.profile
+    # Remove any new preferences that are the same as the parent's preferences
+    # parent should have all preference settings
+    for k, v in parent_profile.preferences.items():
+        if new_preferences.get(k, None) == v:
+            del new_preferences[k]
```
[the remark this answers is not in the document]
nafiz1001: This is outdated review ;)

backend/fms_core/services/profile.py
```diff
@@ -0,0 +1,47 @@
...
+
+def get_profile_by_name(name = "Default") -> Profile:
+    return Profile.objects.get(name=name)
+
+def get_profile_by_user_id(user_id: int) -> Profile:
+    fm_user = get_freezeman_user(user_id)
+    return fm_user.profile
+
+def update_profile_preferences(user_id: int, new_preferences: dict) -> tuple[Profile, list[str], list[str]]:
+    new_preferences = dict(new_preferences)
```
UlysseFG, line 13: why name it with the same name ?

backend/fms_core/services/profile.py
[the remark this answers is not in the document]
UlysseFG: The comment about an error if you have a key with a None value is still valid.

backend/fms_core/viewsets/profile.py
UlysseFG: for anything ;)

backend/fms_core/migrations/0074_profile_freezemanuser.py
```diff
@@ -0,0 +1,98 @@
+import re
+
```
UlysseFG: Merge this file into 0072 migration.

backend/fms_core/tests/test_services/profile_test.py
```diff
@@ -0,0 +1,43 @@
+from django.test import TestCase
+
+from fms_core.models import Profile, FreezemanUser
+
+# from fms_core.services.profile import update_user_profile
+
+class ProfileServicesTestCase(TestCase):
```
UlysseFG: This test is failing ...

UlysseFG:
I think we need to review the way we apply the preferences to the frontend ...
The current hook approach is going to get really complex really fast (if we have many preferences on the same pages).
I think one basic thing is that we should centralize the variables that are affected by preferences and make sure they get set once when loading the profile at the start of the app (either using the profile value or a default value).
This will make it so we don't bloat the code all over the frontend with tests and calls to the profile. The variables will be set at that point with the correct preference value ... The local copies would use these centralized values as default and still be modifiable locally.

UlysseFG:
We should remove the commented out parts ... We have at least 2 files in there that are useless. Make a copy somewhere of the bits you want to keep (or take not of the commit hash for references). This is a slippery slope and since we don't know when we'll get back into preferences, i don't want this to be there forever.

backend/fms_core/migrations/0072_v5_3_0.py
```diff
@@ -215,6 +215,47 @@ def create_ffpe_sample_kind(apps, schema_editor):
...
+    with reversion.create_revision(manage_manually=True):
+        admin_user = User.objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment(f"Create default profile.")
+        reversion.set_user(admin_user)
+
+        default_profile = Profile.objects.create(
+            name='Default',
+            parent=None,
+            preferences={
+                "table.sample.page-limit": 20,
```
UlysseFG: your default is wrong ... it should be 100.

frontend/src/modules/profiles/selectors.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+)
+
+const DEFAULT_PREFERENCES: FMSProfilePreferences = {
+    'table.sample.page-limit': 100,
+}
+export const selectCurrentPreferences = createSelector(
+    [selectCurrentUserProfile],
+    (profile) => {
+        return profile ? profile.preferences : DEFAULT_PREFERENCES
```
UlysseFG: this won't protect against a missing entry in the preference object... if there is a single key missing you will still get a missing key when asked for that key... Maybe you should wrap this as selectCurrentPreference(key) and in the selector return either the profile.preferences[key] or DEFAULT_PREFERENCES[key] otherwise ...

UlysseFG:
The protocol page no longer load.
The loading of sample is a bit longer but it is not so bad. (ok)

UlysseFG (approved):

PR #1052, opened 2025-10-03
Correct dropdown for container kind on Pacbio experiment template.

UlysseFG:


nafiz1001:
I get `pacbio-revio smrt cell tray` when I pick `Revio (r84240)`.

PR #1136, opened 2026-04-14
Python 3.14

nafiz1001:


backend/fms_core/template_importer/_utils.py
```diff
@@ -46,12 +47,8 @@ def panda_values_to_str_list(row_data):
     return ['' if x is None else str(x) for x in row_data.values.flatten().tolist()]
 
 
-def blank_and_nan_to_none(s: Any):
-    """
-    Returns None if the argument is a nan (pandas) or a blank string, or the argument with no
-    changes otherwise.
-    """
-    return None if (s == "" or pd.isnull(s)) else s
+def blank_and_nan_to_none(d: pd.DataFrame) -> pd.DataFrame:
```
UlysseFG, line 50: was it necessary to replace s with an as bad (or worse) variable name ? also you can keep the comment ...with an extra note (warning) about Pandas replacing None by NaN any occasion it gets ...

backend/fms_core/workbooks/sample_rename.py
```diff
@@ -92,7 +92,7 @@ def red_text(cell: Cell):
 
         center_alignment = Alignment(horizontal="center")
 
-        ws = self.get_sheet_by_name(SHEET_NAMES[0])
+        ws = self[SHEET_NAMES[0]]
```
UlysseFG, line 95: this is much less readable ...

backend/requirements.txt
```diff
@@ -1,92 +1,38 @@
...
-xlrd==1.2.0
-xlwt==1.3.0
-zipp==3.1.0
+python-dateutil==2.9.0.post0
+pytz==2026.1.post1
+psycopg2==2.9.11
+PyYAML==6.0.3
+referencing==0.37.0
+rfc3339-validator==0.1.4
+rpds-py==0.30.0
```
UlysseFG, line 35: Really ? or are you trying to creep in some package creep ?

backend/fms_core/workbooks/sample_rename.py
nafiz1001: `get_sheet_by_name` is getting deprecated ;)

backend/requirements.txt
nafiz1001: I don't think I've included those explicitly. Blame requirements.txt not me ;)
(ideally we would be using `pyproject.toml` to distinguish between direct and indirect dependencies)

nafiz1001: Well technically it was my idea to upgrade all packages at once.

nafiz1001: You can check this for justification of each package I've installed explicitly:
https://github.com/c3g/freezeman/pull/1136/changes/baa89b61af840f96359dd6073c395cd0033b3081

PR #1137, opened 2026-04-28
Some things that should've been logged

nafiz1001:


nafiz1001:
Anything else to log?

backend/fms/settings.py
```diff
@@ -245,9 +245,13 @@
...
             "backupCount": LOG_BACKUP_COUNT
         },
+        "console": {
+            "formatter": "fms",
+            "class": "logging.StreamHandler",
+        }
     }
     logger = {
-        "handlers": ["file"],
+        "handlers": ["file", "console"],
```
UlysseFG: I am not sure why you are doing this ... The logging is mostly disabled in prod to preserve performance. You can add any number of print() locally whenever you are debugging locally. Is your plan to have the code riddled with log messages  ? This is a really weird thing to do. log a random code bit that i can't recall ever failing. You are adding more prod level traceability for dev purposes. You should not be adding code to match your debugging needs whenever they arise... Addendum : sorry i am a bit grumpy. I see this can help with some case where our local structure does not match the server structure and could lead to solve left over issues hard to test locally.

backend/fms_core/template_importer/importers/_generic.py
```diff
@@ -58,25 +60,28 @@ def import_template(self, file: Path | InMemoryUploadedFile, dry_run, user = Non
...
+                            new_file_name = f"{file_name}_{time.strftime('%Y-%m-%d_%H-%M-%S')}_{user.username}{file_format}"
+                            file_path = os.path.join(settings.TEMPLATE_UPLOAD_PATH, new_file_name)
+                            try:
+                                self.imported_file = ImportedFile.objects.create(filename=new_file_name, location=file_path, created_by_id=user.id)
+                            except Exception as err:
+                                self.base_errors.append(err)
+                        self.import_template_inner()                            
+                        reversion.set_comment("Template import")
+                except:
+                    self.logger.debug("Error during template import. Transaction rolled back.", exc_info=True)
```
UlysseFG: why not use logger.error() ?

UlysseFG: Do not forget to raise again ...

backend/fms/settings.py
nafiz1001: It is admittedly a very spammy log anyway. So yeah I should remove it.

backend/fms_core/template_importer/importers/_generic.py
nafiz1001: Oh woops. Right. Forgot. About both.

UlysseFG (approved):

PR #1138, opened 2026-05-01
the scrollable ExperiementRunsDetailsContent issue is corrected

nabalma:
This is issue Number 2792. It was a css issue on the overflow attribute to bet set to scroll instead of hidden


nafiz1001 (approved):
Looks good to me.

UlysseFG (approved):

PR #1139, opened 2026-05-01
the sample details Content Page is now scrollable

nabalma:
the sample details Content page is now scrollable . This is a fix for the issue Number 2793

### DIFF
--- a/frontend/src/components/samples/details/SampleDetailsContent.js
+++ b/frontend/src/components/samples/details/SampleDetailsContent.js
@@ -39,7 +39,7 @@ const depletedStyle = {
 
 const pageStyle = {
   padding: 0,
-  overflow: "scroll",
+  overflow: "hidden",
 }
 
 const tabStyle = {
--- a/frontend/src/components/samples/details/SampleDetailsContent.js
+++ b/frontend/src/components/samples/details/SampleDetailsContent.js
@@ -39,7 +39,7 @@ const depletedStyle = {
 
 const pageStyle = {
   padding: 0,
-  overflow: "hidden",
+  overflow: "scroll",
 }
 
 const tabStyle = {
--- a/frontend/src/hooks/usePaginatedList.js
+++ b/frontend/src/hooks/usePaginatedList.js
@@ -88,7 +88,7 @@ export const usePaginatedList = ({
             loading: loading || isCurrentPageUnloaded,
             childrenColumnName: 'UNEXISTENT_KEY',
             onChange: onChangeTable,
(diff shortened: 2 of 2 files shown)


frontend/src/components/samples/details/SampleDetailsContent.js
```diff
@@ -39,7 +39,7 @@ const depletedStyle = {
 
 const pageStyle = {
   padding: 0,
-  overflow: "hidden",
+  overflow: "scroll",
```
UlysseFG: This addresses the immediate issue with seeing the table content but we'd prefer a solution that restrain the content of the table to the viewable page and scrolls the table content instead ... This was done in a few places already (you can check the sample and library list pages ... they list more than the page can show but offer to scroll the table content keeping the pagination section section visible at all time).

nabalma: Hi , this is how I address it now.
- 1) I rolled back the scroll that was done in the preview commit in the sampleDetailsContent file
- 2) After checking, The table was fed by "tableProps" that comes from the "usePaginatedList" hook. That hook return an x attribut scroll, (y attribut was not included in the return ). I have included an y attribut so that the table will be y scrollable, allowing  users to see all samples listed.
image

frontend/src/hooks/usePaginatedList.js
```diff
@@ -88,7 +88,7 @@ export const usePaginatedList = ({
             loading: loading || isCurrentPageUnloaded,
             childrenColumnName: 'UNEXISTENT_KEY',
             onChange: onChangeTable,
-            scroll: { x: 300 },
+            scroll :{ x: '100%', y: '50vh' },
```
nafiz1001: The colon should have space after, not before it.

nafiz1001 (approved):

frontend/src/components/samples/details/SampleDetailsContent.js
UlysseFG: The issue is fixed. I however have a couple nitpicks that maybe you can resolve.
1. There is a lot of white space left on my screen resolution (which is not very high). Maybe we can get the table to cover more of the available space.
2. The columns content seem to behave differently than before. They seem to force text wrapping wereas the same page on production is more compact. Not sure if this can be solved elegantly while preserving the scrolling table,

nabalma: Thanks @UlysseFG , 
Of course, I will have a look of these !!!

PR #1142, opened 2026-05-07
Labwork page taking half page

nabalma:
The page content of the LabworkOverview component had a maxWidth: '50rem' style set as a property. 
I wonder why. I have commented it out.

### DIFF
--- a/frontend/src/hooks/usePaginatedList.js
+++ b/frontend/src/hooks/usePaginatedList.js
@@ -88,7 +88,7 @@ export const usePaginatedList = ({
             loading: loading || isCurrentPageUnloaded,
             childrenColumnName: 'UNEXISTENT_KEY',
             onChange: onChangeTable,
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/usePaginatedList.js
```diff
@@ -88,7 +88,7 @@ export const usePaginatedList = ({
             loading: loading || isCurrentPageUnloaded,
             childrenColumnName: 'UNEXISTENT_KEY',
             onChange: onChangeTable,
-            scroll: { x: 300 },
+            scroll: { x: '100%', y: '50vh' },
```
UlysseFG, line 91: is this part of this PR ? looks like another PR content creeped over here ...

nabalma: I see !. My Bad. Sorry.
May be because I created the new branch from the the another, not from the master branch. 
But yeah, it should have been another separated not mixed PR.

PR #1144, opened 2026-05-08
Selection count for IndexCuration and SampleRename

nafiz1001:


UlysseFG:
Very nice to have this at the top ...
Nitpick: it makes it even more obvious there is a loading glitch that makes everything selected a fraction of a second ...
Any way to get rid of that ?

frontend/src/utils/tableHooks.tsx
```diff
@@ -346,7 +346,7 @@ export function useSmartSelectionProps<RowData extends AntdAnyObject>({
 		if (defaultSelection && exceptedItems.length === totalCount) {
 			setDefaultSelection(false)
 			setExceptedItems([])
-		} else if (!defaultSelection && exceptedItems.length === totalCount) {
+		} else if (!defaultSelection && exceptedItems.length === totalCount && totalCount > 0) {
```
UlysseFG, line 349: defaultSelection is not really an informative name ... even more so that by default it is false ...
I really would prefer not to have to do a deep dive to just figure if this is a positiveSelection (you count actually selected items) or a negativeSelection (you count the items that are not selected).

UlysseFG, line 349: the additional check could be added to the first if too. No point in forcing refresh on a empty list.

frontend/src/utils/tableHooks.tsx
```diff
@@ -438,6 +438,12 @@ export function useSmartSelectionProps<RowData extends AntdAnyObject>({
 
 
     const totalSelectionCount = defaultSelection ? totalCount - exceptedItems.length : exceptedItems.length
+    console.info({
```
UlysseFG: spam

UlysseFG:
The current version seem to generate an infinite loop ... The backend spams error messages non-stop.

nafiz1001:
> The current version seem to generate an infinite loop ... The backend spams error messages non-stop.

I don't see the spam on my side. To me, each action that triggers API call happens only once.

frontend/src/components/management/PooledSamples/index.tsx
UlysseFG: Maybe we should rename that requestID to something that makes more sense in the current setup (this request could originate from a nameCuration).

frontend/src/components/management/PooledSamples/index.tsx
```diff
@@ -151,7 +151,7 @@ export function PooledSamples({
             total: response.data.count,
             data: response.data.results
         }
-    }, [dispatch])
+    }, [dispatch, filterOptions])
```
UlysseFG: For SampleRename the filterOptions is a default value. Each time the component rerender the value changes forcing the callback to be rerun. Each rerun cancel the previous request forcing an infinite rerender loop. I fixed the issue on my side by providing a memoed empty value ({}) to the component prop in the parent component.

frontend/src/components/management/SampleRename/index.tsx
```diff
@@ -42,6 +42,7 @@ export function SampleRename() {
         actionUrlBase={"/management/sample-rename"}
         templateAction={templateAction}
         templatePrefill={templatePrefill}
+        requestID={"SampleRename.fetchPooledSamples"}
```
UlysseFG: This is a weird design choice ... the requestID could be one of many used inside the component ... If you want to be fancy maybe use a context manager to provide the appropriate ID for each cases in your pooledSample component ? I don't want to make this more complex that this is ... I just find this to be too specific a prop for the component distance.

UlysseFG (approved):

PR #1145, opened 2026-05-08
Fix visibility of select dropdown in Last Launched Experiments

nafiz1001:


nafiz1001:
On the master branch on my work computer, it is actually invisible. I'm not sure why why.

UlysseFG:
I am not sure what this aim to address. I don't see a dropdown in that card anywhere ...

UlysseFG:
If you want to get this merged for next version it would be the time ...

nafiz1001:
I also magically fixed the issue by just `npm install`ing.

PR #1147, opened 2026-05-11
Minor tweaks to migration

UlysseFG:


nafiz1001 (approved):
That's a lot to see. But I see "ultima wafer" in each so looks alright to me.

PR #1148, opened 2026-05-13
Create tmp trigger files then move to final destination once written.

UlysseFG:


backend/fms_core/services/dataset.py
```diff
@@ -1,5 +1,5 @@
 import json
-from os import path
+from os import path, rename
```
nafiz1001, line 2: It's valid but it doesn't feel right seeing `rename` all alone without the `os.` 🥹

backend/fms_core/services/dataset.py
```diff
@@ -397,9 +397,12 @@ def create_validation_info_file(dataset_obj: Dataset, is_validation_revocation:
         file_definition = {"readset_id": dataset_file.readset.id, "filepath": dataset_file.file_path}
         validated_data["files"][dataset_file.id] = file_definition
     try:
-        # Create file if it doesn't already exist
-        with open(file_path, "x") as fp:
-            fp.write(json.dumps(validated_data, indent=4))
+        # Create file with .tmp extension if it doesn't already exist
+        tmp_file_path = file_path + ".tmp"
```
nafiz1001: `f"{file_path}.tmp"`
Just personal preference.

backend/fms_core/services/dataset.py
```diff
@@ -397,9 +397,12 @@ def create_validation_info_file(dataset_obj: Dataset, is_validation_revocation:
...
+        with open(tmp_file_path, "x") as fp:
+            fp.write(json.dumps(validated_data, indent=4))
```
nafiz1001, line 403: Something to note is there's https://docs.python.org/3/library/json.html#json.dump which takes `fp` as argument.

UlysseFG: I am not sure i understand the intent of your comment.

nafiz1001: Basically, instead of `fp.write ∘ json.dumps` (function composition), it could be just `json.dump`

UlysseFG: I honestly think the current version is clearer ... I am pretty sure json.dump should not exist it is confusing (it is at one typo from a different function altogether) all to save 10 characters worth of code. I will keep it as is, but thanks for pointing that out.

PR #1157, opened 2026-06-11
Validate External Project ID for validation and release

nafiz1001:


backend/fms_core/services/dataset.py
```diff
@@ -303,6 +308,10 @@ def set_dataset_release_status(dataset_id: int, readsets_release_status: dict[st
         readset_ids = [int(i) for i in readsets_release_status.keys()]
         readsets = Readset.objects.filter(dataset=dataset_id, id__in=readset_ids)
 
+        if not dataset_obj.project.external_id:
```
UlysseFG, line 309: Why put this after the readsets lines ? no point in running these if there is not external id ...

nafiz1001: I blame AI ;)

frontend/src/modules/experimentRunLanes/actions.ts
```diff
@@ -31,18 +32,28 @@ export function loadReadsPerSample(experimentRunId: FMSId, lane: LaneNumber) {
...
-			// everything from the server, we just update the lane's status in the redux store.
-			dispatch({
-				type: SET_LANE_VALIDATION_STATUS,
-				experimentRunId: lane.experimentRunId,
-				laneNumber: lane.laneNumber,
-				status: status
-			})
-		}
-	}
+        try {
```
UlysseFG, line 35: Check why this relies on try catch instead of processing normally the error message.

UlysseFG (approved):

PR #1159, opened 2026-06-15
Set lane validation status atomic

UlysseFG:


nafiz1001 (commented):
Wouldn't returning `HttpResponseServerError` cause a rollback anyway?

UlysseFG:
> Wouldn't returning `HttpResponseServerError` cause a rollback anyway?

http response does not imply there is an active transaction open ... I would tend to think not. As far as i can see the normal behaviour is to autocommit. A response even an error response is not an raise exception (and don't start raising exception on the backend side too).

nafiz1001 (approved):

nafiz1001 (commented):
Actually, wouldn't it be better to make the service atomic instead?

UlysseFG:
> Actually, wouldn't it be better to make the service atomic instead?

An atomic service could still lead to an issue ... you could call the service mutliple times in a single endpoint call.

PR #1162, opened 2026-06-22
Skippable step

nafiz1001:
- [No longer require explicit React import...](https://github.com/c3g/freezeman/pull/1162/commits/b7d4c0fe6daca744ef9d8f1172a042cf3bc96d19)
- [highlight "Assign Workflow" in sidebar](https://github.com/c3g/freezeman/pull/1162/commits/4df3a4fa62ed2e36aff1abb41ed559cae58c3ded)
- Skip action in "Assign Workflow"
- Skip action in study tabs of project (similar to removal)

### DIFF
--- a/backend/fms_core/migrations/0082_v5_9_0.py
+++ b/backend/fms_core/migrations/0082_v5_9_0.py
@@ -8,10 +8,26 @@ import django.db.models.deletion
 
 ADMIN_USERNAME = 'biobankadmin'
 
+def make_library_normalization_step_optional(apps, schema_editor):
+    StepOrder = apps.get_model('fms_core', 'StepOrder')
+    with reversion.create_revision(manage_manually=True):
+        admin_user = get_user_model().objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment('Make Normalization (Library) step optional for Ultima workflow.')
+        reversion.set_user(admin_user)
+
+        step_order = StepOrder.objects.get(
+            step__name="Normalization (Library)",
+            workflow__name="Ready-to-Sequence Ultima"
+        )
+
+        step_order.mandatory = False
+        step_order.save()
+        reversion.add_to_revision(step_order)
+
 def populate_parent_project(apps, schema_editor):
     Project = apps.get_model("fms_core", "Project")
     ParentProject = apps.get_model("fms_core", "ParentProject")
-    
+
     with reversion.create_revision(manage_manually=True):
         admin_user = get_user_model().objects.get(username=ADMIN_USERNAME)
 
@@ -46,7 +62,7 @@ def populate_parent_project(apps, schema_editor):
 def populate_foreign_key_to_parent_project(apps, schema_editor):
     Project = apps.get_model("fms_core", "Project")
     ParentProject = apps.get_model("fms_core", "ParentProject")
-    
+
     with reversion.create_revision(manage_manually=True):
         admin_user = get_user_model().objects.get(username=ADMIN_USERNAME)
 
@@ -118,4 +134,16 @@ class Migration(migrations.Migration):
             name='container',
             field=models.ForeignKey(help_text='Container in which the sample is placed.', limit_choices_to={'kind__in': ('axiom 96-format array pmra', 'axiom 96-format array ukbb', 'infinium epic 8 beadchip', 'infinium gs 24 beadchip', 'dnbseq-g400 flowcell', 'dnbseq-t7 flowcell', 'illumina-novaseq-x-1.5b flowcell', 'illumina-novaseq-x-5b flowcell', 'illumina-novaseq-x-10b flowcell', 'illumina-novaseq-x-25b flowcell', 'illumina-novaseq-sp flowcell', 'illumina-novaseq-s1 flowcell', 'illumina-novaseq-s2 flowcell', 'illumina-novaseq-s4 flowcell', 'illumina-miseq-v2 flowcell', 'illumina-miseq-v3 flowcell', 'illumina-miseq-micro flowcell', 'illumina-miseq-nano flowcell', 'illumina-miseq-i100-5m flowcell', 'illumina-miseq-i100-25m flowcell', 'illumina-miseq-i100-50m flowcell', 'illumina-miseq-i100-100m flowcell', 'illumina-iseq-100 flowcell', 'pacbio-revio smrt cell tray', 'ultima wafer', 'tube', 'tube strip 2x1', 'tube strip 3x1', 'tube strip 4x1', 'tube strip 5x1', 'tube strip 6x1', 'tube strip 7x1', 'tube strip 8x1', '96-well plate', '384-well plate')}, on_delete=django.db.models.deletion.PROTECT, related_name='samples', to='fms_core.container'),
         ),
+
+        migrations.AddField(
+            model_name='steporder',
+            name='mandatory',
+            field=models.BooleanField(default=True, help_text='Samples cannot skip this step in this workflow.'),
+        ),
+        migrations.AlterField(
+            model_name='stephistory',
+            name='workflow_action',
+            field=models.CharField(choices=[('NEXT_STEP', 'Step complete - Move to next step'), ('DEQUEUE_SAMPLE', 'Sample failed - Remove sample from study workflow'), ('REPEAT_STEP', 'Repeat step - Move to next step and repeat current step'), ('REPEAT_QC_STEP', 'Repeat QC step - Repeat current QC step'), ('SKIP_STEP', 'Step skipped - Move to next step'), ('IGNORE_WORKFLOW', 'Ignore workflow - Do not register as part of a workflow')], default='NEXT_STEP', help_text='Workflow action that was performed on the sample after step completion.', max_length=30),
+        ),
+        migrations.RunPython(make_library_normalization_step_optional, reverse_code=migrations.RunPython.noop),
     ]
--- a/backend/fms_core/models/step_order.py
+++ b/backend/fms_core/models/step_order.py
@@ -18,6 +18,7 @@ class StepOrder(TrackedModel):
     next_step_order = models.ForeignKey("self", null=True, blank=True, on_delete=models.PROTECT, related_name="previous_step_order", help_text="The next step following the one defined here.")
     order = models.PositiveIntegerField(help_text="Ordinal value of the step in the workflow (starting at 1).")
     workflow = models.ForeignKey(Workflow, on_delete=models.PROTECT, related_name="steps_order", help_text="Workflow of the step order.")
+    mandatory = models.BooleanField(default=True, help_text="Indicates that this step cannot be skipped in this workflow.")
 
     class Meta:
         indexes = [
--- a/backend/fms_core/services/sample_next_step.py
+++ b/backend/fms_core/services/sample_next_step.py
@@ -1,4 +1,5 @@
 from django.core.exceptions import ValidationError
+from django.db import transaction
 from django.db.models import Case, When, Q, F, BooleanField
 from fms_core.models import SampleNextStep, SampleNextStepByStudy, StepOrder, Sample, Study, Step, ProcessMeasurement, StepHistory
 from fms_core._constants import WorkflowAction
@@ -15,9 +16,9 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order: 
         `study_obj`: Study instance that will be associated to the sample.
         `order`: Positive integer indicating the insertion order of the sample in the workflow. Optional: defaults to the start
                  order of the study workflow.
-                 
+
     Returns:
-        Tuple containing the created SampleNextStep instance (if applicable, otherwise None), the error messages and the warning messages. 
+        Tuple containing the created SampleNextStep instance (if applicable, otherwise None), the error messages and the warning messages.
 
     """
     sample_next_step = None
@@ -36,13 +37,13 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order: 
         order = study_obj.start
     elif order < study_obj.start or order > study_obj.end:
         errors.append(f"Order must be a positive integer between {study_obj.start} and {study_obj.end}.")
-    
+
     step_order = None
     try:
         step_order = StepOrder.objects.get(order=order, workflow=study_obj.workflow)
     except StepOrder.DoesNotExist:
         errors.append(f"No step found for the given order.")
-    
+
     # Queueing to study workflow implies an existing step order.
     if step_order and sample_obj and study_obj and not errors:
         if sample_obj.matches_sample_type(step_order.step.expected_sample_type):
@@ -65,14 +66,14 @@ def queue_sample_to_study_workflow(sample_obj: Sample, study_obj: Study, order: 
 
 def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_obj: Study, order: int) -> Tuple[bool, List[str], List[str]]:
     """
-    Deletes a SampleNextStep instance to indicate the removal of a sample from a workflow at a specific step. 
+    Deletes a SampleNextStep instance to indicate the removal of a sample from a workflow at a specific step.
     Args:
         `sample_obj`: Sample instance to be dequeued from the study workflow.
         `study_obj`: Study instance currently associated to the sample.
         `order`: Positive integer indicating the current position of the sample in the workflow.
-                 
+
     Returns:
-        Tuple containing a boolean stating whether the SampleNextStep instances was deleted or not, the error messages and the warning messages. 
+        Tuple containing a boolean stating whether the SampleNextStep instances was deleted or not, the error messages and the warning messages.
 
     """
     dequeued = False
@@ -86,11 +87,11 @@ def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_o
     if not isinstance(study_obj, Study):
         study_obj = None
         errors.append(f"A valid study instance must be provided.")
-    
+
     if not order:
         order = None
         errors.append(f"A step order must be provided.")
-    
+
     step_order=None
     # Dequeuing from a specific step
     if order is not None:
@@ -98,7 +99,7 @@ def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_o
             step_order = StepOrder.objects.get(order=order, workflow=study_obj.workflow)
         except StepOrder.DoesNotExist:
             errors.append(f"No step found for the given order.")
-    
+
     # Dequeuing from study workflow by deleting the SampleNextStep instance
     if sample_obj and study_obj and not errors:
         try:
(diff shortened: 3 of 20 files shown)


backend/fms_core/services/sample_next_step.py
```diff
@@ -313,7 +334,12 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
             for sample_next_step_by_study in SampleNextStepByStudy.objects.filter(sample_next_step=current_sample_next_step).all() :
                 next_sample_next_step = None
                 study = sample_next_step_by_study.study
+                
                 current_step_order = sample_next_step_by_study.step_order
+                if current_step_order.mandatory and workflow_action == WorkflowAction.SKIP_STEP:
```
UlysseFG, line 343: This is a very sensible check ... however if you let it run you'll potentially complete the operation for some sample_next_step while the whole function will error out. The exact case this could happen is not really obvious (to me too) but the code is permissive and would let this happen and it is not desired. I would recommend moving that test out of the loop and skipping the rest uf there is an error.  However the rest of the function already play it loose so maybe we could think of a better way to deal with errors in here. The code here was clearly meant to be run within a transaction that would be rolled back on error, but nothing forces a transaction to be used.

backend/fms_core/migrations/0082_v5_9_0.py
```diff
@@ -8,10 +8,26 @@
 
 ADMIN_USERNAME = 'biobankadmin'
 
+def make_library_normalization_step_optional(apps, schema_editor):
+    StepOrder = apps.get_model('fms_core', 'StepOrder')
+    with reversion.create_revision(manage_manually=True):
+        admin_user = get_user_model().objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment('Make Normalization (Library) step optional.')
```
UlysseFG: add "for Ultima" to the comment to clarify this is not broadly applied.

backend/fms_core/models/step_order.py
```diff
@@ -18,6 +18,7 @@ class StepOrder(TrackedModel):
     next_step_order = models.ForeignKey("self", null=True, blank=True, on_delete=models.PROTECT, related_name="previous_step_order", help_text="The next step following the one defined here.")
     order = models.PositiveIntegerField(help_text="Ordinal value of the step in the workflow (starting at 1).")
     workflow = models.ForeignKey(Workflow, on_delete=models.PROTECT, related_name="steps_order", help_text="Workflow of the step order.")
+    mandatory = models.BooleanField(default=True, help_text="Whether this step can be skipped in this workflow.")
```
UlysseFG: The help text seem to match the reverse bool ... mandatory indicates that this step cannot be skipped in this workflow.

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,19 +326,27 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+    with transaction.atomic():
         for current_sample_next_step in current_sample_next_steps.all():
-            for sample_next_step_by_study in SampleNextStepByStudy.objects.filter(sample_next_step=current_sample_next_step).all() :
+            for sample_next_step_by_study in SampleNextStepByStudy.objects.filter(sample_next_step=current_sample_next_step).all():
                 next_sample_next_step = None
                 study = sample_next_step_by_study.study
+                
                 current_step_order = sample_next_step_by_study.step_order
+                if current_step_order.mandatory and workflow_action == WorkflowAction.SKIP_STEP:
+                    errors.append(f"Step '{current_step.name}' cannot be skipped in workflow '{study.workflow.name}'.")
```
UlysseFG: I am not sure this should be an error ... A warning may be more advised. If we have a sample Queued to a workflow step on 2 workflow ... 1 is mandatory 1 is not ... if i was to make it skip the step i would hope it skip the step for the one that is not mandatory and let it in place for the one that is mandatory.

backend/fms_core/services/sample_next_step.py
UlysseFG: I would think an error here would be worthy of a rollback...

UlysseFG: if a mandatory step prevented partially for a sample to be moved we would need to validate each sample_next_step_by_study before removing them when completing a skip_step to make sure we are not removing the part that was mandatory ...

backend/fms_core/services/sample_next_step.py
```diff
@@ -353,11 +385,13 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
                         sample_next_step_by_study.delete()
                     current_sample_next_step.delete()
```
UlysseFG: if there is still sample_next_step_by_study for the current_sample_next_step you cannot remove it yet.

backend/fms_core/services/sample_next_step.py
```diff
@@ -123,6 +124,27 @@ def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_o
             errors.append(err)
     return dequeued, errors, warnings
 
+def skip_sample_over_specific_step_study_workflow(sample: Sample, study: Study, order: int):
```
UlysseFG: this function does not deliver on the expected merchandise ... The specific part of this function get lost as soon as you get the step_order ... the step_order is shared by different studies using the same workflow ... converting a step_order to a step go even further ... you are losing the order so you might actually skip another step_order in the same workflow using the same step. The underlying move_sample_to_next_step() call is where you need to compromise ... the real solution to get a specific skip is to either modify the move_sample_to_next_step to take a study and step order or to make a new service that is more specific. You could also rename the service and change the params to have an unspecific skip... I do not think this has to be specific if all checks are done in move_sample_to_next_step(), but we should not sell it as such.

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,63 +323,103 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+        except Exception as err:
+            errors.append(f"Failed to create new sample next step instance.")
+    try:
+        # Create the entry in StepHistory
+        StepHistory.objects.create(study=study,
+                                step_order=current_step_order,
+                                process_measurement=process_measurement,
+                                sample=current_sample,
+                                workflow_action=workflow_action)
+        pass
```
UlysseFG: pass what ?

nafiz1001 (commented):
Welp. Some tests failed...

nafiz1001:
Oh, never mind. Just had to `collectstatic` 😪

backend/fms_core/services/sample_next_step.py
```diff
@@ -123,6 +124,24 @@ def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_o
             errors.append(err)
     return dequeued, errors, warnings
 
+def skip_by_sample_next_step_by_study(sample_next_step_by_study: SampleNextStepByStudy):
+    errors = []
+    warnings = []
+
+    try:
+        if sample_next_step_by_study.step_order.mandatory:
+            warnings.append(f"Step '{sample_next_step_by_study.step_order.step.name}' cannot be skipped in workflow '{sample_next_step_by_study.study.workflow.name}'.")
```
UlysseFG: this should be an error still.

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +323,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+            if SampleNextStep.objects.filter(step=next_step_order.step, sample=new_sample).exists():
+                next_sample_next_step = SampleNextStep.objects.get(step=next_step_order.step, sample=new_sample)
+                if not SampleNextStepByStudy.objects.filter(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order).exists():
+                    SampleNextStepByStudy.objects.create(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order)
+                elif new_sample.is_pool:
+                    warnings.append(("Sample {0} is already queued for step {1} "
+                                    "of study {2} of project {3}.", [new_sample.name, next_step_order.order if next_step_order is not None else '', study.letter, study.project.name]))
+                else:
+                    errors.append(f"Sample {new_sample.name} is already queued for step {next_step_order.order if next_step_order is not None else ''} "
+                                f"of study {study.letter} of project {study.project.name}.")
```
UlysseFG: The indentation is wrong everywhere ... is this the new normal ?

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +323,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
 
-    return new_sample_next_steps, errors, warnings
+    new_sample = next_sample if next_sample is not None else current_sample
+
+    next_sample_next_step = None
+    study = sample_next_step_by_study.study
+
+    current_step_order = sample_next_step_by_study.step_order
+    if current_step_order.mandatory and workflow_action == WorkflowAction.SKIP_STEP:
+        errors.append(f"Step '{current_step.name}' cannot be skipped in workflow '{study.workflow.name}'.")
```
UlysseFG: this should be a warning...

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +323,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+        return None, errors, warnings
```
UlysseFG: should return the current sample next step

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +323,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+                    else None
+
+    if next_step_order is not None:
+        try:
+            if SampleNextStep.objects.filter(step=next_step_order.step, sample=new_sample).exists():
+                next_sample_next_step = SampleNextStep.objects.get(step=next_step_order.step, sample=new_sample)
+                if not SampleNextStepByStudy.objects.filter(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order).exists():
+                    SampleNextStepByStudy.objects.create(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order)
+                elif new_sample.is_pool:
+                    warnings.append(("Sample {0} is already queued for step {1} "
```
UlysseFG: This message have always been vague ... Maybe we can indicate "Pooled sample {0} is already ... {3} as part of the normal workflow advancement procedure for pools." to clarify that this due the sample pooled status and is normal.

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +323,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+                                sample=current_sample,
+                                workflow_action=workflow_action)
+    except Exception as err:
+        errors.append(f"Failed to create StepHistory.")
+
+    try:
+        if not keep_current:
+            sample_next_step = sample_next_step_by_study.sample_next_step
+            sample_next_step_by_study.delete()
+            if not sample_next_step.studies.exists():
```
UlysseFG: i would prefer to force a DB hit with a fresh queryset (SampleNextstepByStudy.objects.filter(...).exists()) in case down the line we prefetch studies ... The current test could lead to sample next step not being deleted if the query is cached.

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -44,6 +45,45 @@ class SampleNextStepByStudyViewSet(viewsets.ModelViewSet):
...
+        stepOrder = request.data.get("step_order", None)
+        if sample_ids == []:
+            return HttpResponseBadRequest("No sample IDs provided.")
+        if study_id is None:
+            return HttpResponseBadRequest("No study ID provided.")
+        if stepOrder is None:
+            return HttpResponseBadRequest("No step order provided.")
+
+        errors = []
+        warnings = set()
```
UlysseFG: It's weird to have errors as a list and warning as a set here ... is there a good reason to have warning as sets?

backend/fms_core/services/sample_next_step.py
```diff
@@ -123,6 +124,24 @@ def dequeue_sample_from_specific_step_study_workflow(sample_obj: Sample, study_o
...
+            warnings.append(f"Step '{sample_next_step_by_study.step_order.step.name}' cannot be skipped in workflow '{sample_next_step_by_study.study.workflow.name}'.")
+        else:
+            _, errors, warnings = _move_sample_to_next_step_by_study_step(
+                sample_next_step_by_study,
+                workflow_action=WorkflowAction.SKIP_STEP
+            )
+    except Exception as err:
+        errors.append(err)
+    
+    return errors, warnings
```
UlysseFG: I would return the sample next step object to be in line with the move_sample_to_next_step behaviour. You can extract the sample id from that object outside the function when in the viewset to provide to the frontend.

backend/fms_core/viewsets/sample_next_step_by_study.py
```diff
@@ -44,6 +45,45 @@ class SampleNextStepByStudyViewSet(viewsets.ModelViewSet):
...
+
+        queryset = SampleNextStepByStudy.objects.select_related("step_order").select_related("sample_next_step")
+
+        with transaction.atomic():
+            skipped = set[int]()
+            for sample_id in sample_ids:
+                try:
+                    sample_next_step_by_study = queryset.get(sample_next_step__sample__id=sample_id, study=study_id, step_order__order=stepOrder)
```
UlysseFG: this get() in a for loop can likely be replaced by a filter on the queryset ... you can get the sample_id from the object returned by the skip_...(...) call.

backend/fms_core/services/sample_next_step.py
nafiz1001: I might have used Emacs at some point which tends to force format.

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/viewsets/sample_next_step_by_study.py
nafiz1001: There used to be repeating warnings.

backend/fms_core/services/sample_next_step.py
nafiz1001: soccer

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/viewsets/sample_next_step_by_study.py
nafiz1001: done?

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +324,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+    next_step_order = current_step_order.next_step_order \
+                    if current_step_order.next_step_order and current_step_order.next_step_order.order <= study.end \
+                    else None
+
+    if next_step_order is not None:
+        try:
+            if SampleNextStep.objects.filter(step=next_step_order.step, sample=new_sample).exists():
+                next_sample_next_step = SampleNextStep.objects.get(step=next_step_order.step, sample=new_sample)
+                if not SampleNextStepByStudy.objects.filter(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order).exists():
+                    SampleNextStepByStudy.objects.create(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order)
```
UlysseFG, line 389: new_sample_next_step should be set here too.

nafiz1001: Which line are you specifically talking about?

backend/fms_core/services/sample_next_step.py
```diff
@@ -304,62 +324,102 @@ def move_sample_to_next_step(current_step: Step, current_sample: Sample, process
...
+                                     "of study {2} of project {3}.", [new_sample.name, next_step_order.order if next_step_order is not None else '', study.letter, study.project.name]))
+                else:
+                    errors.append(f"Sample {new_sample.name} is already queued for step {next_step_order.order if next_step_order is not None else ''} "
+                                  f"of study {study.letter} of project {study.project.name}.")
+            else:
+                next_sample_next_step = SampleNextStep.objects.create(step=next_step_order.step,
+                                                                    sample=new_sample)
+                if next_sample_next_step is not None:
+                    SampleNextStepByStudy.objects.create(sample_next_step=next_sample_next_step, study=study, step_order=next_step_order)
+                    new_sample_next_step = next_sample_next_step
```
UlysseFG: hint ...

backend/fms_core/services/sample_next_step.py
```diff
@@ -383,7 +383,7 @@ def _move_sample_to_next_step_by_study_step(
     if next_step_order is not None:
         try:
             if SampleNextStep.objects.filter(step=next_step_order.step, sample=new_sample).exists():
-                next_sample_next_step = SampleNextStep.objects.get(step=next_step_order.step, sample=new_sample)
+                resulting_sample_next_step = next_sample_next_step = SampleNextStep.objects.get(step=next_step_order.step, sample=new_sample)
```
UlysseFG: ugh what is this 🥔 ... You are allowed to assign a 🥔 twice  the same way.

backend/fms_core/migrations/0082_v5_9_0.py
nafiz1001: DONE

backend/fms_core/models/step_order.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: I GUESS DONE.

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE?

nafiz1001: DONE?

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE?

backend/fms_core/services/sample_next_step.py
nafiz1001: PASSED

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: DONE

backend/fms_core/services/sample_next_step.py
nafiz1001: ZZZ

nafiz1001:
Ignore the merge conflict for now...

UlysseFG (approved):

PR #1163, opened 2026-06-29
Recommend 1: Adding project search by Hercules project name from External ID results

nabalma:
### Summary

This PR implements Recommendation 1 by making it easier to locate projects associated with an External ID without requiring users to navigate to Hercules.

### Changes
Added an endpoint to retrieve projects grouped by External ID.
Added the ProjectsByExternalIDPage.
Integrated the new API into the Project Overview.
Enabled searching/filtering projects by Hercules project name.
Updated the Project Overview page to use URL parameters for External ID filtering.
Updated links to include Hercules project names when navigating to the project list.

###  Motivation : 
@UlysseFG Request enhanced by Customer feedback below .
Screenshot From 2026-06-29 10-01-35


nafiz1001:
It would make the review easier if you undo some of the unintended changes like the ones from formatOnSave.

nabalma:
> It would make the review easier if you undo some of the unintended changes like the ones from formatOnSave.

Thanks @nafiz1001.  Will let you know.

nabalma:
> > It would make the review easier if you undo some of the unintended changes like the ones from formatOnSave.
> 
> Thanks @nafiz1001. Will let you know.

@nafiz1001 
I reverted the unintended autopep8 formatting changes and disabled format on save.

backend/fms_core/viewsets/project_overview.py
```diff
@@ -14,6 +18,14 @@ def get_external_id_from_request(request):
     return external_id
 
 
+def get_external_id_number(external_id):
+    if not external_id:
+        return None
+
+    match = re.search(r"\d+", external_id)
+    return int(match.group()) if match else None
```
nafiz1001: This might be overkill. The format is consistently PXXX, so we can simply do `int(external_id[1:])` or actually we can even leave it as string `external_id[1:]` to include leading zeros.

backend/fms_core/viewsets/project_overview.py
```diff
@@ -121,3 +130,39 @@ def reads(self, request):
...
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
+
+    @action(detail=False, methods=["get"])
+    def projects_by_external_id(self, request):
+        projects = (
+            Project.objects
+            .filter(deleted=False)
+            .order_by("name", "id")
+        )
```
nafiz1001: - Filtering by `deleted=False` is overkill since projects are basically never deleted.  So it should be removed.
- `.order_by` is also not useful here since sorting will be done on `grouped_projects` after.

Instead, you can do `Project.objects.all()`.

backend/fms_core/viewsets/project_overview.py
nafiz1001, line 1: The way this viewset is implemented can't take advantage of pagination. The number of projects might be small enough that it's not a concern yet, but it's something to consider.

nafiz1001 (commented):
We should avoid using the term hercules since it is a very specific service rather than being general. We should instead use "external project" or something along that line.

backend/fms_core/viewsets/project_overview.py
```diff
@@ -121,3 +130,39 @@ def reads(self, request):
...
+            group_key = project.external_id or ""
+            grouped_projects_map[group_key].append(project)
+
+        grouped_projects = [
+            {
+                "external_id": external_id or None,
+                "external_id_number": get_external_id_number(external_id),
+                "hercules_project_name": projects[0].external_name if projects else None,
+                "project_count": len(projects),
+                "projects": projects,
```
nafiz1001, line 150: `project_count` isn't necessary since you already provide an array of `projects` with known size.

Actually, it might depend on if we plan to have projects under external project id paginated or not. So, for now, we can keep both fields here.

backend/fms_core/viewsets/project_overview.py
nabalma: int(external_id[1:]) applied ...

backend/fms_core/viewsets/project_overview.py
nabalma: Fixed ...

nabalma:
> We should avoid using the term hercules since it is a very specific service rather than being general. We should instead use "external project" or something along that line.


Amended with "external project" in both the frontend and the backend. Many Thanks !

PR #1165, opened 2026-07-07
Fix container search during adding a single sample

nafiz1001:


nafiz1001:
This is a draft because Ulysse doesn't agree with my solution.

UlysseFG:
> This is a draft because Ulysse doesn't agree with my solution.

I actually read a bit more about it and it seem we are not using the fuzy search at the moment.
Your solution is possibly a pure upgrade to the way we do the search now, but it is rigid. 
I wonder if it would not be sufficient and more efficient to actually use the fuzy search. Looking at the query viewset i can see this being used for containers without too much pain ... We would get exact matches as first hits and we would get all other field hits without having to be too rigid with the code.

PR #1166, opened 2026-07-09
Add instrument illumina-miseq-i100

nafiz1001:


backend/fms_core/containers.py
```diff
@@ -20,6 +20,7 @@
     "CONTAINER_SPEC_ILLUMINA_MISEQ_V3_FLOWCELL",
     "CONTAINER_SPEC_ILLUMINA_MISEQ_MICRO_FLOWCELL",
     "CONTAINER_SPEC_ILLUMINA_MISEQ_NANO_FLOWCELL",
+    "CONTAINER_SPEC_ILLUMINA_MISEQ_i100_FLOWCELLS",
```
UlysseFG: Use all caps (i.e. ...`_I100_`) for the variable name.

backend/fms_core/containers.py
```diff
@@ -289,6 +290,17 @@ def __str__(self):
...
 
+CONTAINER_SPEC_ILLUMINA_MISEQ_i100_FLOWCELLS = [
+    ContainerSpec(
+        container_kind_id=f"illumina-miseq-i100-{m}m flowcell",
+        coordinate_spec=(alphas(1), ints(1, pad_to=2)), #1 lane
+        coordinate_overlap_allowed=False,
+        children=(),  # Leaf node; sample-holding
+        is_run_container=True,
+    )
+    for m in [5, 25, 50, 100]
```
UlysseFG, line 301: All of the flowcells are single lane ?

nafiz1001: That's a good question... I wonder if the lab only talked about the 5m container.

UlysseFG, line 301: I am not sure this will not cause errors in the future ... there is no real reason to iterate through them instead of defining them independantly... we'll see...

UlysseFG:
Add the new instrument type to the Pooling template and Sample submission template. They have a dropdown that verify index collisions using the intented instrument type used.

backend/fms_core/containers.py
UlysseFG: I think they are all single lane.

backend/fms_core/migrations/0079_v5_7_1.py
```diff
@@ -0,0 +1,64 @@
+import reversion
```
UlysseFG, line 1: Rename file name to match the typical format.

UlysseFG: This will also force the renaming of the currently merged migration files on dev...

UlysseFG (approved):

PR #1175, opened 2026-07-20
Extract external project from project model.

UlysseFG:


backend/fms_core/models/project.py
```diff
@@ -40,9 +39,6 @@ def clean(self):
         def add_error(field: str, error: str):
             _add_error(errors, field, ValidationError(error))
 
-        if self.external_id == "":
-            add_error("external_id", "Project External ID cannot be empty text.")
-
```
nafiz1001, line 45: So we're allowing creation of project with missing ~~external~~ parent project?

UlysseFG: We keep the current state of things on that side ... We are allowed as admin to create them without any parent project. You kind of proposed to keep and since there is no real issue with that i am keeping this as an option ... The users however will need to provide a external id and name.

The external ID field does not allow blank anymore so that specific validation is no longer needed.

backend/fms_core/migrations/0082_v5_9_0.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # parent projects taken from project model
+        for project_obj in Project.objects.all():
+            if project_obj.external_id and not ParentProject.objects.filter(external_id=project_obj.external_id).exists():
+                parent_project_name = project_obj.external_name or project_obj.name # use project external name in priority and defaults to the project name
+                parent_project_obj = ParentProject.objects.create(external_id=project_obj.external_id,
+                                                                  name=parent_project_name,
+                                                                  created_by_id=admin_user.id,
+                                                                  updated_by_id=admin_user.id)
+            reversion.add_to_revision(parent_project_obj)
```
nafiz1001, line 44: Wouldn't it be more optimal to build a queryset of external project id, loop over each id (leveraging queryset's laziness), and do a `ParentProject.objects.get_or_create`?

backend/fms_core/tests/test_models/dataset_test.py
```diff
@@ -56,15 +58,15 @@ def test_dataset(self):
...
+        self.assertEqual(dataset.project.parent_project.external_id, "P031553")
         self.assertEqual(dataset.experiment_run.name, "test_run")
         self.assertEqual(dataset.lane, 1)
 
     def test_dataset_with_report(self):
         dataset = Dataset.objects.create(project=self.my_project, experiment_run=self.my_experiment_run, lane=1, metric_report_url="https://www.FakeMetricReport.com")
         self.assertEqual(Dataset.objects.count(), 1)
         self.assertEqual(dataset.project.name, "test")
-        self.assertEqual(dataset.project.external_id, "P031553")
+        self.assertEqual(dataset.project.parent_project.external_id, "P031553")
```
nafiz1001, line 69: I'd imagine migration would be simpler if you defined a `@property` for `project.external_id` and maybe even `project.external_name`.

backend/fms_core/tests/test_models/parent_project_test.py
nafiz1001, line 1: This file could be an opportunity to unit test using pytest's approach which is plain functions with `assert` keyword. I already wrote a test like that for rename curation template (`backend/fms_core/tests/test_workbooks/sample_rename_test.py`)

backend/fms_core/tests/test_models/project_test.py
```diff
@@ -91,18 +92,23 @@ def test_project_with_similar_name(self):
...
+        parent_project_obj = ParentProject.objects.create(external_id="P000010", name="TESTS")
+        my_project = Project.objects.create(name=self.name,
+                                            principal_investigator=self.principal_investigator,
+                                            requestor_name=self.requestor_name,
+                                            requestor_email=self.requestor_email,
+                                            targeted_end_date=self.targeted_end_date,
+                                            parent_project=parent_project_obj,
+                                            comment=self.comment)
+        self.assertEqual(my_project.parent_project.external_id, "P000010")
+        self.assertEqual(my_project.parent_project.name, "TESTS")
```
nafiz1001, line 114: You forgot to check for matching foreign key vs. primary key ;)

backend/fms_core/migrations/0082_v5_9_0.py
UlysseFG: This is a one time execution on a very simple table i will not overcomplicate this.

backend/fms_core/tests/test_models/dataset_test.py
UlysseFG: external_id is not a property of project anymore ...

backend/fms_core/tests/test_models/parent_project_test.py
UlysseFG: asserts are one of Python failures ... i would not use them. They are a break with the language syntax and really provide no useful benefice in the testing context. For someone that is so big on standardization this is backward ... By providing no structure to tests, you are opening the door to an anything goes approach... and for test this is bad.

backend/fms_core/tests/test_models/project_test.py
UlysseFG: mmm i think we can skip this one ;).

backend/fms_core/viewsets/project.py
```diff
@@ -36,6 +38,99 @@ class ProjectViewSet(viewsets.ModelViewSet, TemplateActionsMixin):
...
+            project_to_update.save()
+        except Exception as err:
+            raise ValidationError(err)
+
+        # Return updated project
+        # Serialize full project using the created project
+        try:
+            instance = self.get_queryset().get(pk=project_to_update.id)
+            serializer = self.get_serializer_class()(instance, data=project_data)
+            serializer.is_valid(raise_exception=True)
```
nafiz1001, line 121: Can't you serialize `project_to_update` insteaed?

backend/fms_report/tests/test_models/production_data_test.py
```diff
@@ -45,7 +45,8 @@ def setUp(self):
         self.protocol, _ = Protocol.objects.get_or_create(name=self.protocol_name)
         self.process = Process.objects.create(protocol=self.protocol, comment="Process test for ExperimentRun")
 
-        self.project = Project.objects.create(name="MY_NAME_IS_PROJECT", external_id="P031553", principal_investigator="MrPotato")
+        parent_project_obj = ParentProject.objects.create(external_id="P031553", name="ClientProject")
+        self.project = Project.objects.create(name="MY_NAME_IS_PROJECT", principal_investigator="MrPotato", parent_project=parent_project_obj)
```
nafiz1001: This pattern is super-reoccuring. It probably should've been a service like `create_project` with arguments for external id and name.

frontend/src/modules/parentProjects/reducers.ts
nafiz1001, line 1: ~~I'll be honest; I believe we shouldn't be caching with nor use redux in general.~~

~~And if you do believe it's the right thing to do anyway... It's better to do it with [redux toolkit's `createSlice`](https://redux.js.org/introduction/getting-started#basic-example)~~

frontend/src/utils/api.ts
```diff
@@ -149,6 +149,11 @@ const api = {
     getReadsPerSampleForLane: (experimentRunId, lane) => get(`/metrics/`, {limit: 100000, name: 'nb_reads', metric_group: 'qc', readset__dataset__experiment_run_id: experimentRunId, readset__dataset__lane: lane})
   },
 
+  parentProjects: {
+    get: parentProjectId => get(`/parent-projects/${parentProjectId}/`),
```
nafiz1001: No return type for this? 🥺

backend/fms_core/viewsets/project.py
UlysseFG: Yeah most likely this was inspired by the sample viewset but i think the project is close enough to the serialized project to use the project instance.

backend/fms_core/tests/test_models/parent_project_test.py
nafiz1001: Fair enough... I do admit they should be handled by testing modules and not by language syntax.

nafiz1001 (approved):

PR #1178, opened 2026-08-05
Project Menu Update  :  Adding submenus for external and internal projects

nabalma:
Adds a new Projects submenu with separate navigation entries for External Overview and Internal Projects, and introduces the initial External Project Overview page and route.

Screenshot From 2026-08-05 16-07-04


frontend/src/components/app/App.js
```diff
@@ -398,6 +414,11 @@ const App = ({userID, usersByID, logOut }) => {
                 <ExperimentRunsPage />
               </PrivateNavigate>
             } />
+            <Route path="/project-overview/*" element={
```
UlysseFG: "projects-overview" to keep the same plural pattern we have for most menu choices.

frontend/src/components/app/App.js
```diff
@@ -398,6 +414,11 @@ const App = ({userID, usersByID, logOut }) => {
...
+              <PrivateNavigate>
+                <ProjectsByExternalIDPage />
```
UlysseFG: ExternalProjectsPage would be a simpler match for what will be listed in the component.

nabalma: Fixed. Thanks

frontend/src/components/app/App.js
nabalma: I chose /external-projects-overview because the current menu selection logic relies on route prefixes. Using a distinct route prefixe ensures a better user experience.

frontend/src/components/app/App.js
```diff
@@ -175,10 +178,23 @@ const App = ({userID, usersByID, logOut }) => {
...
+              url: "/external-projects-overview",
+              text: "External Overview",
+              key: "external-project-overview",
+            },
+            {
+              icon: <ApartmentOutlined />,
+              url: "/projects",
+              text: "Internal Projects",
+              key: "internal-project",
+            },
```
nafiz1001, line 196: @UlysseFG  I wonder if the lab would be comfortable navigating to a page called "Internal Projects" for freezeman projects. I also think "External Overview" would be a strange name to present to service team; but they most probably are aware of our project name shenanigans.

UlysseFG: well the naming is always tricky if you guys have better ideas for the subsection i am open to it ... we could potentially call "internal projects" straight up "projects" and find a more appealing other name but i don't want to spend too much time on naming at the moment.

UlysseFG:
This might be a bit too partial a PR ... Ideally each PR leads to a stable / acceptable state. This ones leads to a "in construction page". In the future, instead of having this lead to a partial page just do not include the link to the destination page / component. This looks good enough for now ...

UlysseFG (approved):

UlysseFG:
If you merge this we have to make sure to merge the next one too that include the page content...

nabalma:

PR #1179, opened 2026-08-07
External project overview landing page

nabalma:
### Note
Please review #1178 first if you haven't already.

Adds a Parent Projects table displaying the list of external projects . Each external project is described with its external ID, name, and number of associated Freezeman projects. Each row can be expanded to show the corresponding internal projects and their key information. The page also supports filtering Parent Projects by external project name.

image

### DIFF
--- a/frontend/src/components/projectOverview/ExternalProjectsPage.tsx
+++ b/frontend/src/components/projectOverview/ExternalProjectsPage.tsx
@@ -1,7 +1,7 @@
 
 import { Alert, Button, Input, Table, Tag } from 'antd'
 import type { ColumnsType } from 'antd/es/table'
-import React, { useCallback, useEffect, useState } from 'react'
+import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import AppPageHeader from '../AppPageHeader'
 import { Link } from 'react-router-dom'
 import { FMSProject,FMSParentProject } from '../../models/fms_api_models'
@@ -92,7 +92,7 @@ const ExternalProjectsPage = () => {
 
 	const clearFilters = useCallback(() => {setFilters({})}, [])
 
-	const parentProjectColumns: ColumnsType<FMSParentProject> = [
+	const parentProjectColumns =useMemo<ColumnsType<FMSParentProject>>(()=> [
 		{
 			title: 'External Project ID',
 			dataIndex: 'external_id',
@@ -169,7 +169,7 @@ const ExternalProjectsPage = () => {
 				)
 			},
 		},
-	]
+	], [filters, setFilters])
 
 	const dispatch = useAppDispatch()
 
@@ -256,9 +256,16 @@ const ExternalProjectsPage = () => {
 					loading={isLoading}
 					
 					expandable={{
-    					expandedRowRender: (parentProject) => {const internalProjects = (parentProject.projects ?? [])
-            				.map((projectID) => internalProjectsByID[projectID])
-							.filter((project): project is FMSProject =>project !== undefined)
+    					expandedRowRender: (parentProject) => {
+							const internalProjects = (parentProject.projects ?? []).reduce<FMSProject[]>(
+								(projects, projectID) => {
+									const project = internalProjectsByID[projectID]
+									if (project) {
+										projects.push(project)
+									}
+									return projects
+								},[]
+							)
 							return (
 								<Table
 									size="small"
@@ -268,7 +275,7 @@ const ExternalProjectsPage = () => {
 									pagination={false}
 								/>
 							)
-                           },
+                        },
                     }}
 					pagination={{
 						pageSize: 20,
--- a/frontend/src/components/projectOverview/ExternalProjectsPage.tsx
+++ b/frontend/src/components/projectOverview/ExternalProjectsPage.tsx
@@ -62,6 +62,25 @@ const internalProjectColumns: ColumnsType<FMSProject> = [
 	},
 ]
 
+// Organise les projets internes dans un objet afin de pouvoir retrouver rapidement chaque projet à partir de son ID.
+const indexProjectsByID = (projects: FMSProject[]): Partial<Record<number, FMSProject>> => 
+	projects.reduce<Partial<Record<number, FMSProject>>>((projectsByID, project) => {
+        projectsByID[project.id] = project
+        return projectsByID
+    }, {})
+
+
+// Récupère les IDs uniques des projets internes associés aux projets externes.
+const getUniqueInternalProjectIDs = (parentProjects: FMSParentProject[]): number[] => [
+    ...new Set(
+        parentProjects.flatMap(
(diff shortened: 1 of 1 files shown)


frontend/src/components/app/App.js
nafiz1001, line 417: Changes to this file should be backported to the #1178 PR since nothing new is being added (just corrections on changes made by the first PR)

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -1,19 +1,286 @@
...
+	const [internalProjectsByID, setInternalProjectsByID] = useState<Partial<Record<number, FMSProject>>>({})
+
+	const [isLoading, setIsLoading] = useState(false)
+	const [error, setError] = useState<string | null>(null)
+	const [filters, setFilters] = useState<FilterSet>({})
+
+	const clearFilters = useCallback(() => {setFilters({})}, [])
+
+	const parentProjectColumns: ColumnsType<FMSParentProject> = [
+		{
```
nafiz1001, line 96: parentProjectColumns should be in a `useMemo` like

```ts
const parentProjectColumns = useMemo<ColumnsType<FMSParentProject>>(() => [...], [...])
```

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -1,19 +1,286 @@
...
+					bordered
+					rowKey="id"
+					dataSource={parentProjects}
+					columns={parentProjectColumns}
+					loading={isLoading}
+					
+					expandable={{
+    					expandedRowRender: (parentProject) => {const internalProjects = (parentProject.projects ?? [])
+            				.map((projectID) => internalProjectsByID[projectID])
+							.filter((project): project is FMSProject =>project !== undefined)
```
nafiz1001:
```ts
(project): project is FMSProject =>project !== undefined
```
That's a really fascinating syntax. I was confused at first as to what it is. But now I get it.

```ts
expandedRowRender: (parentProject) => {const internalProjects = (parentProject.projects ?? [])
```
The `const internalProjects` should be on another line.

Lastly,
I believe it's better to use `reduce` than `map`+`filter`. (javascript sadly doesn't have `mapFilter` or `filterMap`)

frontend/src/components/app/App.js
nabalma: Done

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
nabalma: Fixed

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -1,19 +1,293 @@
...
+			key: 'external_project_name',
+			filteredValue: filters[EXTERNAL_PROJECT_NAME_FILTER_KEY]?.value ? [String(filters[EXTERNAL_PROJECT_NAME_FILTER_KEY].value)] : null,
+			filterIcon: (filtered: boolean) => <SearchOutlined style={{ color: filtered ? '#1890ff' : undefined }} />,
+			filterDropdown: ({ setSelectedKeys, selectedKeys, confirm, clearFilters }) => (
+				<div style={{ padding: 8 }}>
+					<Input
+						placeholder="Search External project name"
+						value={selectedKeys[0]}
+						onChange={(event) => setSelectedKeys(event.target.value ? [event.target.value] : [])}
+						onPressEnter={() => confirm()}
```
UlysseFG: Theres is discrepancy in the way the search behave compared to the rest of the app. Your approach is not bad but it clashes with other pages.
1. Typically on pressing the search magnifying glass icon, the focus is given to the input field in order to allow direct entry instead of having to manually giving the focus to the fields first.
2. Most (if not all) search in the app will perform a debounced search as the user type the search content in and a search is also performed on pressing enter before closing the modal window.
3. A search on the external project ID would be good to make a quick search in case the user knows the ID.
Matching the other pages behaviour would make it more natural to the user.

UlysseFG:
This is starting to shape up interestingly. I think in the future we should move the PI to the parent project in order to be able to filter by it... This can be done later.

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -1,19 +1,293 @@
...
+
+	const clearFilters = useCallback(() => {setFilters({})}, [])
+
+	const parentProjectColumns =useMemo<ColumnsType<FMSParentProject>>(()=> [
+		{
+			title: 'External Project ID',
+			dataIndex: 'external_id',
+			key: 'external_id',
+			width: 120,
+			render: (externalID: string, parentProject: FMSParentProject) => (<Link to={`/external-projects-overview/${parentProject.id}#submissions`}>{externalID}</Link>)
```
UlysseFG: This link leads to nowhere yet. Keep each PR centered around itself. Make it so the link does not appear here unless there is a working destination. If you build a PR on top of the stack of PR you can add the link to the child PR once you define the destination. We want to be able to merge each PR being confident the app is self contained without having to worry about dead links and dangling parts.

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -1,19 +1,293 @@
...
+	}, [fetchParentProjectsWithInternalProjects])
+
+	
 	return (
 		<>
 			<AppPageHeader title="Projects by External ID" />
 
-			
+			 <PageContent>
+				{error && (<Alert type="error" title={error} style={{ marginBottom: 16 }} />)}
```
UlysseFG: I am not sure this would be very useful. The page would not render most likely if the backend was down. The actual information provided is very limited. I would tend to go for simplicity (removing error state) here given the small value of providing this error message to the user.

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
UlysseFG: what does `#submissions` points to ? Looks like it is left over from an older version.

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
nabalma: Noted, with Thanks. Working in progress ....

nabalma: Points 1,2,3 are fixed ... (using getFilterPropsForDescription)
image

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
nabalma: Fixed. I removed the link because the target details page is not part of this PR.

nabalma:

PR #1180, opened 2026-08-12
External project details submissions tab

nabalma:
### Note:
Please review PR #1179 first, if you haven't already.

This PR adds the External Project Details page and the Associated Freezeman Projects tab. It introduces navigation from the External Projects landing page via the Parent Project ID, loads the selected Parent Project along with its associated internal projects, and displays them in a paginated table with CSV export support.

Screenshot From 2026-08-12 08-16-57


frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+import FlexBar from '../shared/Flexbar'
+
+import React, { useCallback, useState, useEffect} from 'react'
+
+import useHashURL from '../../hooks/useHashURL'
+import AppPageHeader from '../AppPageHeader'
+import PageContent from '../PageContent'
+
+import ProjectSubmissionsTab from './ProjectSubmissionsTab'
+// import ProjectSamplesTab from './ProjectSamplesTab'
```
UlysseFG: Do not keep future or old code commented out. If you have an open PR for these move the code to these but keep each PR as it's own thing as if this was to be the final version.

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+            finally {
+                setIsLoading(false)
+            }
+        },[fetchParentProject,fetchInternalProjectsByIDs,],
+    )
+
+    useEffect(() => {
+        if (parentProjectId === null) {
```
UlysseFG, line 99: I am not sure  that would be the best way to deal with the situation ... I would instead return the user to the parent project list so they can choose one project. This situation should never happen under normal circumstances and if it does displaying an error is likely not as useful as being rerouted to the list.

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+            setError('Invalid parent project ID')
+            return
+        }
+
+	    fetchParentProjectWithInternalProjects(parentProjectId)
+
+    }, [parentProjectId,fetchParentProjectWithInternalProjects,])
+
+
+    const [activeKey, setActiveKey] = useHashURL('submissions')
```
UlysseFG: This hash name is weird. there is no real submission on this page as far as i know...

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+    return (
+        <>
+            <AppPageHeader title={parentProject ? `Project Overview : (External ID ${externalID})`: 'Project Overview'}/>
+
+            <PageContent tabs>
+
+                {error && (	<Alert type="error"	title={error}	style={{ marginBottom: 16 }} />)}
+
+                <FlexBar style={{ alignItems: 'center', justifyContent: 'flex-start', gap: 8, marginBottom: 16 }}>
+                    <Typography.Text strong style={{ fontSize: 18 }}>{parentProject?.name}</Typography.Text>
```
UlysseFG: The parent project name should be in the app page header instead of the external ID. You can keep the external ID too. If the external project name is too long you can use an ellipse on the name to prevent the header from wrapping. The current Flexbar reduces the space available on the page.

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+                            key: 'submissions',
+                            children: (
+                               <ProjectSubmissionsTab
+                                    internalProjects={internalProjects}
+                                    isLoading={isLoading}
+                                    externalID={externalID}
+                                />
+                            ),
+                        },
+                        // {
```
UlysseFG: keep only what is in the scope of the PR...

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+                <Tabs
+                    activeKey={activeKey}
+                    onChange={setActiveKey}
+                    size="large"
+                    type="card"
+                    items={[
+                        {
+                            label: 'Associated Freezeman Projects',
+                            key: 'submissions',
```
UlysseFG: Not a big fan of the submissions name here again...

frontend/src/components/projectOverview/ProjectLibrariesTab.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
```
UlysseFG: Keep this clean of future stuff... remove those tab stubs.

frontend/src/components/projectOverview/ProjectSubmissionsTab.tsx
```diff
@@ -0,0 +1,132 @@
...
+			external_id: project.external_id ?? '',
+			principal_investigator: project.principal_investigator,
+			requestor_name: project.requestor_name,
+			status: project.status,
+			created_at: project.created_at,
+		})),
+	[internalProjects],
+	)
+
+	const generateCsvContent = useCreateCsvExportFunction(exportProjects)
```
UlysseFG, line 86: Exporting data as a csv would normally rely on the backend... This works as long as there is few projects but at some point it might break down ... I need to think this over so don't go rewriting this just yet but this will not work for the other tab where we'll want to filtering ...

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
nabalma: Agreed.  Fixed...

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
nabalma: Fixed

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
nabalma: Agreed. I renamed the hash from submissions to projects to match the actual content of the tab.

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
nabalma: Fixed. 
The parent project name is now displayed in the app page header, with the external ID. Long project names are truncated with an ellipsis as well.

frontend/src/components/projectOverview/ExternalProjectDetailsPage.tsx
nabalma: replaced by projects instead.

image

frontend/src/components/projectOverview/ProjectSubmissionsTab.tsx
nabalma: Understood. I’ll keep the current frontend export for now. We can revisit a backend-based export later when appropriate.

nabalma:

PR #1181, opened 2026-08-17
External project details readsets tab

nabalma:
### Note : 
Please review PR #1180  first, if you haven't already.

This PR adds the Readsets tab. It follows the Project Submissions tab PR.


nafiz1001:
You might have forgotten to add a dependency to package.json

```
ERROR in ./src/components/projectOverview/ExternalIDReadSetDashboard.tsx 17:0-44
Module not found: Error: Can't resolve '@ant-design/charts' in '/home/user/projects/freezeman/frontend/src/components/projectOverview'
```

Fixed with `npm install @ant-design/charts`

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+    "dataset__experiment_run__run_type__platform__deleted": False,
+    "derived_sample__deleted": False,
+    "derived_sample__biosample__deleted": False,
+    "derived_sample__biosample__individual__deleted": False,
+    "derived_sample__library__deleted": False,
+    "derived_sample__library__library_type__deleted": False,
+    "derived_sample__derived_by_samples__deleted": False,
+    "derived_sample__derived_by_samples__sample__deleted": False,
+    "derived_sample__derived_by_samples__sample__container__deleted": False,
+}
```
nafiz1001: I would call this `DEFAULT_PARENT_PROJECT_READSET_FILTERS`

nafiz1001: And I also believe this should be moved inside `get_parent_project_readsets_queryset`

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+    "run_name": F("dataset__experiment_run__name"),
+    "lane": F("dataset__lane"),
+    "reference_genome_id": F("derived_sample__biosample__individual__reference_genome_id"),
+    "sequencing_index_name": F("derived_sample__library__index__name"),
+    "run_start_date": F("dataset__experiment_run__start_date"),
+    "alias": F("derived_sample__biosample__alias"),
+    "cohort": F("derived_sample__biosample__individual__cohort"),
+    "library_type": F("derived_sample__library__library_type__name"),
+    "number_of_reads": F("production_data__reads"),
+}
```
nafiz1001: These should be moved inside `get_parent_project_readsets_queryset` unless they have a use outside the function.

backend/fms_core/queries/parent_project/readsets.py
nafiz1001, line 1: Having codes in `backend/fms_core/queries/` is stepping outside the usual pattern for creating queryset in Freezeman.

nafiz1001: They are normally achieved with services or viewsets or filters.py or serializers.py

backend/fms_core/queries/parent_project/__init__.py
nafiz1001, line 1: Brand new submodules shouldn't be introduced without ~~master's permission~~ @UlysseFG's agreement.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+    TeamOutlined,
+} from '@ant-design/icons'
+import { Column } from '@ant-design/charts'
+import { ProjectOverviewReadset } from './types'
+
+const { Text } = Typography
+
+const getQcCompletenessData = (items: ProjectOverviewReadset[]) => {
+    const total = items.length || 1
+
```
nafiz1001: It's strange to always be needing at least total of 1.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+    }).length
+
+    const incomplete = items.length - complete
+
+    return {
+        complete: Math.round((complete / total) * 100),
+        incomplete: Math.round((incomplete / total) * 100),
+        completeCount: complete,
+        incompleteCount: incomplete,
+    }
```
nafiz1001, line 37: Ah, that's why you wanted total of at least 1. I feel like it would be better instead to do something like
```ts
        complete: Math.round((complete / Math.max(total, 1)) * 100),
        incomplete: Math.round((incomplete / Math.max(total, 1)) * 100),
```

or

```ts
        complete: total == 0 ? 0 : Math.round((complete / total) * 100),
        incomplete: total == 0 ? 0 : Math.round((incomplete / total) * 100),
```

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+}
+
+function ExternalIDReadSetDashboard({ readsets }: { readsets: ProjectOverviewReadset[] }) {
+    const metrics = useMemo(() => {
+        const total = readsets.length || 1
+
```
nafiz1001: Similar suggestion I've made for `frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx`

frontend/src/components/projectOverview/ExternalProjectsPage.tsx
nafiz1001, line 1: This is corrections that were supposed to be done on the PR that introduced those code.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+				.filter((libraryType): libraryType is string => Boolean(libraryType)),
+		),
+	).map((libraryType) => ({
+		text: libraryType,
+		value: libraryType,
+	}))
+
+	const projectOverviewReadsetColumns = getProjectOverviewReadsetColumns(libraryTypeFilters)
+
+	if (isLoading) {
```
nafiz1001, line 485: `getProjectOverviewReadsetColumns` result should be memoized.

backend/fms_core/queries/parent_project/readsets.py
UlysseFG: A small note : The deleted flag is typically used to save a last version in reversion before deleting an object for good. This means there should be no "deleted" objects in the DB except in the reversion tables. These filter will likely slow the query without much to offer.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+        const validAlignments = readsets
+            .map((x) => x.pf_reads_aligned)
+            .filter((value) => value !== null && value !== undefined)
+            .map(Number)
+
+        return {
+            totalReadsets: readsets.length,
+            totalReads: readsets.reduce((sum, x) => sum + Number(x.number_of_reads || 0), 0),
+            totalRuns: new Set(readsets.map((x) => x.run_name)).size,
+            totalSamples: new Set(readsets.map((x) => x.readset_sample_name)).size,
```
UlysseFG: sample name is not unique. You should not trust it for counting independant samples. You would be better to count biosample_id. It is likely to be mostly the same but there might be case where this will make a difference.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+            totalCohorts: new Set(readsets.map((x) => x.cohort)).size,
```
UlysseFG, line 169: This is a metric that will likely not be very useful. Projects usually don't have many cohorts and the number of cohort is not an indicator for much of anything. You can leave it, but keep in mind if you ever need to free some space on the UI that this is a good candidate for replacement.

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+            ),
+            duplicate_aligned=Max(
+                "metrics__value_numeric",
+                filter=Q(metrics__name="duplicate_rate",metrics__deleted=False,),
+            ),
+            readset_file_paths=ArrayAgg(
+                "files__file_path",
+                filter=Q(files__deleted=False),distinct=True,default=Value([]),
+            ),
+            readset_file_sizes=ArrayAgg(
```
UlysseFG: It would be better to combine each file with its size in a single cell instead of having to rely on matching position in a list. I admit this would not render well in a CSV but the list are likely to break the regularity of the structure.

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+                "files__size",
+                filter=Q(files__deleted=False),distinct=True,default=Value([]),
+            ),
+            barcodes=ArrayAgg(
+                "derived_sample__derived_by_samples__sample__container__barcode",
```
UlysseFG: I think there might still be some confusion between sample barcodes and container barcodes ... I would be very surprised we have to report container barcode anywhere in the project overview unless it is run container barcodes... At the readset level i think the barcode you might have seen in the original report on Nanuq would map to the index (index is sometime called barcode by some people making this more confusing than it needs to be ;).

UlysseFG: you have the library on `derived_sample__library__index__name` and the sequences is available also at `derived_sample__library__index__sequences_3prime` and `derived_sample__library__index__sequences_5prime`

backend/fms_core/queries/parent_project/readsets.py
UlysseFG: Nafiz is right, but i honestly find your solution tempting ... The main downside i find to it is that these queries are likely not going to be reused outside a specific viewset and in the end it would just force code reader to move back and forth between code sections (one of the things i dislike the most about our frontend ;). I would tend to recommend we move these query within the viewset itself.

backend/fms_core/viewsets/parent_project.py
```diff
@@ -21,4 +26,22 @@ class ParentProjectViewSet(viewsets.ModelViewSet):
         **_parent_project_filterset_fields,
     }
 
-    ordering = ["external_id"]
\ No newline at end of file
+    ordering = ["external_id"]
+
+
+    @action(detail=True, methods=["get"])
+    def readsets(self, request, pk=None):
```
UlysseFG: maybe call this overview_readsets to reduce the chance people might confuse what is sent back to normal readsets.

backend/fms_core/serializers.py
```diff
@@ -616,6 +618,44 @@ class Meta:
...
+    )
+
+    duplicate_aligned = serializers.DecimalField(
+        max_digits=40,
+        decimal_places=20,
+        allow_null=True,
+    )
+    readset_file_paths = serializers.ListField(required=False)
+    readset_file_sizes = serializers.ListField(required=False)
+    validation_status = serializers.IntegerField(allow_null=True,)
```
UlysseFG: Maybe call this run_validation_status instead to make it clear it is the flag that is assigned when validating the run. The library tab and sample will likely get a validation/qc_status flag also that will be differ from this one.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }, [readsets])
+
+    const qcCompleteness = useMemo(() => getQcCompletenessData(readsets), [readsets])
+
+    const libraryTypeData = useMemo(() => {
+        const grouped = new Map<string, number>()
+
+        readsets.forEach((item) => {
+            const libraryType = item.library_type?.trim() || 'Non renseigné'
```
UlysseFG: Might be better not to mix language in there. This is something that should essentially never happen (all sequenced sample should have a library_type... This may change but this would essentially be something wrong that would need investigating on our part.). For now you can use "unknown" or "sample without library".

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
UlysseFG: I support this approach it is less risky if we ever decide to use total elsewhere. 0 is also a very valid value when not used as a denominator... All projects will have 0 readset at some point.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+            .filter((value) => value !== null && value !== undefined)
+            .map(Number)
+
+        return {
+            totalReadsets: readsets.length,
+            totalReads: readsets.reduce((sum, x) => sum + Number(x.number_of_reads || 0), 0),
+            totalRuns: new Set(readsets.map((x) => x.run_name)).size,
+            totalSamples: new Set(readsets.map((x) => x.readset_sample_name)).size,
+            totalCohorts: new Set(readsets.map((x) => x.cohort)).size,
+            avgQuality: readsets.reduce((sum, x) => sum + Number(x.average_quality), 0) / total,
```
UlysseFG: Averaging averages based on a variable number of bases is likely to skew the value... The correct way would be to use a weighted average using the number of bases for each readset . The number of bases is typically stored on the `yield` metrics field. I think some run do not provide this information you can assume an equal number of bases in this case.

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+    "external_id": F("dataset__project__parent_project__external_id"),
+    "run_name": F("dataset__experiment_run__name"),
+    "lane": F("dataset__lane"),
+    "reference_genome_id": F("derived_sample__biosample__individual__reference_genome_id"),
+    "sequencing_index_name": F("derived_sample__library__index__name"),
+    "run_start_date": F("dataset__experiment_run__start_date"),
+    "alias": F("derived_sample__biosample__alias"),
+    "cohort": F("derived_sample__biosample__individual__cohort"),
+    "library_type": F("derived_sample__library__library_type__name"),
+    "number_of_reads": F("production_data__reads"),
```
UlysseFG: This is not the best place to fetch the reads ... The production_data is extracted only once a run is validated and is not live... It is aggregated daily making this a bad source unless you want the have the reads that passed validation only... In our case, the source should be the readset metric `nb_reads`.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+            avgAlignment:
```
UlysseFG: For alignment and duplication you are averaging averages again ... ideally we would do a weighted average using the number of reads of the readset.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+                <Col xs={24} sm={12} lg={6} xl={3}>
+                    <Card size="small" styles={{ body: { padding: '8px 12px' } }}>
+                        <Statistic title="Samples" value={metrics.totalSamples} prefix={<TeamOutlined style={iconStyle('#13c2c2', '#e6fffb')} />} />
+                    </Card>
+                </Col>
+
+                <Col xs={24} sm={12} lg={6} xl={3}>
+                    <Card size="small" styles={{ body: { padding: '8px 12px' } }}>
+                        <Statistic title="Cohortes" value={metrics.totalCohorts} />
```
UlysseFG: `Cohorts`

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+
+			{projectOverviewReadsets.length > 0 ? (
+				<Table
+					dataSource={projectOverviewReadsets}
+					columns={projectOverviewReadsetColumns}
+					rowKey="id"
+					size="small"
+					bordered
+					scroll={{ x: 'max-content', y: 400 }}
+					pagination={{
```
UlysseFG, line 517: It would be great to have the number of hits listed in the pagination.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+	},
+	{
+		title: 'Library Type',
+		dataIndex: 'library_type',
+		key: 'library_type',
+		width: 140,
+		onHeaderCell: compactHeaderCell,
+		filters: libraryTypeFilters,
+		filterIcon: (filtered) => <FilterOutlined style={{ color: filtered ? '#1677ff' : undefined }} />,
+		onFilter: (value, record) => record.library_type === value,
```
UlysseFG, line 167: Nice touch that filter.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+		title: 'Readset',
+		dataIndex: 'name',
+		key: 'name',
+		//fixed: 'left',
+		width: 450,
+		onHeaderCell: compactHeaderCell,
+		render: (name: string) => <Text strong>{name}</Text>,
+	},
+	{
+		title: 'Sample',
```
UlysseFG, line 85: Would be great to have a filter on that column too.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+	}
+
+	return (
+		<>
+			{!isLoading && projectOverviewReadsets.length > 0 && (
+				<div style={{ display: 'flex', justifyContent: 'flex-end', marginBottom: 8 }}>
+					<ProjectOverviewExportButton data={exportButtonData} />
+				</div>
+			)}
+			{!isLoading && isActive && <ExternalIDReadSetDashboard readsets={projectOverviewReadsets} />}
```
UlysseFG: I would like to test having the dashboard at the very top of the tab with the export button just above the table... This would allow for a clear filters button to be inserted also next to the export button.

backend/fms_core/queries/parent_project/readsets.py
UlysseFG: Seeing the container barcodes there might actually be of some use if there is someone investigating the readset to check if there was some mixup ... It would be important though to mention that this is container_barcodes.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+			</div>
+		),
+
+		onFilter: (value, record) =>
+			String(record.run_name ?? '')
+				.toLowerCase()
+				.includes(String(value).toLowerCase()),
+	},
+	{
+		title: 'Run Start',
```
UlysseFG, line 210: A filter here might be nice...

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+    "dataset__experiment_run__name",
+    "dataset__lane",
+    "id",
+]
+
+PARENT_PROJECT_READSET_VALUE_FIELDS = [
+    "id",
+    "name",
+    "average_quality",
+    "pf_reads_aligned",
```
UlysseFG: There is quite a few other metric fields that are typically available... you can check in schema_validators.py for RUN_PROCESSING_SCHEMA for the list of expected metrics... Don't implement all of them (some are nearly duplicates of each others), but we might do a quick pass to see if we can't push some additional metrics once we are happy with the general thing.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
```diff
@@ -1,11 +1,381 @@
...
+		return <Spin />
+	}
+
+	if (error) {
+		return <Alert type="error" title={error} showIcon />
+	}
+
+	const exportButtonData: ProjectOverviewExportButtonData = {
+		exportType: 'Project Readsets',
+		exportFunction: generateCsvContent,
```
UlysseFG, line 495: Same comment as for the project... we typically send a request to the backend to export only the currently select items... Your csv builder is much more efficient than the package we are using ... I tested the big project MOH-YR5_CRCHUM ... The initial load time is quite large but the frontend filtering and export are much faster that way ... I am afraid that this might lead to some projects being undisplayable because the request times out... We will be adding more metrics to the query and this will slow down the call and some projects can run for years with potentially thousands of readsets ... Maybe it would be better to consider the overview_readsets to have its own endpoint to be able to use a similar behaviour as elsewhere. I am however curious if we could not improve our csv generation on the backend side.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+        borderRadius: 8,
+        marginRight: 4,
+    })
+
+    return (
+        <div style={{ background: '#f5f7fb', padding: 0 }}>
+            <Row gutter={[16, 16]} style={{ margin: 8, borderRadius: 4 }}>
+                <Col xs={24} sm={12} lg={6} xl={3}>
+                    <Card size="small" styles={{ body: { padding: '8px 12px' } }}>
+                        <Statistic title="Readsets" value={metrics.totalReadsets} prefix={<ExperimentOutlined style={iconStyle('#1677ff', '#e6f4ff')}/>} />
```
UlysseFG: Numbers displayed should not use the `,` to separate thousands. You can use a space instead.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+                <Col xs={24} sm={12} lg={6} xl={3}>
+                    <Card size="small" styles={{ body: { padding: '8px 12px' } }}>
+                        <Statistic title="Runs" value={metrics.totalRuns} prefix={<ClusterOutlined style={iconStyle('#722ed1', '#f9f0ff')} />} />
+                    </Card>
+                </Col>
+
+                <Col xs={24} sm={12} lg={6} xl={3}>
+                    <Card size="small" styles={{ body: { padding: '8px 12px' } }}>
+                        <Statistic title="Samples" value={metrics.totalSamples} prefix={<TeamOutlined style={iconStyle('#13c2c2', '#e6fffb')} />} />
```
UlysseFG: This will likely go to the thousands too so you should also replace the `,`.

backend/fms_core/queries/parent_project/readsets.py
```diff
@@ -0,0 +1,95 @@
...
+    "barcodes",
+    "validation_status",
+]
+
+PARENT_PROJECT_READSET_VALUE_ALIASES = {
+    "readset_sample_name": F("sample_name"),
+    "external_id": F("dataset__project__parent_project__external_id"),
+    "run_name": F("dataset__experiment_run__name"),
+    "lane": F("dataset__lane"),
+    "reference_genome_id": F("derived_sample__biosample__individual__reference_genome_id"),
```
UlysseFG: This is not really informative without getting the `assembly_name`. This would need to be fetched separately from the store on the frontend or directly sent back (`derived_sample__biosample__individual__reference_genome__assembly_name`).

backend/fms_core/queries/parent_project/readsets.py
nabalma: Thanks. I removed the redundant deleted=False filters ... Fixed.

backend/fms_core/queries/parent_project/readsets.py
nabalma: Fixed ..

backend/fms_core/queries/parent_project/readsets.py
nabalma: Agreed. I moved the parent project readsets queryset logic into ParentProjectViewSet as a private helper, since it is specific to this action.

backend/fms_core/queries/parent_project/__init__.py
nabalma: Thanks .. 
Addressed by removing the new queries package and moving the action-specific logic into ParentProjectViewSet.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Updated to keep the actual total and protect the percentage calculations when total is 0.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Also fixed ....

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: Fixed ...

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Thanks a lot ... This is fixed ....

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Noted ...

backend/fms_core/queries/parent_project/readsets.py
nabalma: Updated to return and render readset files as a single list of objects containing both file_path and size,

backend/fms_core/queries/parent_project/readsets.py
nabalma: Updated to keep the barcode information and expose it as container_barcodes .

backend/fms_core/viewsets/parent_project.py
nabalma: action renamed 
image

backend/fms_core/serializers.py
nabalma: Done

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: fixed

backend/fms_core/queries/parent_project/readsets.py
nabalma: Fixed. number_of_reads is now retrieved from the readset nb_reads metric instead of production_data.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Fixed

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: addressed...

image

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: Thanks!

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: Added

nafiz1001:
When I exported on http://localhost:9000/external-projects-overview/193#readsets, the readset_files column shows up like this in the csv:

duplicate_aligned | readset_files | run_validation_status
-- | -- | --
0.133361 | [object Object],[object Object] | 1
0.414306 | [object Object],[object Object] | 1
0.136845 | [object Object],[object Object] | 1
0.40548 | [object Object],[object Object] | 1


Thanks. Fixed now ...
image

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Fixed. I updated the average quality calculation to weight each readset by its yield (number of bases). The calculation falls back to an unweighted average when that information is incomplete.

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Fixed. I updated both the alignment and duplication calculations to use weighted averages based on each readset’s number of reads.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: Updated. The dashboard is now displayed at the top of the tab, with the export button above the table. 

image

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: Date range filter added to the Run Start column.

image

backend/fms_core/queries/parent_project/readsets.py
nabalma: Noted . Thanks,
For coming steps, I’ll do a quick pass through the metrics available in RUN_PROCESSING_SCHEMA and submit a shortlist for your review.

frontend/src/components/projectOverview/ProjectReadSetsTab.tsx
nabalma: That makes sense. As discussed, could we address this in a follow-up ?

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Fixed ...Numeric values now use spaces instead of commas as thousands separators in both the dashboard statistics and the readsets table.
image

frontend/src/components/projectOverview/ExternalIDReadSetDashboard.tsx
nabalma: Fixed ...
image

backend/fms_core/queries/parent_project/readsets.py
nabalma: Updated. The backend now returns the reference genome assembly name

nabalma:

PR #1188, opened 2026-08-31
Create a simplified workflow that does not use libraries.

UlysseFG:


backend/fms_core/migrations/0082_v5_9_0.py
```diff
@@ -60,6 +61,53 @@ def populate_foreign_key_to_parent_project(apps, schema_editor):
...
+    Step = apps.get_model("fms_core", "Step")
+
+    with reversion.create_revision(manage_manually=True):
+        admin_user = get_user_model().objects.get(username=ADMIN_USERNAME)
+        reversion.set_comment(f"Allow Experiment Run Illumina step to take extracted samples instead of only libraries.")
+        reversion.set_user(admin_user)
+
+        step = Step.objects.get(name="Experiment Run Illumina")
+        step.expected_sample_type = SampleType.EXTRACTED_SAMPLE # This will result in samples being accepted during experiment run along libraries
+        step.save()
```
nafiz1001, line 94: Wouldn't this lead to excluding libraries from being submitted to Experiment Run Illumina?

nafiz1001:
```python
        elif sample_type == SampleType.EXTRACTED_SAMPLE:
            return self.is_kind_extracted and not self.is_library
```

backend/fms_core/services/sample.py
nafiz1001, line 1: I feel like `refresh_from_db()` should be done by functions calling `transfer_sample` or `extract_sample` or `extract_sample`.

backend/fms_core/migrations/0082_v5_9_0.py
UlysseFG: Mmm looks like i assumed it was something else ... I will fix that ...

backend/fms_core/services/sample.py
UlysseFG: yeah and no. For performance it would be better but for robustness and coding logic it would not. The caller have no idea how the volume or any other things are managed internally. This makes it illogical to ask for a refresh outside the internal function... By doing additional reading i read that save() should update the in-memory instance unless the reference is not the same across calls ... i will check if i can fix this by passing the exact same object reference.

backend/fms_core/services/experiment_run.py
nafiz1001, line 1: My condolences to you.

backend/fms_core/services/experiment_run.py
```diff
@@ -102,15 +106,62 @@ def create_experiment_run(experiment_run_name,
...
-                                                                                     execution_date=start_date,
-                                                                                     coordinates_destination=container_coordinates,
-                                                                                     volume_destination=volume_destination,
-                                                                                     comment=comment,
-                                                                                     workflow=workflow)
+            study_obj = None
+            if workflow is not None:
+                try:
+                    study_obj = Study.objects.get(sample_next_step_by_study__sample_next_step__sample=source_sample,
+                                                  sample_next_step_by_study__sample_next_step__step=workflow["step"])
```
nafiz1001: I just realized that `sample_next_step_by_study` is singular, but I feel it should've been plural...

nafiz1001: But yeah, I presume there will always be at most one study per combination of sample and step, correct?

nafiz1001 (approved):
Running experiment run has gotten so much easier 😍

backend/fms_core/services/experiment_run.py
UlysseFG: Nope, there is the non zero chance we will eventually have a second type of library-without-library and that we will run both on a single sample in a single run ...

UlysseFG: This is the first draft i came up with that can satisfy the immediate need, but i would like to implement a generic full proof solution ... Maybe now maybe next sprint.

PR #1191, opened 2026-09-04
Add merge_group for checks_requested in workflow

nafiz1001:
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group


UlysseFG (approved):
I really don't know how this will end up working but go ahead we'll tweak. Ideally only the backend test are on the merge group the other checks can be done on push ...

PR #1196, opened 2026-09-18
Provide error message when missing external project name.

UlysseFG:


UlysseFG:
Made the mistake of correcting a single indent on a file i checked ... ignore all changes to ProjectsListContent.tsx it is just prettier stuff.

nafiz1001 (approved):

PR #1197, opened 2026-09-19
Address feedback for the Parents Overview navigation and scrolling

nabalma:
### Summary
- Renamed and reordered the Projects submenu items:
  - Projects
  - Parents Overview
- Updated the Parents Overview table so its content scrolls within the table frame.

image


frontend/src/components/projectOverview/ExternalProjectsPage.tsx
```diff
@@ -251,7 +251,7 @@ const ExternalProjectsPage = () => {
           dataSource={parentProjects}
           columns={parentProjectColumns}
           loading={isLoading}
-
+          scroll={{ y: "70vh" }}
```
UlysseFG: There is the same issue with scrolling with the `ExternalProjectDetailsPage.tsx`. Would be great to apply the fix to the detail page too.

nafiz1001:
I believe I have found the last fix needed to fix all of our scrolling problem.

UlysseFG (approved):

